PR #69, opened 2013-10-01
If an object is renamed. There is actually no traking if it is a default...

hoka:
... page.

Hi,

here a little usability improvement. 

If a default page is renamed, now the relation to the parend is fixed.


pbauer:
I implemented this for Plone 5 in https://github.com/plone/plone.app.content/pull/24

PR #224, opened 2014-06-24
Fixed bug blocking adding portlets to content types from 'typesUseViewActionInListings' list.

lewicki:
Fixed bug with adding the first portlet to objects if content type is added to typesUseViewActionInListings property. (Image and File by default.).
If portlet with edit form (i.e. calendar portlet is working) is added the error is raised as None is concatenated with string.


jensens:
after there is no response i close this one. if there is any work planned on this one please just reopen.

Issue #286, opened 2014-10-28
move mark_special_links to mockup


fulv:
this script need to be adapted to be mockup friendly and moved to another js package or resource directory:

mark_special_links

davisagli:
Seems to be working.

PR #332, opened 2014-12-15
Replace legacy mark_special_links javascript with corresponding pattern ...

fulv:
...from mockup (#286).

This goes with my pull request on plone.app.layout.


frapell:
@fulv @thet I have added additional commits to this PR, for stuff I think was missing. There are 2 missing things though:

1. There is no style. When setting the 'mark_special_links' variable to 'true' in the portal_properties, the ```<i class="glyphicon link-external"></i>``` is properly added to the markup, but there's no style visible in the rendered page.

2. The pattern is not included in the compiled JS, and for it to work, you have to put the Resource Registry in development mode.

I think those points are in the scope of this PR, and they should be fixed before we can merge.

fulv:
Thanks @frapell!
I'm not sure I understand how to handle these issues, though.  The pattern has a .less file, which works in the demo.  And I don't know how to have it included in the compiled JS.
Also, I don't know if these issues belong in CMFPlone, or in mockup.

frapell:
@fulv right, I think is the .less file I included when registering in the registry in 48c5b0cbed9fb0cc1fd7e3cb9aae34c05e740a5f right? However, I'm not seeing anything rendered... just the css class being applied, perhaps something else is needed?

As per compiling, I think we just need to compile mockup and include the resulting stuff with this merge, although I'm not sure... maybe @thet or @vangheem know if this is ok?

My guess is that when this PR gets merged, you should be able to use the feature in a Plone site, so that's why I would like to have these two missing things figured it out

vangheem:
I'll try and look at this later.

vangheem:
Should these settings get removed from the control panel and it just be assumed you modify these values through the Resource Registries base pattern options?

fulv:
I don't think they should be removed from the control panel.  (My personal
opinion.)
I was going to work on this ahead of next week's mockup meeting.

vangheem:
I'm just not sure how much I like using the IPatternsSettings way of defining options. Or, I need to provide a default IPatternsSettings for plone that handles multiple plone related pattern default options.

I was planning on removing plone_javascript_variables.js and this affects it a bit so I was just going to merge your stuff too. I can probably get this done in a few hours here.

fulv:
I can't work on it today, so if you are, don't let me stop you.

vangheem:
This is done now.

vangheem:
I manually merged. Ignore the branch. Delete it if you feel okay with it.

PR #344, opened 2015-01-02
fix markup for user/group control panel nav

davisagli:
Use tabs in the content area for navigation within the user and group control panels, instead of hijacking the edit bar.

(This looks like more changed than actually did -- the actual control panel markup is unchanged, but it got wrapped in a new div for the tabs.)


tisto:
Looks good. I guess we can merge as soon as Jenkins is green again.

PR #370, opened 2015-02-06
PLIP#20256 Add social media settings control panel

vangheem:
required by https://github.com/plone/plone.app.layout/pull/41


tisto:
Is there a PLIP for this? What's the idea behind such a control panel?

vangheem:
This is required by this pull request https://github.com/plone/plone.app.layout/pull/41

There is no plip. Probably should be but I was lazy and thought I could squeeze something simple in.

The meta tags are something we add to every site we deploy so that twitter and facebook format linked pages nicely. The functionality is close to list dublin core metadata tags.

I thought that the control panel might be slightly misleading as to what it does(since we're just adding meta tags to rendered pages) so maybe we just remove the control panel and editing those options would just be done in the registry.

tisto:
I think we need a PLIP for this, because we are adding new functionality to Plone. If we as FWT member do not stick to our own rules, how can we expect other to do so? I think lots of PLIPs that have been accepted in the past, haven't been reviewed properly by the FWT, because the PLIP implementers were members of the FWT (and I explicitly include my own PLIPs here as well). The least we can do is write a PLIP, it does not need to be long.

I think @esteele declared a feature freeze for Plone 5 some time ago already. We postponed new PLIPs that other people submitted to 5.1 because of that. Though, one could argue that this is a smaller change to the dublin core metadata that is very unlikely to break things. Therefore something @esteele and the FWT have to decide.

Apart from that I think it makes sense to include that kind of functionality in Plone core.

vangheem:
Sounds fine. Plip added: https://dev.plone.org/ticket/20256

svx:
I miss documentation about this, If this is going into core I would like to have docs about this new functionality and by this I mean docs for the end-user who will see a 'new thing' in the control panel. 
Can we have docs with [robot] screen-shots [easy to do when you write test anyway, just 2 more extra minutes]. With that we would have also immediately docs about this new feature which we can easily include into docs.plone.org

vangheem:
What would docs accomplish besides restating the control panel descriptions?

svx:
In the pilp you describe a risk:
<q>
Perhaps the control panel is a bit confusing as to what the configuration options are for. A user might suspect that social media buttons are getting added to the site. 
</q>
What I am trying to say is, that I think we need documentation about this new feature, which we can add do docs.plone.org to the part of the control panel and explaining for what these new options are for.

vangheem:
Fair enough. I think my point with that was more, I guess, to just get rid of the control panel and have it so the settings are only configured via the Configuration registry if it could lead to confusion that plone core was adding much more social media integration.

gforcada:
Just to note that this exists: https://github.com/collective/collective.opengraph in case there is something worth copying from it

vangheem:
If it's accepted, I'll add some docs. @svx Where in docs would this belong, http://docs.plone.org/working-with-content/index.html ?

svx:
@vangheem I would add it under: http://docs.plone.org/adapt-and-extend/config/index.html

btw, we are thinking about a different approach in general. Would it make sense to have in the repo of CMFPlone a docs dir where we have all docs related to cmf and would make it also sense to add here somewhere a dir for robot-screen-shots ? We were talking with @tisto about that.
With that we would be able to auto-include this into docs.plone.org and we would have always up to date docs with up to date pictures in general.
What do you think ?

vangheem:
Sounds cool to me. I defer to the testing and docs experts for that though cc @datakurre

esteele:
Part of the reason we changed up the structure of the Framework Team and review process was so that smaller PLIPs that came in while we were working on the bigger ones didn't get immediately kicked to the next release. We don't want to have a never-ending influx of features while we're trying to finish things up, but a smaller one like this seems like it won't interfere with that.

vangheem:
Perhaps the control panel should be changed to "metadata" with a few more options for microdata like phone numbers, company info, etc.

hvelarde:
I have an opinion on this: https://dev.plone.org/ticket/20256

tisto:
@vangheem @bloodbare The social media control panel does not have any robot framework tests. All other control panels have those tests, which are also important for documentation...

vangheem:
Okay. I'll add some.

tisto:
Great! Just have a look at the other robot tests, should be pretty straight forward:

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/tests/robot/test_controlpanel_site.robot

hvelarde:
so this was merged anyway? why?

PR #377, opened 2015-02-11
Prevent patterns from being scanned/initialized multiple times.

jcbrand:
For reference, see my comment here:
https://github.com/Patternslib/Patterns/issues/396#issuecomment-73722048

This commit is also relate to https://github.com/plone/mockup/pull/462

*Here follow the same explanation as in that pull request:*

If we have multiple bundles that all pull in pat-registry and then call registry.scan, then patterns will be scanned and initialized multiple times.

Instead, we need to call registry.init if it's not been called before. If it's been called before, we don't need to do anything because the patterns will automatically be registered as soon as they are loaded by require.js and evaluated by the JS engine and then automatically scanned by the registry.

*Note:* this change relies on the Patternslib registry and won't work with the fork @vangheem made of pat-registry. Hopefully that fork can now be discarded.

### DIFF
--- a/Products/CMFPlone/static/plone.js
+++ b/Products/CMFPlone/static/plone.js
@@ -30,7 +30,7 @@ if (window.jQuery) {
 
 require([
   'jquery',
-  'mockup-registry',
+  'pat-registry',
   'mockup-patterns-base',
 
   'mockup-patterns-select2',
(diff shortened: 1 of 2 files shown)


vangheem:
plone-logged-in.js would also need to be changed. Additionally, registry.xml entry missing, js should be re-cooked.

However, see my comment on patternslib. I don't think this is enough.

Products/CMFPlone/static/plone.js
```diff
@@ -48,12 +48,16 @@ require([
...
   'bootstrap-collapse',
   'bootstrap-tooltip'
-], function($, Registry) {
+], function($, registry, Base) {
   'use strict';
 
   // initialize only if we are in top frame
   if (window.parent === window) {
-    Registry.init();
+    $(document).ready(function() {
```
vangheem: FWIW, patternslib already does the $(document).ready call

PR #406, opened 2015-03-23
Implement NewsML feed syndication

hvelarde:
This was ported from collective.syndication and will need to be ported to Plone 5.0.

Refs #151 and #268

### DIFF
--- a/Products/CMFPlone/browser/syndication/adapters.py
+++ b/Products/CMFPlone/browser/syndication/adapters.py
@@ -8,6 +8,7 @@ from DateTime import DateTime
 from OFS.interfaces import IItem
 
 from Products.CMFCore.utils import getToolByName
+from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
 
 from Products.CMFPlone.interfaces.syndication import IFeed
 from Products.CMFPlone.interfaces.syndication import IFeedItem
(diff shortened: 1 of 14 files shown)


vangheem:
Does this PR address the issues in https://github.com/plone/Products.CMFPlone/pull/151 ?

Mainly need to use check=False in registry settings lookup and also provide an upgrade step in plone.app.upgrade.

Also, a PR for 5.0 would be nice.

hvelarde:
I'm migrating a customer to 4.3 and we need this; @tcurvelo will create the upgrade step and port that to 5.0.

vangheem:
the check=False in forInstance call for https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/syndication/settings.py#L29 is important also.

tcurvelo:
Here is the upgrade: https://github.com/plone/plone.app.upgrade/pull/30
I've tweaked the upgrade made by @jpgimenez. Could someone review it, please?
cc @hvelarde

vangheem:
Looks fine. Merge both as you please.

vangheem:
please make the plone 5 pull request for this also!

hvelarde:
@vangheem please review: https://github.com/plone/Products.CMFPlone/pull/408

there was a conflict on Products/CMFPlone/browser/syndication/templates/search-rss.pt

Issue #389, opened 2015-03-04
meta viewport in Plone 4 prevents zooming in beyond 100%, starts zoomed out


fulv:
This is in plonetheme.sunburst's main_template.pt:

`<meta name="viewport" content="width=device-width, initial-scale=0.6666, maximum-scale=1.0, minimum-scale=0.6666" />`

This is bad for two reasons:
- it starts the screen zoomed out and 
- it prevents zooming in beyond 100% 

max-zoom:1 is considered an accessibility problem.  What was the thinking behind this?

The current theme in Plone 5 uses:  `width=device-width, initial-scale=1.0`, so I'm only wondering if there is any risk in changing it in plonetheme.sunburst, for any further releases of Plone 4.

fulv:
Credit: @mrwweb

mrwweb:
Background: http://a11yproject.com/posts/never-use-maximum-scale/

agitator:
never had a project where ^ this was sane ... 

+1 for 
```
<meta name="viewport" content="width=device-width, initial-scale=1.0">
```

polyester:
good point. I used that one in https://github.com/plone/Products.CMFPlone/issues/407 for consistency, but there is indeed no sane reason for maximum scale.

jensens:
For those still using sunburst: This is still true https://github.com/plone/plonetheme.sunburst/blob/master/plonetheme/sunburst/skins/sunburst_templates/main_template.pt

PR #407, opened 2015-03-28
Update image_view_fullscreen.pt for mobile friendliness

polyester:
Google complains loudly about this template, and will deduct mobile karma points.


fulv:
But see:  https://github.com/plone/Products.CMFPlone/issues/389 about the `viewport` meta directive that is sunburst's default.

I added the same change to plone.app.contenttypes:  https://github.com/plone/plone.app.contenttypes/pull/219

PR #408, opened 2015-03-30
Implement NewsML feed syndication

hvelarde:
This was ported from collective.syndication and Plone 4.3.


davisagli:
FYI: this merge got reverted because it broke a test when we added CMFPlone back to checkouts: http://jenkins.plone.org/job/plone-5.0-python-2.7/674/

hvelarde:
@tcurvelo could you please take care of this?

Issue #419, opened 2015-04-08
Packages with broken or missing dependencies are no longer visible in the add-ons form


cewing:
If you have an add-on with metadata dependencies on other packages, any of which is missing or broken, then the add-on is not displayed in the add-on management form.  The data is available from the [QuickInstaller tool](https://github.com/plone/Products.CMFQuickInstallerTool/blob/master/Products/CMFQuickInstallerTool/QuickInstallerTool.py#L155) but is not used in the new Products.CMFPlone/controlpanel/browser/quickinstaller.py view.

thet:
Isn't that a more like a feature? IMO, the dependencies in a package should be well defined.

davisagli:
Unless you're developing the package and haven't defined them well yet. In Plone 4 it showed warnings in this situation. Fortunately @cewing is almost done with pull request to restore that functionality.

cewing:
It's also helpful if you're a site manager and you add an add-on and it did not properly declare package dependencies in its `setup.py`.  I'm generally of the opinion that more information is better than less when it comes to "why isn't that thing I just created/downloaded showing up to be installed"

thet:
+1. listing broken packages in an extra section is of course much better than just ignoring it.

PR #420, opened 2015-04-08
Provide information about broken packages in add-ons control panel

cewing:
This pull request provides a listing of add-on packages that are broken and therefore not installable in the add-ons control panel.

The purpose is to assist product developers in diagnosing the reason for their package not appearing in the list of available packages.  It would also be of use to site managers who attempt to add an add-on that does not properly declare package dependencies.  

A [related pull request](https://github.com/plone/plonetheme.barceloneta/pull/13) to plonetheme.barceloneta provides the css styles required to keep the styling of this new listing in accordance with those for the existing listings.

Addresses issue #419

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/quickinstaller.pt
+++ b/Products/CMFPlone/controlpanel/browser/quickinstaller.pt
@@ -196,6 +196,28 @@
           </section>
         </section>
 
+        <section tal:define="products view/get_broken;
+                             num_products python:len(products);"
+                 tal:condition="num_products"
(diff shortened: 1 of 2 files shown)


cewing:
It should be noted that I added no tests for the code created in this PR.  There are no tests currently covering the view in Products.CMFPlone.controlpanel.browser.quickinstaller.  It would be nice to write some, but I was not sure of how to begin appropriately to do so.

vangheem:
This looks fine to me and yes, someone should write some tests...

PR #456, opened 2015-04-27
Remove Zope branding

aclark4life:
This is absolutely, positively **not** to be interpreted as anti-Zope sentiment, rather it's a pro-consistent-Plone-marketing statement. We no longer need to associate Plone with Zope technologies in any public-facing way **inside** the Plone product nor should we. Strategic alignment with Zope on websites, social media, etc is still absolutely fine and should still be encouraged (e.g. I still <3 Zope!) But the Plone product itself should scream "I am Plone!" not "I am a mixture of Plone and Zope!" The latter serves no purpose anymore but to confuse people and hinder Plone's mission to be the leading open source enterprise CMS.

### DIFF
--- a/Products/CMFPlone/browser/templates/plone-overview.pt
+++ b/Products/CMFPlone/browser/templates/plone-overview.pt
@@ -133,7 +133,7 @@
         tal:attributes="href string:${context/absolute_url}/manage_main"
         title="Go to the ZMI"
         i18n:attributes="title;"
-        i18n:translate="label_zmi_link">Zope Management Interface</a>
+        i18n:translate="label_zmi_link">Advanced Management Interface</a>
       <span i18n:translate="label_zmi_link_description"> &#151; low-level technical configuration.</span>
     </p>
     <p>
--- a/Products/CMFPlone/profiles/default/controlpanel.xml
+++ b/Products/CMFPlone/profiles/default/controlpanel.xml
@@ -98,7 +98,7 @@
     i18n:attributes="title">
   <permission>Plone Site Setup: Security</permission>
  </configlet>
- <configlet title="Zope Management Interface" action_id="ZMI" appId="ZMI"
+ <configlet title="Advanced Management Interface" action_id="ZMI" appId="ZMI"
     category="Plone" condition_expr=""
     icon_expr="string:$portal_url/zope_icon.png"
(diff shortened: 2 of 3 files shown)


aclark4life:
I would also suggest branding the ZMI in a way similar to https://pypi.python.org/pypi/zope2_bootstrap. It's not apparent from the README, but when zope2_bootstrap is installed with Plone the Plone logo appears (along with other niceties.) 

screenshot 2015-04-26 20 34 09

The monkey patch used to create ``zope2_bootstrap`` was ripped off from Plone's monkeypatches to hack the ZMI, so it should not be hard to do this. I'm willing to put together another PR if folks agree we should move in this direction.

vangheem:
This is cool with me. I'm merging. Scream at us and revert if you have objections. Thanks.

aclark4life:
Thanks @vangheem. To reiterate: this is not about reducing Zope's role in Plone. If you feel it is, I suggest adding some Zope branding in the footer e.g. "Powered by Zope!"

This **entirely** about consistent branding in the UI. I don't want to see "Zope" when I'm working within the Plone UI. I don't even want to see it in the ZMI. But that does not mean Plone does not <3 Zope and Plone certainly appreciates Zope's contribution to Plone over the years starting with contribution #1: existing.

tisto:
+100

davisagli:
I don't like putting the Plone logo on the ZMI...it seems like it would make things more confusing rather than less, since there would then be 2 different user interfaces that are both labeled Plone.

I also think our goal should be to get rid of the ZMI and replace the parts of it that matter with more usable UI, rather than making it even harder to maintain by adding monkey patches.

vangheem:
These monkey patches were already in place. We're not replacing the ZMI any time soon.

Additionally, we might be forking zope anyways...

jcbrand:
I really like the idea behind zope2_bootstrap. Ideally the patches could be merged.

I agree with @vangheem, the ZMI is not going away for a long time.

jcbrand:
To allay @davisagli's fear about confusing users, perhaps the logo could be slightly modified in the ZMI, for example adding the subtext "Advanced Management Interface".

esteele:
Guys, can you just trust me that *right now* is not the time to do this, or to be discussing it publicly? I'm happy to merge in the future, but not yet.

PR #459, opened 2015-04-28
we only support `utf-8` at the moment

tomgross:
(see http://sourceforge.net/p/plone/mailman/message/29678307/)


tisto:
I think we should try to get rid of the default_charset stored in portal properties alltogether. This will allow us to get rid of those properties in future Plone versions.

PR #485, opened 2015-05-04
Make typesToList read metaTypesNotToList from new p.a.registry settings.

tisto:


Products/CMFPlone/utils.py
```diff
@@ -234,15 +234,15 @@ def getEmptyTitle(context, translated=True):
...
-    bl = ntp.getProperty('metaTypesNotToList', ())
-    bl_dict = {}
-    for t in bl:
-        bl_dict[t] = 1
-    all_types = ttool.listContentTypes()
-    wl = [t for t in all_types if not t in bl_dict]
-    return wl
+    from zope.component import getUtility
+    from plone.registry.interfaces import IRegistry
+    from Products.CMFPlone.interfaces import INavigationSchema
```
jensens: I think its fine to import these in the files top

tisto: This is work in progress. Tests failed after moving imports. I need to investigate...

tisto:
http://jenkins.plone.org/job/pull-request/109/console passes

Products/CMFPlone/utils.py
```diff
@@ -234,15 +237,13 @@ def getEmptyTitle(context, translated=True):
...
-    ttool = getToolByName(context, 'portal_types')
-    bl = ntp.getProperty('metaTypesNotToList', ())
-    bl_dict = {}
-    for t in bl:
-        bl_dict[t] = 1
-    all_types = ttool.listContentTypes()
-    wl = [t for t in all_types if not t in bl_dict]
-    return wl
+    registry = getUtility(IRegistry)
+    from Products.CMFPlone.interfaces import INavigationSchema
```
jensens: so this creates a circular import problem in the header.
I looked up why:
in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/interfaces/controlpanel.py#L4 we import from utils a json validator. this validator is used *only* in ``interfaces/controlpanel``. I propose to either move the validator out of utils into ``plone.app.z3cform`` - we have already validations defined there and json validation is very generic.

tisto:
Jenkins is green: http://jenkins.plone.org/job/pull-request/110/

tisto:
plone.app.contentlisting fix: https://github.com/plone/plone.app.contentlisting/pull/7 (though, I'm not sure where this is actually used).

Products/CMFPlone/utils.py
tisto: We are about to prepare the beta2 release. I will merge this pr now to get the fix in. We can work on that later. Can you point me to the plone.app.z3cform validators?

PR #493, opened 2015-05-07
implement tinymce control panel settings

vangheem:
This implements most of the tinymce configuration settings. I've left out a few things as we need to decide if we want to keep supporting EVERYTHING. This gets us most of the way there and includes support for AtD which is pretty important for some content editors.


vangheem:
@tisto @gforcada how do I get tests running for this on jenkins?

gforcada:
@vangheem I wrote about that here: http://gil.badall.net/2015/04/17/testing-pull-requests-and-multi-repository-changes

vangheem:
I'm going to merge this. There is a test failure related to plone.app.upgrade that I'll fix.

PR #495, opened 2015-05-07
move imaging control panel to cmfplone

vangheem:


vangheem:
This needs to be merged along with: https://github.com/plone/plone.app.imaging/pull/13 and https://github.com/plone/plone.app.upgrade/pull/31

PR #587, opened 2015-05-30
Do not require 'Enable LiveSearch'.

tisto:


davisagli:
I don't think this is the right fix. A boolean value *is* required; we want either True or False stored for this field, not None. Instead, I think the right fix is to make the checkbox widget doesn't show a required indicator; the indicator is irrelevant for that particular widget since the widget doesn't provide a way to enter None. I already took care of that in plone/plonetheme.barceloneta@9bdb7ef5b8ec9ca57bd1c7fbddf55005fff87e66

vangheem:
Won't the default value define what is stored? I don't think None will be used when default is set to False.

davisagli:
The default widget for Bool doesn't provide a way to enter anything other than True or False, so it's more of a theoretical problem. I'm saying someone could conceivably write a different widget that *does* distinguish between required and non-required booleans (providing a way to enter None for the latter but not the former), and with that widget we'd want this field to have required=True

PR #603, opened 2015-06-06
Fix issue where upgrading a product on Plone 5 resulted an error status message, because of broken mapping value for i18n message

datakurre:


davisagli:
Sorry, I think this was my fault (flake8 didn't catch it because it's a valid set literal!). Thanks for the fix.

Issue #620, opened 2015-06-14
mail-controlpanel's "Site 'From' address" field should be a text-line, not text area.


esteele:
See the final field at /@@mail-controlpanel. A textarea doesn't make sense for this one.

PR #641, opened 2015-06-20
Update 'From address' field

gforcada:
It should be a line not a text area.

Fixes: https://github.com/plone/Products.CMFPlone/issues/620


gforcada:
Tests running on: http://jenkins.plone.org/job/pull-request-5.0/27

PR #660, opened 2015-07-01
Fix search page

ebrehault:
- add a missing closing tag
- use `ajax_load` when loading results for performances sakes and to avoid funky side effects in Diazo


ebrehault:
Jenkins job: http://jenkins.plone.org/job/pull-request-5.0/43/

PR #671, opened 2015-07-13
hide ATContentTypes from products install form

vangheem:


vangheem:
Running tests here: http://jenkins.plone.org/job/pull-request-5.0/61/

PR #682, opened 2015-07-15
Adding a before publishing traverse to avoid the default view on REST calls.

tisto:

PR #689, opened 2015-07-16
Adding a before publishing traverse to avoid the default view on REST calls.

tisto:


jensens:
-1 see my comment in #682

jensens:
i close this one, if still needed feel free to reopen.

PR #697, opened 2015-07-16
align bower versions with mockup

thet:
@bloodbare is a separate grunt run necessary? i guess so.


jensens:
change log is missing

vangheem:
Not sure what a changelog would say here as it's not upgrading any bundles or js packages, just updating the bower config.

PR #747, opened 2015-07-23
Rework timezone selector.

jaroel:
Group language in the timezone selector by 'continent'.
This allows us to type 'Amsterdam' instead of 'Europe/Amsterdam'

screen shot 2015-07-23 at 10 53 19

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,11 +10,15 @@ Changelog
 
 - Fix link to documentation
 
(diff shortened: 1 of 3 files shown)


jaroel:
Running tests: http://jenkins.plone.org/job/pull-request-5.0/132/

thet:
Looks great!
What if we offer these grouped lists in the language and event control panels too?

jaroel:
All for it, got to start somewhere :)

PR #757, opened 2015-07-26
Rename showEditableBorder to showToolbar

vangheem:


vangheem:
http://jenkins.plone.org/job/pull-request-5.0/153/

hvelarde:
shouldn't we also state when this deprecated methods are going to be removed?

PR #758, opened 2015-07-28
move Plone specific getDefaultPage code+view from plone.app.layout ov…

jensens:
…er to Products.CMFPlone. This avoids a circular dependency. Also its not really layout only related code.

After this is in plone/plone.app.layout#52 needs to be merged, but no need to do at once.


jensens:
try: http://jenkins.plone.org/job/pull-request-5.0/182/

jensens:
try: http://jenkins.plone.org/job/pull-request-5.0/182/

jensens:
again - now it should work http://jenkins.plone.org/job/pull-request-5.0/184/

PR #765, opened 2015-07-29
use more general ITraversable as base interface.

jensens:
in order to make plone.app.discussion happy


pbauer:
@jensens Would you care to explain why?

jensens:
http://jenkins.plone.org/job/pull-request-5.0/191/ passes

jensens:
plone.app.discussion uses base objects that are not IZopeObject.
I tend to change it back to Interface (*) later on.

PR #777, opened 2015-08-01
Remove Products.ExternalEditor

gforcada:
Seems to not be used anywhere...

There should fail only the tests in CMFPlone itself that test that it works... should that code be moved to P.ExternalEditor itself?

Actually the point of this pull request is to get rid of the last svn dependencies that we still have...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,6 @@ setup(
         'Products.CMFUid',
         'Products.DCWorkflow',
         'Products.ExtendedPathIndex',
-        'Products.ExternalEditor',
         'Products.GenericSetup >=1.4',
         'Products.MimetypesRegistry',
         'Products.PasswordResetTool',


jensens:
So, why not import it into the zopefoundation or collective organisations?
ExternalEditor is used by a bunch of people. I'am not sure if removing from core is a good solution. At least we should keep it in the ecosystem. 
Then removing code related to it from CMFPlone is fine and moving it over to ExternalEditor is a great idea.

hvelarde:
yes, please lets stop removing stuff from the core just because one of us is not using it.

removal of old stuff is always needed but should be done on in a planned and orderly way. deprecation warnings were used a lot previously in Plone and now we are just skipping them. somebody has to centralize this kind of decision making to avoid unwanted surprises for everybody else.

what about @esteele?

jensens:
I'd say @esteele together with @plone/framework-team - I'll put the question at the agenda for next meeting

gforcada:
No longer needed as it's already out of svn.zope.org :tada: 

https://github.com/plone/buildout.coredev/commit/9efd99a5dbbaf271a1200d6f3f872924d20448f9

Issue #793, opened 2015-08-04
Adding extra TinyMCE styles from a Plone add-on


ebrehault:
If we want to add new styles in our TinyMCE config, we have to declare them in our profile's `registry.xml` and provide the path to the related CSS in our theme's `manifest.xml`. That's fine when we create a Plone theme.

But how to do that if we are just creating an add-on?
We can still declare our styles in `registry.xml`, but there is no way to give the path to our CSS as we do not have a Diazo theme into our module.

vangheem:
Do you mean CSS used in the editor or the styles drop down?

vangheem:
Can you look at the tinymce configuration panel? @@tinymce-controlpanel

A lot is configurable with the registry and does not require a theme. More is configurable with registry than there is with the theme.

ebrehault:
I need both:
- I want to user to be able to choose one of my add-on specific styles in the TinyMCE menu (this does work by declaring the needed stuff in `registry.xml`)
- I also want the choosen style to render specifically in the editor, and that's what I cannot do.

I have noticed the "Choose the CSS used in WYSIWYG Editor Area" parameter in @@tinymce-controlpanel, but apparently it is not loaded by the editor at the moment, and anyway, I would probably need it to be a list, not single value, as I do not want my add-on to replace the existing CSS.

vangheem:
Right, as with current tinymce on plone 4, it seems like the single content css value line carried over. It's a comma separated value.

vangheem:
The content css value to be a list so it can easily be configured with registry.xml. Then the pattern config can just convert that into a comma-separated value for tinymce.

ebrehault:
OK, so I guess we just need to parse the registry value here and append the resulting URLs to the existing list here https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/patterns/__init__.py#L32

Right?

vangheem:
It's already doing that.

I think we should change https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/interfaces/controlpanel.py#L400 to be a list.

ebrehault:
Yes we need to change the schema, but I do not see any mention of the `content_css` attributes in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/patterns/__init__.py#L32 at moment.
[...]

PR #794, opened 2015-08-04
@@tinymce-controlpanel content_css attribute must be a list

ebrehault:
- Turn @@tinymce-controlpanel `content_css` field into a list, so we can add several CSS URLs (useful when add-ons need to provide extra TinyMCE styles),
- fix TinyMCE config getter so it considers the ``content_css`` value.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,11 @@ Changelog
 - Rework language selection in @@plone-addsite.
   [jaroel]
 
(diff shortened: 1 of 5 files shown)


ebrehault:
Jenkins job: http://jenkins.plone.org/job/pull-request-5.0/208/

ebrehault:
Fix #793

PR #809, opened 2015-08-06
Punctuation update to syndication.py

3dogMcNeill:
added periods, minor word edits

### DIFF
--- a/Products/CMFPlone/interfaces/syndication.py
+++ b/Products/CMFPlone/interfaces/syndication.py
@@ -162,12 +162,12 @@ class ISiteSyndicationSettings(Interface):
 
     search_rss_enabled = schema.Bool(
         title=_(u'Search RSS enabled'),
-        description=_(u'Allows users to subscribe to feeds of search results'),
+        description=_(u'Allows users to subscribe to feeds of search results.'),
         default=True)
 
     show_author_info = schema.Bool(
         title=_(u'Show author info'),
-        description=_(u'Should feeds include author information'),
+        description=_(u'If feeds should include author information.'),
         default=True)
 
     max_items = schema.Int(
@@ -203,11 +203,11 @@ class ISiteSyndicationSettings(Interface):
     show_syndication_button = schema.Bool(
         title=_(u"Show settings button"),
         description=_(u"Makes it possible to customize syndication settings "
-                      u"for particular folders and collections "))
+                      u"for particular folders and collections. "))
 
     show_syndication_link = schema.Bool(
         title=_(u"Show feed link"),
-        description=_(u"Enable RSS link document action on the syndication "
+        description=_(u"Enable RSS link document action on the syndicated "
                       u"content item."))
 
 


ebrehault:
If you want to change those description strings, you have to update https://github.com/collective/plone.app.locales accordingly, because those english strings are used as .po keys.
If you change the key without updating the .po files, the matching is broken and the default value (= the english string) will appear instead of the translated version.

ebrehault:
By the way, if that's the first time you work in plone.app.locales or with .po files, be aware you are not supposed to change manually all the .po files for the languages to update your keys.
There is resync tool, see doc here: https://github.com/collective/plone.app.locales/blob/master/utils/README.txt

3dogMcNeill:
Yes, this is first time I've worked with these. Thought it would be easier to make changes than list them and have somebody else do it. Alas... So what is the safest option—I don't want to mess anything up.

ebrehault:
Well, actually, if you do not feel too confident about resyncing the .po files, you can just change the english translation of your entries:
- you keep the descriptions as they were in Products.CMFPlone, for instance:
```
description=_(u'Allows users to subscribe to feeds of search results')
```
- you modify the corresponding english translation in this .po file https://github.com/collective/plone.app.locales/blob/master/plone/app/locales/locales/en/LC_MESSAGES/plone.po#L522, which is at the moment like that:
```
#: CMFPlone/interfaces/syndication.py:165
msgid "Allows users to subscribe to feeds of search results"
msgstr ""
```
meaning there is no english translation for this given key, so the key is used by default.
If you update it that way:
```
#: CMFPlone/interfaces/syndication.py:165
msgid "Allows users to subscribe to feeds of search results"
msgstr "Allows users to subscribe to feeds of search results."
```
then this `msgstr` will become the english translation for the key, and will be displayed in english context instead of the current default value (and in other lang context, the existing translations will not be impacted as the msgid value hasn't changed).

So basically, you give up the current PR in Products?CMFPlone and you make a new PR in plone.app.locales.

I say that cheap because it would be more consistent to update the key and the value, but on the other hand, it would not happen if .po keys were actual keys (something like `CMFPlone_search_rss_enabled_description` for instance) and not english sentences.

For your information: the canonical way to call the translating mechanism is:
```
description=_('CMFPlone_search_rss_enabled_description', default=u'Allows users to subscribe to feeds of search results')
```
That way, the key is just a key, so it can stay the same forever, the default parameter is used as a fallback if no translation provided for the current lang, plus it eases the reading of the source (as it is more meaningful than the key). And if we want to change this default value, we just change it in one single place.

3dogMcNeill:
Thank for the explanation, it's very helpful! I'll figure out how to get rid of this PR and will add the English translations.

vangheem:
Wouldn't the correct way of doing this be to provide ids for these messages?

I thought the english defaults were always updated from what was in the code.

Sounds like I've been doing it wrong also otherwise because I only ever update what is in the code and let others do translation.

3dogMcNeill:
I'm glad to do it whichever way is best and within my skill level. I'll hold off on doing anything until there's a decision on what to do.

jensens:
yes, please add message ids. Current problem is, that someone forgot them to set before. So afterwards it looks like:

    description=_(
        u'search_rss_enabled_description', 
        default=u'Allows users to subscribe to feeds of search results.'
    ),

jensens:
This Pull Request has stalled for a while now and is closed. If you care about this Pull Request please reopen at any time.

PR #832, opened 2015-08-20
initial checkin of minimal bundle just for toolbar and public UI

djay:


vangheem:
@djay I'm fine with this. I rebased and fixed a few things. Would you might reviewing to check this is what you're looking for?

jensens:
changelog entry is missing

djay:
I think it still needs some work:
- change the name to something like ```toolbaronly``` to indicate it includes nothing else but what is needed to get the toolbar to work. It includes nothing like livesearch or input groups etc. The user can include other bundles to get that.
- remove fonts and other css that aren't related to the toolbar. Weird fonts shouldn't be needed for the toolbar.
- dropdowns don't currently work for this bundle. What is needed to make that work @vangheem? Was that the only problem you encountered @jcerjak?

Also still not sure on how to use bundles when you are using a custom theme that includes another backend theme. I'm guess what we have to do is
- in our manifest include all the bundles
- in our frontend.cfg we filter out all the bundles except our custom bundle + toolbaronly bundle.
- in the backend.cfg in barceloneta we filter out everything except the standard plone bundles. but somehow we also need to include the custom tinymce styles? Any ideas on this @bloodbare ?

jcerjak:
@djay , I didn't do a thorough test, but the following didn't work:
* submenus - it opens a new page when clicking on menu instead of opening a submenu
* toolbar doesn't collapse/expand

There was one error in the console about jquery.min.map missing, but I don't know if it's related to the issues..

jensens:
This Pull Request has stalled for a while now and is closed. If you care about this Pull Request please reopen at any time.

hvelarde:
be careful on leaving stale branches

jensens:
indeed, i'am also the one delting them so i know the pain, but i think its better to delete after a a while manually - say 6 month - to make it easy to reopen.

PR #835, opened 2015-08-24
Require POST for several mail scripts

mauritsvanrees:
Otherwise someone could be tricked by a spammer into clicking a normal link and this could already send an email.

See https://pypi.python.org/pypi/collective.honeypot for background and for more radical fixes. I am looking into getting some of that into Plone 5. But these POSTs are the simple fixes that may already help a bit. Good for 4.3 I would say.

### DIFF
--- a/Products/CMFPlone/skins/plone_form_scripts/discussion_reply.cpy
+++ b/Products/CMFPlone/skins/plone_form_scripts/discussion_reply.cpy
@@ -14,6 +14,9 @@ from Products.CMFPlone import PloneMessageFactory as _
 mtool = getToolByName(context, 'portal_membership')
 dtool = getToolByName(context, 'portal_discussion')
 req = context.REQUEST
+if req.get('REQUEST_METHOD', '').upper() != 'POST':
+    from zExceptions import Forbidden
+    raise Forbidden('Use POST please.')
 
 if username or password:
     # The user username/password inputs on on the comment form were used,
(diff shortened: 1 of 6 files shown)


mauritsvanrees:
http://jenkins.plone.org/job/pull-request-4.3/79/

mauritsvanrees:
This caused a few test failures in ATContentTypes, where the tests relied on GET requests to discussion_reply. I have fixed those tests to use a POST request.
Try again, Jenkins: http://jenkins.plone.org/job/pull-request-4.3/81/

mauritsvanrees:
Jenkins is happy. Ready for review.

Products/CMFPlone/skins/plone_form_scripts/discussion_reply.cpy
```diff
@@ -14,6 +14,9 @@ from Products.CMFPlone import PloneMessageFactory as _
 mtool = getToolByName(context, 'portal_membership')
 dtool = getToolByName(context, 'portal_discussion')
 req = context.REQUEST
+if req.get('REQUEST_METHOD', '').upper() != 'POST':
+    from zExceptions import Forbidden
```
hvelarde: @mauritsvanrees it is an accepted practice to do the imports in the middle of the code?

mauritsvanrees: In Python skin scripts I would say: absolutely.  In such a script, every line that gets executed, is checked for security, which takes relatively long. So you want to avoid an import unless you really need it.

hvelarde: I understand, awesome!

Issue #125, opened 2013-11-07
Remove portal_properties Tool.


tisto:
Once all settings have been moved to plone.app.registry, we should be able to remove portal_properties.

esteele:
All of the properties currently set by Plone core have been migrated to the configuration registry. I think we can mark PropertiesTool as deprecated in 5.0 and remove in 5.1.

vangheem:
Moved from plip to bug. We need to mark it as deprecated and fully remove this in 6 IMO.

pbauer:
How do we properly deprecate portal_properties? ONe option is to simplyt mention it in the release notes. Another would be to throw deprecation warnings when accessing properties or creating propertysheets. (i.e. decorating addPropertySheet, _setPropValue and getProperty)

mauritsvanrees:
When running some tests, I saw this warning again:

```
DeprecationWarning:
The portal portal_properties tool will be removed in Plone 6.1.
Use the portal_registry instead.
Check https://github.com/plone/Products.CMFPlone/issues/125 for more details.
```

I have added the current issue to the Plone 6.1 project board.

Grepping for `portal_properties` only in Python code already pulls up a lot of lines.  Some will be for backward compatibility and can simply be removed.  I hope this is true for all.

```
$ ack --python portal_properties
Products/CMFPlone/tests/testPortalCreation.py
45:        self.properties = self.portal.portal_properties
951:        self.properties = self.portal.portal_properties

Products/CMFPlone/tests/testNavigationView.py
126:        ntp = self.portal.portal_properties.navtree_properties
137:        ntp = self.portal.portal_properties.navtree_properties

Products/CMFPlone/tests/test_zmi.py
124:    def test_portal_properties(self):
125:        url = "portal_properties/manage_main"

Products/CMFPlone/tests/testWebDAV.py
23:        ptool = getToolByName(self.portal, "portal_properties")

Products/CMFPlone/PropertiesTool.py
28:    id = "portal_properties"
84:    #   'portal_properties' interface methods
107:        "The portal portal_properties tool will be removed in Plone 6.1. "
135:        "The portal portal_properties tool will be removed in Plone 6.1. "

Products/CMFPlone/setuphandlers.py
78:        "portal_properties": "General settings registry",

Products/CMFPlone/browser/author.py
200:        self.portal_properties = getUtility(IPropertiesTool)

Products/CMFPlone/exportimport/tests/testPropertiesTool.py
[...]

PR #864, opened 2015-09-05
Remove no-longer-used properties from portal_properties

esteele:
These properties either:
* Have been migrated to the configuration registry
* Are no longer used in Plone core

See also: https://github.com/plone/plone.app.upgrade/pull/45

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ Changelog
 5.0b5 (unreleased)
 ------------------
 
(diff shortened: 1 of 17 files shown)


esteele:
Refs plone/Products.CMFPlone#125

hvelarde:
@esteele please try to squash you commits before merging them; see:

https://github.com/plone/plone.api/pull/274/files

Issue #346, opened 2015-01-07
missing site_description and display_pub_date_in_byline in ISite controlpanel


thet:
the ``ISiteSchema`` in ``plone.app.controlpanel`` 2.3.x [1] defines two fields, which are missing in ``Products.CMFPlone`` ``ISiteSchema`` interface [2]:

- [ ] site_description: The site description is available in syndicated content and in search engines.
- [ ] display_pub_date_in_byline: Displays content publication date on site pages. (as per [PLIP 8699](https://dev.plone.org/ticket/8699))

[1] https://github.com/plone/plone.app.controlpanel/blob/2.3.x/plone/app/controlpanel/site.py
[2] https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/interfaces/controlpanel.py

/cc @tisto

tisto:
Site description was removed because it was never used anywhere in Plone 4.3.x, not in syndicated content nor for search engines. https://github.com/plone/Products.CMFPlone/issues/139

I don't recall why or when display_pub_date_in_byline was removed though. 

I'd be ok with re-adding both settings if we actually implement the functionality and write robot tests for it.

hvelarde:
@tisto @rodfersou is working on that, please guide him on how to write the tests you want.

Issue #741, opened 2015-07-21
Should be easy to create have a byline


djay:
## User problem

Some sites show some combination of
- last updated date
- last published date
- creators name
- last editors name

The person in charge of deciding _what_ information is shown on the page (information architect/content team leader/site admin/themer) should be able to include this information in the page easily and flexibly being able to pick which of the above they want to show. Showing at least the publication date is a common feature of blog posts and news items in other CMS's
## History

The byline was a feature used both by internal editors and reused to show to the public (with reduced functionality, ie no history). Currently for 5.0 it is hidden by default because there is last updated date is shown in the toolbar and the history view provides the other info at a single click.

replaces https://github.com/plone/Products.CMFPlone/issues/668 and https://github.com/plone/Products.CMFPlone/issues/714
## Options
1. Have a tickbox to turn the byline on
   - Fix the current tickboxes in security and site which previously determine if the byline is visible to the public and if it should show last published date. This would turn it on for everyone or off for everyone.
     - should probably move somewhere more sensible like theming.
     - improve them to give more options about showing date without author #668
     - pro: backwards compatible setting. 
     - con: makes it harder for themers to cope with visual elements that can be turn on thus breaking the theme (but we already have that with things like TOC)
2. Have the viewlet hidden but easily enabled by switching on the viewlet ttw. Document how to use diazo to hide parts you don't want.
3. ~~Always have the byline turned on. Document how to use diazo to hide parts you don't want or all of it~~
4. ~~Document how to use diazo to create your own byline using parameter expression like `modified = context/modified` and a diazo rule like `<insert-after css:theme="#content h1"><p>Last modified {$modified}</p></insert-after>`~~
   ~~\- pro: teaches them more flexible approach. code is small and easy to copy
   ~~- con: means that have to modify an out of the box theme.~~
5. solve "way to include dynamic or rich content into the text of a page" #469, ie by mosaic or similar. byline would be a tile. or by htmlfragments (requires coding)
6. ~~have something like collective.listingviews built into plone which enables admins to add portlets with common metadata without coding which can then be styled using diazo.~~

davilima6:
It seems you forgot to mention original thread's most voted suggestion: byline stays exactly as it is but hidden. When Mosaic comes out, byline gets available as a configurable tile and then we delete old settings. As a bonus we avoid changing twice the recommended way of having a byline in Plone.

[...]

PR #865, opened 2015-09-05
Add display_publication_date_in_byline to Site control panel

esteele:
See also: 
* https://github.com/plone/plone.app.layout/pull/58
* https://github.com/plone/plone.app.upgrade/pull/46

Refs plone/Products.CMFPlone#346

### DIFF
--- a/Products/CMFPlone/interfaces/controlpanel.py
+++ b/Products/CMFPlone/interfaces/controlpanel.py
@@ -967,6 +967,12 @@ class ISiteSchema(ILockSettings):
         default=u'',
         required=False)
 
+    display_publication_date_in_byline = schema.Bool(
+        title=_(u'Display publication date'),
+        description=_(u'Show the date a content item was published in the byline.'),
+        default=False,
+        required=False)
+
     toolbar_position = schema.Choice(
         title=_(u'Position where the toolbar is displayed'),
         description=_(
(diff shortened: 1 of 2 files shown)


djay:
Isn't the byline switched off for public views? Wouldn't it be confusing to retain this setting if it has no effect on the theme? see discussion in #741 and the issues it replaces

vangheem:
@djay the byline viewlet code is still there. That is what @esteele is updating.

What is the status of byline anyways? Did it ever get re-added. Or did we decide on a compromise. It looks like the viewlet class is still there so we just have to add the zcml include back in?

djay:
@vangheem personally I think the byline as it has been done traditionally is a bit weird. It served two functions. one for the public should your theme require it. Another for editing/admin purposes for logged in users. Thats why it has this toggle mode to determine if its public or just for logged in users. We no longer need it for logged in users as the toolbar replaces it.
For the public hector makes the argument that many sites use this, but I think really just for certain content types like forum posts, blog posts or news posts. But we implement as all or nothing across the board. So you’d have to do fancy diazo to restrict it down to just the content types you want.
Some sites do have a generic public “page last modified on” but they are not that common anymore I think. 

Another alternative is a type specific mechanism to add a byline, and one thats within the themers control. 
Maybe a byline behaviour enabled by default for news types that has some specific rules in the barceloneta theme.
But that still makes and extra step for someone to make a generic theme. I still think its really a theme choice as to if you want to display that information to the public or not. Same as having a footer or a search bar or a hit counter. But others seem to disagree with me on this. So no clear decision. 
BTW a theme implemented byline would be something in barceloneta that uses theme TAL expressions. I’m not even sure you can put theme TAL expressions in a manifest.cfg right?

hvelarde:
@djay this has to be reverted as it was removed without consensus; we can talk about it later, but we need the standard behavior back in Plone 5 final.

Issue #845, opened 2015-08-28
Still compatibility issues with legacy bundles, require.js (collective.elasticindex on Plone 5.0b4)


zopyx:
Testing our implementation of collective.elasticindex with Plone 5.0b4

https://bitbucket.org/dasgehirn/collective.elasticindex/src

After installing collective.elasticindex in Plone 5 (bin/buildout -c buildout-50.cfg) 
I see the following error on the JS console:

Uncaught Error: Mismatched anonymous define() module: function (){return Mark;}
http://requirejs.org/docs/errors.html#mismatchmakeError @ require.js:166intakeDefines @ require.js:1239(anonymous function) @ require.js:1427

vangheem:
Master should have a workaround for these situations. Please try again when a new beta is out and reopen is you still see problems. Thank you.

vangheem:
For reference: https://github.com/plone/Products.CMFPlone/pull/870

PR #870, opened 2015-09-07
reset requirejs when plone legacy bundle is rendered to prevent errors

vangheem:


vangheem:
@bloodbare What do you think about this?

This is to address possible issues with legacy js files trying to use requirejs. See https://github.com/plone/Products.CMFPlone/issues/845

It seems harmless--just trying to prevent possible issues for people.

bloodbare:
Yes its a bit harmless, why it should be a problem to load a require module on legacyjs ? I know its not a good pattern, but does it produce some error ? What I was worried related to that is that if the the bundle is not compiled correct, there is no way to force recompilation of legacyjs like bundles.

vangheem:
See https://github.com/plone/Products.CMFPlone/issues/845 mismatch error can happen.

Right now, you can force re-compilation of legacy bundle by reordering the legacy resource list.

bloodbare:
I see the error, shouldn't we register resources that are AMD ? btw I'm sad we can't just use requirejs modules on non compiled js, but I feel fine with this PR.

vangheem:
Yes, users should register the resources AMD style but not all users do... Some users will only ever think about, "I need to include this file" and not know how to do dependency management.

bloodbare:
I still think that people should register js that is using and use the Products.CMFPlone.resources.add_resource_on_request function to add an specific resource to their view.

vangheem:
I agree...

PR #876, opened 2015-09-07
getToolByName: accept tool from mock module.

mauritsvanrees:
This is needed for me locally when running:

  bin/test -s Products.Archetypes -m test_pawidgets

No idea why Jenkins is happy.  Locally I get errors:

  TypeError: Object found is not a portal tool (portal_properties)

This is because those tests are accessing
self.context.portal_properties where self.context is an instance of
mock.Mock, so the portal_properties are also mock.Mock.

If anyone knows why Jenkins is hopping along nicely without failures, that would be interesting to know. Maybe it is because these are errors, not failures, but I do not see those errors on Jenkins either.


mauritsvanrees:
Test is running at http://jenkins.plone.org/job/pull-request-5.0/267/

mauritsvanrees:
Tests pass. Ready for review.

PR #882, opened 2015-09-08
Better default rules for robots.txt

hvelarde:
@pbauer @vangheem plesae take a look


miohtama:
Not sure if this is valid anymore, but in the past I have added `/set_language` on this list.

PR #906, opened 2015-09-10
Fix #862; input checkbox and radio now have min-width equal to width

### LINKED ISSUE #862: Profile listing on site creation has alignment issues
When using the advanced version of the plone-addsite view (/@@plone-addsite?advanced=1), the list of profiles has some alignment issues:

screen shot 2015-09-04 at 6 06 31 pm

ichim-david:
This is done in order to override the big 180px rule set on normal inputs.

I'm attaching a Google Chrome inspector session which shows the fix in action.

without-min-width

with-min-width

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,8 @@ Changelog
 5.0rc2 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- fix #862: Profile listing on site creation has alignment issues
+  [ichim-david]
 
 
 5.0rc1 (2015-09-08)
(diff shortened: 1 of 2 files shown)


pbauer:
@sneridagh or @albertcasado can you please check if that change is ok?

ichim-david:
@pbauer  @sneridagh @albertcasado  I have looked where ++resource++plone-admin-ui.css and I found it is used in:
@@plone-upgrade
@@plone-overview
and @@plone-addsite?advanced=1 - only here we actually have inputs.
Given this fact I think that the change is more than safe and effective at fixing the problem at hand.

pbauer:
@ichim-david thanks!

PR #909, opened 2015-09-10
add search_results_description_length to registry and use in livesearch

pbauer:


pbauer:
@vangheem better?

Products/CMFPlone/browser/search.py
```diff
@@ -203,11 +205,15 @@ def __call__(self):
         results = self.results(batch=False, use_content_listing=False)
         batch = Batch(results, per_page, start=(page - 1) * per_page)
 
+        registry = queryUtility(IRegistry)
+        length = registry.get('plone.search_results_description_length')
```
pbauer: @vangheem better?

vangheem: That works!

PR #919, opened 2015-09-11
upgrade tinymce

vangheem:
Hi all, we need to upgrade TinyMCE and it seems like the update to 4.2.5 changes the look and feel a bit for TinyMCE.

screen shot 2015-09-10 at 11 58 29 pm

What does everyone think about this?


frapell:
I am +1 with upgrading, the jump from tinymce 3.x was bigger than this, so I don't see why we shouldn't do it this time?

jensens:
overall +1 here too - but question is if we want to do this in Plone 5 RC phase? What does our release manager @esteele think?

esteele:
Please go ahead and update this.

vangheem:
tests running: http://jenkins.plone.org/job/pull-request-5.0/293/

jensens:
http://jenkins.plone.org/job/pull-request-5.0/292/

PR #954, opened 2015-09-11
Login settings migration

esteele:
Move login settings out of portal_properties and into the registry. Since these are settings you really aren't going to use unless you know you need them, I opted to not put them into a control panel. They can easily be set from the configuration registry UI.


vangheem:
tests running: http://jenkins.plone.org/job/pull-request-5.0/295/

PR #992, opened 2015-09-16
add hero-view to be included by barceloneta rules.xml

pbauer:


bloodbare:
+1 to merge

hvelarde:
-1 that message should be on the body text of the fontpage object; I still don't understand why we need a view for this.

pbauer:
The hero unit is outside the text (above the title). Adding it in the richtext as special html and moving (not copying) it to the top with diazo is possible but I guess pretty ugly xslt.

PR #1016, opened 2015-09-17
Make sure portal_actions are imported before default portlets

vangheem:


vangheem:
tests: http://jenkins.plone.org/job/pull-request-5.0/333/

Issue #785, opened 2015-08-03
Turning off "Generate tabs for items other than folders." has no effect on pages, events ...


agitator:

pbauer:
During the migration of the navigation-controlpanel that got lost. The registry-value set in is `nonfolderish_tabs`, but the navigation checks for a site_property called `disable_nonfolderish_sections`. In Plone 4 the setting nonfolderish_tabs was set to change said site_property but that functionality is gone. It could be fixed by having the navigation check if `nonfolderish_tabs` is False.

But: Is that setting actually useful? I would expect that being able to select which types should appear is enough since you can create the same results with more control. Sometimes you maybe don't know which of types are folderish and which are not. I'd consider removing it completely.

hvelarde:
+1 makes sense

pbauer:
Fixed (not removed) in 938d68c733be49ba84f8c6ec7508f0ce29f4b39d

PR #1021, opened 2015-09-17
Remove nonfolderish_tabs. Show all except File and Image in navigation

pbauer:
See https://github.com/plone/Products.CMFPlone/issues/785

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@ Changelog
 5.0rc3 (unreleased)
 -------------------
 
+- Remove setting 'nonfolderish_tabs' (ex-'disable_nonfolderish_sections')
+  and by default display all types except File and Image in Navigation.
+  See https://github.com/plone/Products.CMFPlone/issues/785
+  [pbauer]
+
 - Fix #950: Missing personal toolbar when expanding the horizontal toolbar
   [ichim-david]
 
--- a/Products/CMFPlone/browser/navigation.py
+++ b/Products/CMFPlone/browser/navigation.py
@@ -105,9 +105,6 @@ class CatalogNavigationTabs(BrowserView):
 
         query['is_default_page'] = False
 
-        if not navigation_settings.nonfolderish_tabs:
-            query['is_folderish'] = True
-
         return query
 
     def topLevelTabs(self, actions=None, category='portal_tabs'):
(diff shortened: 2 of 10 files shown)


pbauer:
Jenkins: http://jenkins.plone.org/job/pull-request-5.0/338/

pbauer:
test are green

tisto:
@pbauer I was under the assumption that the nonfolderish_tabs setting is about the global navigation tabs, not about the portlet navigation. Those are two different things. I usually turn the nonfolderish_tabs setting off for all my Plone sites and use the displayed_types setting to control the portlet navigation.

pbauer:
Wow. I did not know that. So ```nonfolderish_tabs ``` is only used for the tabs whereas ```displayed_types```is used for tabs and navigation-portlets. 
Would it be good to have ```displayed_types``` only used in the navigation-portlets and a new setting ```types_displayed_in_tabs``` only used for the tabs? I still think that the distinction between folderish and no-folderish is not really useful here (especially since all types created ttw are folderish). 

My proposed defaults would be:
types_displayed_in_tabs: All default types except files and images
displayed_types: All default types

tisto:
I think so, yes. I still think making TTW types folderish by default is a big mistake. This will confuse the hell out of people and moving partly into that direction is by far the worst thing we could do in my opinion. Do it right or don't do it at all, everything else will be a disaster.

A new setting is not necessary in my opinion, if we stick with folderish/non-folderish types in Plone 5.0, and this is what the FWT decided. Especially since we are talking about release candidates now.

pbauer:
Ok, then we'll simply keep the setting that is now on the registry and postopne this change to a later time. I clarified the wording for the navigation-control-panel here: https://github.com/plone/Products.CMFPlone/pull/1022

@tisto Your rant about the default for folderish types in the dexterity-control-panel is but a little out of place here. You should discuss that in the FWT if you think it will be a disaster.

tisto:
@pbauer I agree. We discussed this in the FWT and made a decision to stick with folderish/non-folderish types for 5.0. This decision also means that we are not going to sneak in small changes towards that direction (in my personal opinion).

jensens:
yes, i agree with @tisto, keep the setting. yesterday it worked ;) as i manually tested it. in future we may have a different definition of "folderish", but now its too early to make changes here.

PR #1022, opened 2015-09-18
Improve wording in navigation controlpanel

pbauer:


Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -787,9 +787,8 @@ class INavigationSchema(Interface):
         title=_(u"Automatically generate tabs"),
         description=_(
             u"By default, all items created at the root level will "
-            u"add to the global section navigation. You can turn this off "
-            u"if you prefer manually constructing this part of the "
-            u"navigation."),
+            u"appear as tabs. You can turn this off if you prefer manually "
```
tisto: Maybe "appear as tabs in the global section navigation".

tisto:
@pbauer maybe we should even move the global navigation settings in a separate fieldset to make a clear distinction between global navigation and portlet navigation.

jensens:
some settings has effects on tabs only, othgers on tabs and navigation. its difficultt. i'd prefer to keep them together. anyway, its another taks and this change here is fine, so i merge.

PR #1044, opened 2015-09-19
Fix #817: When saving the filter control panel show a flash message with info on caching.

### LINKED ISSUE #817: HTML filtering requires a "re save" on each page for the settings to take effect.
Using:  Plone 5.0b4.dev0 (5007)

Been testing the HTML filtering settings using iFrame video. 
When I add an "iframe" video the saved page does not show the video (as expected). But when I toggle ON the "Disable HTML Filtering" checkbox the video still does not display on the page (even after emptying cache). If I edit the video page and immediately "Save" (making no edits) the video displays as expected. The reverse is also true. If I have video correctly displaying on the page (with html filtering disabled) and uncheck "Disable HTML Filtering" checkbox in the HTML Filtering panel, go back to the video page and reload the video still displays. If I edit the page and immediately save the page the video properly does not display. Is this the expected behavior?

In addition. If I add "iFrame" to the "Custom tags" dialogue box I cannot get the video to display with "Disable HTML Filtering" checkbox OFF. Even if I do the "Edit" and "Save" trick from above.

jcerjak:
See #817 for more information.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@ Changelog
 5.0rc4 (unreleased)
 -------------------
 
+- Fix #817: When saving the filter control panel show a flash message with
+  info on caching.
+  [jcerjak]
+
 - Remove Chrome Frame from ``X-UA-Compatible`` HTTP header as it's deprecated.
   [hvelarde]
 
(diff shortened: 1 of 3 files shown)


vangheem:
Should it be a warning box instead of an info?

jcerjak:
Was thinking about that myself, but wasn't sure. Perhaps warning is more appropriate here, I agree.

jcerjak:
Warning box has a white background, is this only my instance broken or is it a bug?

Issue #1041, opened 2015-09-19
Issues with new features in GenericSetup


pbauer:
The PR https://github.com/zopefoundation/Products.GenericSetup/pull/16 was merged but has negative effect for Plone. See https://github.com/zopefoundation/Products.GenericSetup/issues/19

pbauer:
I removed Products.GenericSetup from the coredev-checkouts untill we know more.

pbauer:
Also it broke the build...

mauritsvanrees:
This should be fixed by 1ff71cad9a4b22182b200ea71fad7edc4c08caa2.  Here we basically revert to the old GS behavior of reapplying all profiles, at least when we create a site. That is probably fine for the moment, until we figure out why the new behavior breaks.
I wrote the code so that it works with old and new GS. That makes it easier to add GS to the checkouts again and see what happens.

mauritsvanrees:
This should be done in Plone 4.3 too, but let's first see if this really helps on Plone 5. Pull request for checkout of GenericSetup added, so we can check this.

mauritsvanrees:
I created GS exports of a Plone Site generated with old and with new setuptools, without my fix from above. This is to see what else might have changed. I see two things:

- registry.xml has a different order for several items. I have checked: the reordered items are still all there, with the same value, just a different order.
- plone.app.discussion workflow settings are not applied, as Philip already noticed. Or rather, they are applied, as I can see with some added print/pdb statements, but something overwrites it.

```
$ diff -U 3 -r good bad
diff -U 3 -r good/registry.xml bad/registry.xml
--- good/registry.xml	2015-09-20 01:51:00.000000000 +0200
+++ bad/registry.xml	2015-09-20 02:05:45.000000000 +0200
@@ -2642,10 +2642,10 @@
       </value_type>
     </field>
     <value>
+      <element>resource-plone-app-discussion-javascripts-comments</element>
       <element>jquery-highlightsearchterms</element>
       <element>resource-plone-app-event-event-css</element>
       <element>resource-plone-app-event-event-js</element>
-      <element>resource-plone-app-discussion-javascripts-comments</element>
     </value>
   </record>
   <record name="plone.bundles/plone-logged-in.compile" interface="Products.CMFPlone.interfaces.resources.IBundleRegistry" field="compile">
@@ -3536,8 +3536,8 @@
       <element key="plone-toolbar-submenu-header-color">lighten(#000, 80%)</element>
[...]

PR #1062, opened 2015-09-22
Purge profile versions. Let plone-final depend on workflow step.

mauritsvanrees:
This signals that nothing has been installed yet, so depencies will get reapplied instead of possibly upgraded.  This could cause problems mostly in tests. Some tests were (needlessly) applying the CMFPlone:plone profile, even though there was a portal already.

And I noticed the plone-final import step was hardly the last one in the list. It need not be exactly the last, but it should at least come after all import steps that the default profile itself uses, more specifically: after workflow. Having it the wrong way around was part of the reason why you could create a Plone Site and miss the comment workflow from plone.app.discussion.

This solves issue #1041 in a better, cleaner way.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,12 @@ Changelog
 - Fix traceback style (closes `#1053`_).
   [rodfersou]
 
+- Purge profile upgrade versions from portal_setup when applying our
+  default CMFPlone:plone profile.  This signals that nothing has been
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
Jenkins: http://jenkins.plone.org/job/pull-request-5.0/380/

pbauer:
Much cleaner. Thanks.

PR #1081, opened 2015-09-24
Remove Chrome Frame from X-UA-Compatible HTTP header as it's deprecated

hvelarde:
See: https://blog.chromium.org/2013/06/retiring-chrome-frame.html


hvelarde:
http://jenkins.plone.org/view/Core/job/pull-request-5.0/398/

PR #1088, opened 2015-09-25
fix: image preview in TinyMCE editor when in modals

Gagaro:
This ensure the `resolveuid` path is at the root of the plone site.

### DIFF
--- a/Products/CMFPlone/patterns/__init__.py
+++ b/Products/CMFPlone/patterns/__init__.py
@@ -248,7 +248,7 @@ class PloneSettingsAdapter(object):
             'tiny': generator.get_tiny_config(),
             # This is for loading the languages on tinymce
             'loadingBaseUrl': '%s/++plone++static/components/tinymce-builded/js/tinymce' % generator.portal_url,  # noqa
-            'prependToUrl': 'resolveuid/',
+            'prependToUrl': '{0}/resolveuid/'.format(generator.portal_url),
             'linkAttribute': 'UID',
             'prependToScalePart': '/@@images/image/',
             'folderTypes': folder_types,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,9 @@ Changelog
   `#1041`_.
   [maurits]
 
+- Fix image preview in TinyMCE editor when in modals.
+  [Gagaro]
+
 
 5.0rc3 (2015-09-21)
 -------------------


Products/CMFPlone/patterns/__init__.py
```diff
@@ -14,6 +14,7 @@
 from plone.app.theming.utils import theming_policy
 from Products.CMFCore.utils import getToolByName
 from zope.component.hooks import getSite
+from plone import api
```
ebrehault: We are not supposed to use `plone.api` in core packages (to avoid circular dependencies).
Moreover, it looks like the portal url is already available in `generator.portal_url`.

vangheem:
So why are circular dependencies considered so bad? 

Seems crazy to me that we can't actually use plone.api in core.

davisagli:
That discussion doesn't really belong in this ticket, but...

So what happened (IIRC) in Zope 3 is, they decided to make a "zope.app.zapi" package which was sort of like plone.api. It depended on many zope.app packages, and many zope.app packages depended on it. As a result, it became difficult to use any package from Zope 3 without pulling in many of the others. That's why Plone depended on zope.app.rotterdam (a Zope 3-only skin) for a while, for example.

So I think it may not be a huge problem to use plone.api in packages that end up pulling in most of Plone already (i.e. CMFPlone and plone.app.*). But we should definitely avoid it in packages that are meant to be usable outside of Plone without pulling in Plone.

ebrehault:
tests: http://jenkins.plone.org/job/pull-request-5.0/405/

ebrehault:
@vangheem, you are right, circular dependencies per se are manageable.
But to me, in term of design, the purpose of an API is to expose the system capabilities to the external world, so that's just a layer on top of the system.
If the API is used to build the internal system, then the API becomes the system itself, and it is not an API anymore.
And if we think the existing internal methods are not satisfying to build the core, and should more similar to what plone.api offers, then the right approach is to change those internal methods, but not to use plone.api into the core.
(I agree that's a quite theorical point of view, but still I think it does matter)

ebrehault:
@Gagaro your PR breaks a test in `plone.app.z3cform`:
http://jenkins.plone.org/job/pull-request-5.0/405/testReport/junit/plone.app.z3cform.tests.test_widgets/RichTextWidgetTests/test_widget_params/

ebrehault:
Test with https://github.com/plone/plone.app.z3cform/pull/27
http://jenkins.plone.org/job/pull-request-5.0/410/

ebrehault:
@tisto I tried to run a jenkins test with 2 PRs, and I got this strange error: http://jenkins.plone.org/job/pull-request-5.0/410/console
Can't we run a 2 PR test? (or is it an error coming from somewhere else?)

jensens:
so we cannot merge this w/o red jenkins, but i merge this and directly afterwards the other PRl plone/plone.app.z3cform#27

PR #1115, opened 2015-09-30
feat: allow all TinyMCE settings to be set from control panel

Gagaro:
This allows the possibility to set all TinyMCE settings. (There are a lot of things possible: http://www.tinymce.com/wiki.php/Configuration)

This change should need an upgrade step, but I'm not sure how to add it in plone.app.upgrade (if I'd have to create a new profile or if I could use `to_500`)?

So if someone could direct me in the right direction, thanks!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,8 @@ Changelog
 5.0.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Allow all TinyMCE settings to be set from control panel.
+  [Gagaro]
 
 
 5.0 (2015-09-27)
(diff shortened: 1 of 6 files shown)


tomgross:
Cool thanks! In plone.app.upgrade you Need to Register an upgrade step in https://github.com/plone/plone.app.upgrade/blob/master/plone/app/upgrade/v50/configure.zcml with a new handler. Let's say to_501.

A changelog entry and a test would be great.

Gagaro:
Upgrade step added, I'll try to add a test now (changelog entry was already there).

Gagaro:
Test added.

tomgross:
http://jenkins.plone.org/job/pull-request-5.0/431/

PR #1138, opened 2015-10-05
upgrade to patternslib 2.0.11

vangheem:


vangheem:
tests running: http://jenkins.plone.org/job/pull-request-5.0/439/

thet:
Great! Tnx @jcbrand , @vangheem

jcbrand:
Thanks @vangheem, I see you also added the upgrade steps.

thet:
@esteele please note:
This includes mockup 2.1.0, see: https://github.com/plone/mockup/blob/master/CHANGES.rst
Pattern initialization is slightly different now and actually incompatible with the old way. IMO, it's important to get this into the next Plone release, so that people building patterns do it the new way.

I'm asking myself, if this change justifies a Plone 5.1 or just a 5.0.1 release?

jcbrand:
@thet, @esteele Older pattens should (and do still work) due to the `mockup-patterns-base` mock that we added (due to @vangheem's foresight). So, while the new way is different, this is not a breaking change for old patterns (see for example the toolbar, which hasn't been updated).

thet:
@jcbrand thanks for clarification. so no probs for the next release be 5.0.1

PR #1139, opened 2015-10-05
fix issue where clicking tabs would cause odd scroll movement

vangheem:


vangheem:
tests running here: http://jenkins.plone.org/job/pull-request-5.0/440/

PR #1174, opened 2015-10-16
Ensure cooked CSS/JS resources are treated as "stable resources" for caching

alecpm:
Use the ++unique++ traverser instead of query string to make unique resource urls.  This has the advantage of marking the resource as stable for caching purposes.  Fix the custom traverser to properly pass persistent resource directories along to the ++unique++ traverser.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@ Changelog
 
 - Add syndication for plone.app.contenttypes collections.
   [do3cc]
+- Use unique traverser for stable resources to set proper cache headers.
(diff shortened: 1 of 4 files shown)


vangheem:
build running here: http://jenkins.plone.org/job/pull-request-5.0/469/

gforcada:
@alecpm tests did pass but the branch can not be merged right away, could you rebase it?

alecpm:
I'm pretty sure it was just the changelog entry, but it's been rebased now.  Thanks.

Issue #1175, opened 2015-10-16
Error if removing all entries from 'plone.types_use_view_action_in_listings'


tomgross:
If all entries from "plone.types_use_view_action_in_listings" in the control panel are removed the following error occurs in Plone 5:

```
2015-10-16 11:51:34 ERROR Zope.SiteErrorLog 1444989094.60.631693912663 http://localhost:8008/Plone9/portal_registry/@@configuration_registry
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module plone.app.registry.browser.records, line 60, in __call__
  Module Products.Five.browser.pagetemplatefile, line 125, in __call__
  Module Products.Five.browser.pagetemplatefile, line 59, in __call__
  Module zope.pagetemplate.pagetemplate, line 132, in pt_render
  Module five.pt.engine, line 93, in __call__
  Module z3c.pt.pagetemplate, line 163, in render
  Module chameleon.zpt.template, line 258, in render
  Module chameleon.template, line 191, in render
  Module chameleon.template, line 171, in render
  Module 590201518ade929608918d3098ac9685.py, line 1192, in render
  Module c0ec7dfa4949e7d97f32a018369e2b28.py, line 282, in render_master
  Module 531b21e8bfc2c61bbbaa26b7ba88abfe.py, line 800, in render_master
  Module z3c.pt.expressions, line 71, in render_content_provider
  Module zope.viewlet.manager, line 112, in update
  Module zope.viewlet.manager, line 118, in _updateViewlets
  Module plone.app.layout.viewlets.social, line 60, in update
  Module Products.CMFPlone.browser.syndication.adapters, line 240, in link
TypeError: argument of type 'NoneType' is not iterable

 - Expression: "provider:plone.htmlhead"
 - Filename:   ... egg/Products/CMFPlone/browser/templates/main_template.pt
 - Location:   (line 25: col 36)
 - Source:     ... 
                                         ^
 - Arguments:  repeat: {...} (0)
               template: <ViewPageTemplateFile - at 0x7fef3d5ad510>
               views: <ViewMapper - at 0x7fef3cfc7ed0>
               modules: <instance - at 0x7fef4f1c6e60>
               args: <tuple - at 0x7fef5b75e050>
               here: <ImplicitAcquisitionWrapper portal_registry at 0x7fef3d54e910>
[...]

PR #1176, opened 2015-10-16
Added missing_value parameter to list and tuple fields

tomgross:


tomgross:
Fixes #1175 but needs to reread registry on upgrade. Upgrade is not included in PR.

PR #1189, opened 2015-10-17
Add patternslib autofocus

timitos:


vangheem:
See my comment here: https://github.com/plone/Products.CMFPlone/pull/1191#issuecomment-148964535

timitos:
I found out that patternslibs autofocus pattern is quite different to what mockup-formautofocus should do. So this PR does not make sense any more.

PR #1193, opened 2015-10-18
Use formautofocus pattern.

timitos:
After pull request for plone/mockup#564 is merged. This change will make the autofocus work again.

### DIFF
--- a/Products/CMFPlone/browser/templates/search.pt
+++ b/Products/CMFPlone/browser/templates/search.pt
@@ -29,7 +29,7 @@
     <form name="searchform"
           action="@@search"
           role="search"
-          class="searchPage enableAutoFocus"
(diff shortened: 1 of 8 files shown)


vangheem:
testing: http://jenkins.plone.org/job/pull-request-5.0/479/

vangheem:
@gforcada @tisto any idea what is going on? Failed immediately and no error output.

timitos:
It looks like the '/' in the branch name caused the problem. Should i maybe rename the branch and create an new pull request?

gforcada:
@timitos is right, the problem is on the regular expression we use to update the branch... if you don't mind renaming it (dashes are fine and used by other contributors).

We can also update the sed expression to use something else, a # for example.

timitos:
New Pull Request #1205

Issue #965, opened 2015-09-11
Remove SecureMailHost


mauritsvanrees:
Products.SecureMailHost host not really been used since Plone 4.0

There are three places that still mention it:
- https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/patches/securemailhost.py says at the top: "This module provides backwards compatiblity for products using the
  SecureMailHost API.  It should be removed entirely for Plone 5.0."
- https://github.com/plone/plone.app.testing/blob/master/plone/app/testing/layers.py#L62 loads its zcml.
- https://github.com/plone/plone.app.upgrade/blob/master/setup.py#L75 has in its `install_requires`:         `'Products.SecureMailHost', # For migration only, when can we remove this?`

Note that there were already some tries to get rid of this, but this failed, perhaps because too many patches were removed at once. See issue #109. So it may be more tricky than it sounds.

Note also that I have just move SecureMailHost out of the plone subversion collective into github, in case anyone needs it: https://github.com/collective/Products.SecureMailHost

thet:
+1 for removing this package. as i see, this
1) needs help,
2) needs an plone.app.upgrade story plan,
3) needs a fix for #1208 too and
4) is a PLIP as it changes the API, even if there was an agreement on that some time ago.

tkimnguyen:
@thet item 3 above has been fixed (#1208)

thet:
Great! PLIP not needed - i just saw in the cmfplone patch that there was already a deprecation warning. Not sure about the upgrade thing - if the product didn't install some persistent tool, we can just remove all the securemailhost references.

jensens:
can we close this one?

thet:
First, we need a PR which removes what @mauritsvanrees mentioned, not?

mauritsvanrees:
plone.app.dexterity uses `secureSend` in documentation a few times. I have removed that: https://github.com/plone/plone.app.dexterity/commit/48da045a46846f9c3c33d8cb7a67b4b97ff7d0c5

mauritsvanrees:
I grepped through code in community packages that I have locally, and only found plone.app.dexterity (fixed) and collective.contentrule.mailtorole (issue created) that used secureSend.
Well, a few more that use secureSend as fallback when they see it is needed, like an [older version of collective.watcherlist](https://github.com/collective/collective.watcherlist/blob/1.2/collective/watcherlist/mailer.py#L13) and collective.confirmableforms.

[...]

PR #1208, opened 2015-10-22
set Reply-to address 

tkimnguyen:
so that replying to an email sent by the contact-info form does the expected thing

### DIFF
--- a/Products/CMFPlone/browser/contact_info.py
+++ b/Products/CMFPlone/browser/contact_info.py
@@ -67,12 +67,13 @@ class ContactForm(form.Form):
 
         try:
             # This actually sends out the mail
-            host.send(
+            host.secureSend(
                 self.generate_mail(data, encoding),
                 send_to_address,
                 from_address,
                 subject=subject,
-                charset=encoding
+                charset=encoding,
+                **{'Reply-to':data['sender_from_address']}
             )
         except (SMTPException, RuntimeError), e:
             log.error(e)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,9 +8,13 @@ Changelog
 5.0.1 (unreleased)
 ------------------
 
+- set Reply-to address in contact-info emails so you can reply to them
+  [tkimnguyen]
+
 - prevent breaking Plone when TinyMCE JSON settings fields contain
   invalid JSON
   [petschki]
+
 - Fix #1199 prevent throwing error with mis-configured bundle
   [vangheem]
 


vangheem:
tests running: http://jenkins.plone.org/job/pull-request-5.0/502/

Changelog!

Products/CMFPlone/browser/contact_info.py
```diff
@@ -67,12 +67,13 @@ def send_message(self, data):
 
         try:
             # This actually sends out the mail
-            host.send(
+            host.secureSend(
```
mauritsvanrees: Why the change to `secureSend`? This method  from SecureMailHost was already deprecated in Plone 4.  It is still there, but we want to get rid of it. See issue #965.

vangheem: because send doesn't allow setting headers. It's unfortunate the API doesn't allow setting headers of the message.

mauritsvanrees: Ah, right. I guess alternatively we could set `message['Reply-to']` for the same effect.

davisagli: Here's the helper I'm using to send mail in one project:

```
from email.Header import Header
from email.mime.multipart import MIMEMultipart
from email.mime.text import MIMEText
import email


def send_mail(portal, subject, message, mfrom=None, mto=None, attachments=None):
    # encode message if necessary
    if isinstance(message, unicode):
        message = message.encode('utf8')

    # add plain text version of html email
    if message.startswith('<html'):
        portal_transforms = getToolByName(portal, 'portal_transforms')
        text = str(portal_transforms.convert('html_to_web_intelligent_plain_text', message))
        msg = MIMEMultipart('alternative')
        msg.attach(MIMEText(text, 'plain', 'utf-8'))
        msg.attach(MIMEText(message, 'html', 'utf-8'))
    else:
        msg = email.message_from_string(message)
        msg.set_charset('utf-8')

    # add attachments (pass in as list of instances of MIMEImage, etc)
    if attachments:
        text = msg
        msg = MIMEMultipart('related')
        msg.attach(text)
        for attachment in attachments:
            msg.attach(attachment)

    # Set Reply-To header
    # (From header will be the portal's From address with the sender's real name)
    realname = None
    if mfrom is not None:
        realname, replyaddr = email.utils.parseaddr(mfrom)
        msg['Reply-To'] = Header(mfrom, 'utf-8')

    if not realname:
        realname = portal.email_from_name
    mfrom = email.utils.formataddr((realname, portal.email_from_address))

    # Send to portal email address if no recipient was specified
    if mto is None:
        mto = mfrom

    # Actually send
    mailhost = portal.MailHost
    mailhost.send(
        msg, subject=subject, mfrom=mfrom, mto=mto,
        immediate=True, charset='utf-8')
```

mauritsvanrees: I have updated the code to add a Header instead of using secureSend. See 0eac3fef65a82c7ab2171585bf0c53d1c4af9748

PR #1210, opened 2015-10-25
be able to stub JavaScript modules to prevent including multiple times

vangheem:


vangheem:
build running: http://jenkins.plone.org/job/pull-request-5.0/510/

Products/CMFPlone/interfaces/resources.py
```diff
@@ -96,3 +96,12 @@ class IBundleRegistry(zope.interface.Interface):
     develop_css = schema.Bool(
         title=_(u'Develop CSS'),
         default=False)
+
+    stub_js_modules = schema.List(
+        title=_(u'Stub JavaScript modules'),
+        description=_(u'Define list of modules that will be not be defined empty '
```
thet: "that will be ~~not~~ defined empty" ... the "not" shouldn't be there, not?

vangheem: correct. I'll fix.

ebrehault:
Hi @vangheem,
So as far as I understand, when we declare a resource as stub_js_modules in a bundle, we say that's a needed dependency but it won't be added to the bundle because we know another bundle will provide it (and our bundle depends on this other bundle).
That's nice.
But, apparently, it does not solve the bundle merging problem: on a given Plone site, we will always have the Plone core bundles + the diazo theme bundle + 1 bundle by installed add-on. So it might result in many many js/css to load in every page of our site, which is very bad (even if those bundles will now be smaller thanks to the present PR).

Do you have any idea about how we could merge all our bundles in one JS bundle + one CSS bundle ?

thet:
my advice at the JS training was, that integrators manually fix this by adding their dependencies to one of the plone-core bundles (plone or plone-logged-in). integrators also had to optimize the resource registry in Plone 4.x.
but i agree, if we could merge all automatically, that would be nice.

ebrehault:
@thet yes for sure, we can do that manually when integrating, but if we consider the regular "Add a new add-on to Plone" story, with Plone 4 it used to be:
- add egg in buildout, run it and restart,
- install module from Site setup.

Now it is broken (either we just do the same and we get a long list of css/js to load in all our pages, either we need to tune bundles manually).

It would be nice if we could restore this story.

That's something we discussed at the Plone conf, and it looked complex to achieve, but with this PR, it seems easier, as we just need to concatenate all the bundles together (I guess).

vangheem:
Bundles are meant to be pre-built.

Since we moved from linear file inclusion to dependency management, it's more difficult to auto merge files.

I wouldn't be against auto-combining.

@bloodbare thoughts?

FYI, everyone who is paying attention. It seems everyone assumes I'm the brains behind the resource registries as I always get cc'd on conversations. Ramon built and designed most of it. I'd recommend using him as the more authoritative voice on this sometimes. I'm fine being cc'd but don't leave @bloodbare out of it.

PR #1229, opened 2015-11-06
build resources

Gagaro:
16 days since last compilation.

What's the `define("/home/gagaro/plone/sources/buildout.coredev/src/Products.CMFPlone/Products/CMFPlone/static/plone.js", function(){});` used for BTW ?


jensens:
since this is 10days old this one seems to be outdated now. Would be cool to have a new PR with fresh compiled stuff.

Gagaro:
Already done there https://github.com/plone/Products.CMFPlone/commit/8542c849cc7026a28c79707296f02dab5ec8e077

PR #1233, opened 2015-11-07
Add icon fallback for addons in Site Setup (closes #1232)

### LINKED ISSUE #1232: Add icon fallback for addons in Site Setup
Site setup configlets in Plone 5 have different icon implementation, using autogenerated css classes and preferably icon fonts. It'd be nice to have a fallback in case updated or even new addons don't register their own icons.

davilima6:
This is the gear icon, using the same character code of the Maintenance icon.

The same icon is used as placeholder inside Fontello for yet to be selected icons such as ZMI, Configuration and Resource Registries but I preferred Maintenance code because I thought it'd the more likely to remain with the generic settings icon.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,9 @@ New:
   customization.
   [tomgross]
 
+- Add icon fallback for addons in Site Setup
+  [davilima6]
+
 Fixes:
 
 - Fixed white space pep8 warnings.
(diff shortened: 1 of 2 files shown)


jensens:
keep #1236 in mind

davilima6:
I am, and thanks for the Discourse link. Still I think this should be merged. I believe addon authors will be happy not to worry with an icon until really needed. I mean, even Plone 5.0 couldn't timely generate specific icons for all vanilla configlets, so a fallback would be useful imo.

This can then be overruled when `icon_expr` is provided. When that implementation is once again available, we can make the pseudo-selector that holds the icon vanish with a lower z-index and add a background-color for the container that holds the image icon. There should be other ways too, depending on how `icon_expr` is reborn.

davilima6:
Thanks!

jensens:
btw, this need to compile resources to activate it, right?

davilima6:
Yep, or test it with Development Mode.

PR #1246, opened 2015-11-17
Minor fix for upgrade from old 4.0 instance

frapell:
When upgrading a 4.0 site, I get:

AttributeError: 'JSRegistryNodeAdapter' object has no attribute 'registry'

Declaring it as a class attribute should fix this

### DIFF
--- a/Products/CMFPlone/resources/exportimport/resourceregistry.py
+++ b/Products/CMFPlone/resources/exportimport/resourceregistry.py
@@ -40,7 +40,8 @@ def importResRegistry(context, reg_id, reg_title, filename):
 class ResourceRegistryNodeAdapter(XMLAdapterBase):
 
     resource_blacklist = set()
-
+    registry = None
+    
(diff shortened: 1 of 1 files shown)


gforcada:
Is that intended to be merged for 5.0 or on 4.3?

I started a job for 5.0 here: http://jenkins.plone.org/job/pull-request-5.0/555/

frapell:
Don't know for sure what is best... maybe both ?

gforcada:
@frapell this pull request will be merged on master, so 5.0, you will have to create a new pull request for 4.3, and as the tests passed: merging. Could you add a changelog entry please? Directly on master

frapell:
@gforcada Thanks, thought it was too small of a change to merit a changelog entry.

I'm now realizing that this is a Plone 5 thing, I will have to test migrating to 4.3 and see if it fails. If so, I will add a fix in 4.3.x as well.

PR #1260, opened 2015-11-23
Don't depend on and install plone.app.widgets. plone.app.z3cform does…

thet:
… it for us.


thet:
http://jenkins.plone.org/job/pull-request-5.0/584/

Issue #1132, opened 2015-10-03
Missing @@dashboard link in Personal Tools


davilima6:
It seems to have been forgotten during the migration to the toolbar.

dtordable:
There's no link to the dashboard but /@@dashboard works as usual...

davilima6:
Strangely enough there's no reference to that action in Plone 4.3: https://github.com/plone/Products.CMFPlone/blob/4.3.x/Products/CMFPlone/profiles/default/actions.xml#L302

I mean, would be that the only Personal Tools link inserted dynamically (via JS)?

davilima6:
Related to https://community.plone.org/t/does-plone-5-still-have-dashboard/1235

PR #1262, opened 2015-11-24
re-add missing dashboard menu item as per #1132

tkimnguyen:
for #1132

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Changelog
 
 New:
 
+- Re-add dashboard personal menu item 
+  [tkimnguyen]
+
 - Add option to show/hide thumbs in site-controlpanel
   https://github.com/plone/Products.CMFPlone/issues/1241
   [fgrcon]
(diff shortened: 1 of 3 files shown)


davilima6:
Thanks, @tkimnguyen. I don't think it's a blocker but it'd be nice to add or update tests so we avoid regressing again. Any idea on where those would go? Couldn't find.

tkimnguyen:
@davilima6 ah, good idea but I don't know either

davisagli:
Since this is a change to a GenericSetup profile (i.e., something that will make a persistent change to the database), we also need an upgrade step.

davisagli:
Re tests, I would add a simple functional test in this package or in plone.app.portlets which logs in and confirms that there's a Dashboard link to click.

tkimnguyen:
@davisagli would something like this in tests/testActionsTool.py suffice?

```
    def testDashboardActionExists(self):
        actions = self.actions.listActions()
        self.assertTrue('dashboard' in [a.id for a in actions])
```

I was looking for a good place to put an upgrade step but couldn't find one in this package - any suggestions?

davisagli:
That'll work...it doesn't seem quite like the right place, since the rest of the tests there are about Plone's additional functionality for the CMF action tool rather than tests of how the Plone profile configures actions, but I don't see an obvious better place to put it either.

Upgrade steps go in plone.app.upgrade.

jensens:
this was fixed afaict

Issue #1226, opened 2015-11-05
Cleanup and rework: contenttype-icons and showing thumbnails for images/leadimages in listings ...


fgrcon:
Replaces https://github.com/plone/Products.CMFPlone/issues/1151, https://github.com/plone/Products.CMFPlone/issues/1074  and adresses  https://github.com/plone/Products.CMFPlone/issues/1185 and may be some more.
**Detailed documentation: https://github.com/fgrcon/Products.CMFPlone/blob/documented/docs/icons_in_plone5.rst

I will reopen pull requests when finished (almost done) and tested (as far as I can).
**PRs are interdependend**

@jensens:
you suggested:
nterdependent pull request are not directly possible by now, this was discussed here plone/jenkins.plone.org#126 . But you can make a branch and PR (which is only for testing, not for merge) in buildout.coredev with adjusted sources.cfg - this can then be tested with the PR>mechanism. Its a workaround , but well...

**I created a branch in coredev (see above) but how to start Jenkins with that? I  created  https://github.com/plone/buildout.coredev/pull/140 -> http://jenkins.plone.org/job/pull-request-5.0/529/ which obviously did not fetch sources from fgrcon....**

gforcada:
@fgrcon interdependent pull requests can be tested now: just put all of them (one per line) on the jenkins job for either 5.0 or 4.3.

Have fun!

fgrcon:
@gforcada  Great! thanks...
http://jenkins.plone.org/job/pull-request-5.0/530/

hvelarde:
@fgrcon IMO, is a bad idea to include different issues on the scope of one PR; this could lead easily to errors and could delay the merging of it.

fgrcon:
@hvelarde   you are absolutely right on that, but this really was not my own  decision. I just wanted to contribute  a small feature,  but i found a  lot of  dependend  issues  .... and if you think you can handle  this better: i will suport you as much  as i can,  No need to merge  my proposals - but  just replace it with other ones ...

fgrcon:
sorry  i dn't close this intentionally,(one click to much ?)

hvelarde:
sorry, I have no time, neither knowledge to do what you're doing; I'm just giving my opinion.

something is rotten in the state of Plone if the contribution of a small feature turned into a nightmare of dependent issues; take some fresh air and think about it again...

jensens:
[...]

PR #1271, opened 2015-12-01
fix: getIcon for archetypes

Gagaro:
Archetypes show a broken img as icon since https://github.com/plone/Products.CMFPlone/issues/1226

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -256,8 +256,16 @@ def is_default_page(obj):
 
 @indexer(Interface)
 def getIcon(obj):
-    """Make sure we index icon relative to portal"""
-    return obj.getIcon(True)
+    """
+    geticon redefined in Plone > 5.0
+    see https://github.com/plone/Products.CMFPlone/issues/1226
+
+    reuse of metadata field,
+    now used for showing thumbs in content listings etc.
+    when obj is an image or has a lead image
+    or has an image field with name 'image': true else false
+    """
+    return bool(getattr(obj.aq_base, 'image', False))
 
 
 @indexer(Interface)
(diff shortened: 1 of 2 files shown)


jensens:
may you add a note to the changelog please?

Products/CMFPlone/CatalogTool.py
```diff
@@ -256,8 +256,18 @@ def is_default_page(obj):
...
+    """
+    geticon redefined in Plone > 5.0
+    see https://github.com/plone/Products.CMFPlone/issues/1226
+
+    reuse of metadata field,
+    now used for showing thumbs in content listings etc.
+    when obj is an image or has a lead image
+    or has an image field with name 'image': true else false
+    """
+    if obj.aq_base.image:
```
jensens: if the ``image`` attribute does not exist an ``AttributeError`` would be raised, is this intended? If not I'd change to

    return bool(getattr(obj.aq_base, 'image', False))

Gagaro: An exception in an indexer would just give an empty value, which is okay in this case. I copied this from https://github.com/plone/plone.app.contenttypes/commit/3cfc044ad8061907967fe84b2c638461e93e7ce5.

I like yours better though, I'll update it.

Gagaro:
Changelog added and indexer updated

Products/CMFPlone/CatalogTool.py
jensens: ah, ok, usally if I see something like 

    if (some condition):
        return True
   return False

I tend to normalize to 

    return bool(condition)

even bool() is not needed always, but for an indexer its good to return explicit values.

jensens:
lets get jenkins opinion :)  http://jenkins.plone.org/job/pull-request-5.0/606/

Issue #1293, opened 2015-12-21
Site Administrator can not reset any user's password in @@usergroup-userprefs if there is a member of Administrators group


fulv:
Note: I labeled this as a regression, because it does not happen in Plone 4.3.7.

  1. Create a Plone 5.0 site
  2. Add a user and add them to the *Administrators* group
  3. Add another user and make them a Site Administrator (does not matter whether through direct role or membership in the Site Administrators group)
  4. Log in as the Site Administrator
  5. Go to @@usergroup-userprefs
  6. Check the reset password box for the Site Administrator user, click Save

Result:
See traceback below.

The reason for this problem is in `Products.CMFPlone-5.0-py2.7.egg/Products/CMFPlone/controlpanel/browser/usergroups_usersoverview.py(175)manageUser()`:

```
                roles = user.get('roles', [])
                if not self.is_zope_manager:
                    # don't allow adding or removing the Manager role
                    if ('Manager' in roles) != ('Manager' in current_roles):
                        raise Forbidden
```

The `roles` array for the Manager user does not contain `Manager`, while `current_roles` does.  That's because in the form that user's row has an image under the Manager column, but no checkbox, so the `Manager` role is not added to the POST.

2015-12-21 12:47:36 ERROR Zope.SiteErrorLog 1450730856.150.395936448296 http://localhost:8086/Plone2/@@usergroup-userprefs
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module Products.PDBDebugMode.runcall, line 70, in pdb_runcall
  Module ZPublisher.Publish, line 48, in call_object
  Module Products.CMFPlone.controlpanel.browser.usergroups_usersoverview, line 40, in __call__
  Module Products.CMFPlone.controlpanel.browser.usergroups_usersoverview, line 175, in manageUser
Forbidden
> /Users/fulvio/Downloads/Plone-5.0/buildout-cache/eggs/Products.CMFPlone-5.0-py2.7.egg/Products/CMFPlone/controlpanel/browser/usergroups_usersoverview.py(175)manageUser()
-> raise Forbidden

fulv:
(The workaround is to remove the Manager users from the Managers group, and give them the Manager role directly.)

fulv:
Also, the steps above are contrived for brevity, but of course the real problem is that the Site Administrator can not reset **any other user's password**, either, not just their own.

[...]

PR #1296, opened 2015-12-22
Fixed Forbidden error in users and groups overview.

mauritsvanrees:
This is when using it as Site Administrator.  This could happen when
there are users that inherit the Manager role from the Administrators
group.

Fixes issue #1293.


mauritsvanrees:
http://jenkins.plone.org/job/pull-request-5.0/643/

PR #1300, opened 2015-12-23
Plip 1277 meta bundles generation

### LINKED ISSUE #1277: PLIP: Meta bundles generation
## Abstract
In Plone 5, resources bundles are loaded separately in all the portal pages. This PLIP proposes to be able to merge them in meta bundles.

## Motivation
When a Plone site uses 10 add-ons, all the pages will contain 10 extras SCRIPT tags and 10 extra LINK tags (plus the default ones provided by Plone). This is very bad for performances, we need to be able to aggregate them.

### Disclaimer
Plone integrators and Plone add-on developers need a simple story:

- When we develop an add-on, it should be easy to declare our specific bundles so they are automatically included in such or such meta bundle.
- When we install an add-on in our site, it should not require to compile bundles or other stuffs.

Consequently, we do not want to interact with the existing bundling mechanism as it requires all the front-end dev tooling.

The objective here is just to concatenate existing JS or CSS bundles together, which should be a basic string processing operation we can trigger easily from a GenericSetup step.

### Proposal
`Products.CMFPlone` will define 2 meta bundles:

- `default` (loaded everytime)
- `logged-in` (loaded if the user is logged in)

Each meta bundle will produces 2 outputs (one JS and one CSS).

Note: I do not think we need to be able to declare other meta bundles. If we need to provide some specific resources in a given page or in a given condition, regular bundles should be satisfying, using the  `expression` property, or using `add_bundle_on_request` http://docs.plone.org/adapt-and-extend/theming/resourceregistry.html#adding-or-removing-bundles-from-a-request .

Bundles will have a new property named `merge_with` which possible values will be `default`, `logged-in` or nothing:

- If `merge_with` is defined, the bundle is aggregated in the corresponding meta bundle, and its `expression` property will be ignored.
- If `merge_with` is nothing, the bundle is published directly in the site pages (just like it does at the moment) and the `expression` property is applied.
- in Development mode, meta bundles are not published, all bundles are published directly.

The meta bundle generation is done by a GenericSetup step.
As bundles can be defined or modified TTW, we will also provide a "Merge bundles for production" button in the Resource registry that will re-generate the meta bundles.

### Implementation
Meta bundles are stored in the "persistent" IResourceDirectory, just like as the current bundle cooking does at the moment.
They are produced by aggregating together all the bundle contents in a single string.

## Deliverables
- a new version of Products.CMFPlone,
- documentation.

## Risks
N/A

## Participants
Eric Bréhault

ebrehault:
Fix #1277

### DIFF
--- a/Products/CMFPlone/resources/browser/combine.py
+++ b/Products/CMFPlone/resources/browser/combine.py
@@ -1,0 +1,101 @@
+from Acquisition import aq_base
+from datetime import datetime
+from plone.registry.interfaces import IRegistry
+from plone.resource.file import FilesystemFile
+from plone.resource.interfaces import IResourceDirectory
+from Products.CMFPlone.interfaces import IBundleRegistry
+from Products.CMFPlone.interfaces.resources import (
+    OVERRIDE_RESOURCE_DIRECTORY_NAME,
+)
+from StringIO import StringIO
+from zope.component import getUtility
+
+PRODUCTION_RESOURCE_DIRECTORY = "production"
+
+
+def get_production_resource_directory():
+    persistent_directory = getUtility(IResourceDirectory, name="persistent")
+    container = persistent_directory[OVERRIDE_RESOURCE_DIRECTORY_NAME]
+    production_folder = container[PRODUCTION_RESOURCE_DIRECTORY]
+    timestamp = production_folder.readFile('timestamp.txt')
+    return "%s/++unique++%s" % (
+        PRODUCTION_RESOURCE_DIRECTORY, timestamp)
+
+
+def get_resource(context, path):
+    resource = context.unrestrictedTraverse(path)
+    if isinstance(resource, FilesystemFile):
+        (directory, sep, filename) = path.rpartition('/')
+        return context.unrestrictedTraverse(directory).readFile(filename)
+    else:
+        if hasattr(aq_base(resource), 'GET'):
+            # for FileResource
+            return resource.GET()
+        else:
+            # any BrowserView
+            return resource()
+
+
+def write_js(context, folder, meta_bundle):
+    registry = getUtility(IRegistry)
+    resources = []
+
+    # default resources
+    if meta_bundle == 'default' and registry.records.get(
+        'plone.resources/jquery.js'
+    ):
+        resources.append(get_resource(context,
+            registry.records['plone.resources/jquery.js'].value))
+        resources.append(get_resource(context,
+            registry.records['plone.resources.requirejs'].value))
+        resources.append(get_resource(context,
+            registry.records['plone.resources.configjs'].value))
+
+    # bundles
+    bundles = registry.collectionOfInterface(
+        IBundleRegistry, prefix="plone.bundles", check=False)
+    for bundle in bundles.values():
+        if bundle.merge_with == meta_bundle:
+            resources.append(get_resource(context, bundle.jscompilation))
+
+    fi = StringIO()
+    for script in resources:
+        fi.write(script + '\n')
+    folder.writeFile(meta_bundle + ".js", fi)
+
+
+def write_css(context, folder, meta_bundle):
+    registry = getUtility(IRegistry)
+    resources = []
+
+    bundles = registry.collectionOfInterface(
+        IBundleRegistry, prefix="plone.bundles", check=False)
+    for bundle in bundles.values():
+        if bundle.merge_with == meta_bundle:
+            resources.append(get_resource(context, bundle.csscompilation))
+
+    fi = StringIO()
+    for script in resources:
+        fi.write(script + '\n')
+    folder.writeFile(meta_bundle + ".css", fi)
+
+
+def combine_bundles(context):
+    persistent_directory = getUtility(IResourceDirectory, name="persistent")
+    if OVERRIDE_RESOURCE_DIRECTORY_NAME not in persistent_directory:
+        persistent_directory.makeDirectory(OVERRIDE_RESOURCE_DIRECTORY_NAME)
+    container = persistent_directory[OVERRIDE_RESOURCE_DIRECTORY_NAME]
+    if PRODUCTION_RESOURCE_DIRECTORY not in container:
+        container.makeDirectory(PRODUCTION_RESOURCE_DIRECTORY)
+    production_folder = container[PRODUCTION_RESOURCE_DIRECTORY]
+
+    # store timestamp
+    fi = StringIO()
+    fi.write(datetime.now().isoformat())
+    production_folder.writeFile("timestamp.txt", fi)
+
+    # generate new combined bundles
+    write_js(context, production_folder, 'default')
+    write_js(context, production_folder, 'logged-in')
+    write_css(context, production_folder, 'default')
+    write_css(context, production_folder, 'logged-in')
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,10 @@ New:
 - Add barceloneta theme path in less configuration.
   [Gagaro]
 
+- Merge JS and CSS bundles into meta-bundles to reduce the number of requests
+  when loading a page (PLIP #1277)
+  [ebrehault]
+
 Fixes:
 
 - Get ``email_from_name`` from the mail settings registry.
--- a/Products/CMFPlone/controlpanel/browser/resourceregistry.py
+++ b/Products/CMFPlone/controlpanel/browser/resourceregistry.py
@@ -208,12 +208,13 @@ class ResourceRegistryControlPanelView(RequireJsView):
     def save_registry(self):
         req = self.request
 
-        self.update_registry_collection(
-            IResourceRegistry, "plone.resources",
-            json.loads(req.get('resources')))
-        self.update_registry_collection(
-            IBundleRegistry, "plone.bundles",
-            json.loads(req.get('bundles')))
+        if req.get('resources') and req.get('bundles'):
+            self.update_registry_collection(
+                IResourceRegistry, "plone.resources",
+                json.loads(req.get('resources')))
+            self.update_registry_collection(
+                IBundleRegistry, "plone.bundles",
+                json.loads(req.get('bundles')))
 
         if self.request.form.get('development', '').lower() == 'true':
             self.registry['plone.resources.development'] = True
--- a/Products/CMFPlone/exportimport/configure.zcml
+++ b/Products/CMFPlone/exportimport/configure.zcml
@@ -100,6 +100,14 @@
       description="Import CMFDiffTool settings">
   </genericsetup:importStep>
 
+  <genericsetup:importStep
+      name="combine-bundles"
+      title="Bundle combination"
+      description="Combine JS/CSS bundles together"
+      handler="Products.CMFPlone.resources.exportimport.bundles.combine">
+    <depends name="plone.app.registry" />
+  </genericsetup:importStep>
+
   <genericsetup:exportStep
       name="content"
       title="Content"
--- a/Products/CMFPlone/interfaces/resources.py
+++ b/Products/CMFPlone/interfaces/resources.py
@@ -1,6 +1,8 @@
 import zope.interface
 import zope.component
 from zope import schema
+from zope.schema.vocabulary import SimpleTerm
+from zope.schema.vocabulary import SimpleVocabulary
 
 from Products.CMFPlone import PloneMessageFactory as _
 
@@ -60,13 +62,15 @@ class IBundleRegistry(zope.interface.Interface):
     expression = schema.ASCIILine(
         title=_(u"Expression to render"),
         description=_(
-            u"In case its a bundle we can have a condition to render it"),
+            u"In case its a bundle we can have a condition to render it (it "
+            "does not apply if the bundle is merged)."),
         required=False)
 
     conditionalcomment = schema.ASCIILine(
         title=_(u"Conditional comment"),
         description=_(
-            u"In case you want to render this resource on conditional comment"),
+            u"In case you want to render this resource on conditional comment "
+            "(it does not apply if the bundle is merged)."),
         required=False)
 
     resources = schema.List(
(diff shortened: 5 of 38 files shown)


jensens:
:+1:

Products/CMFPlone/resources/browser/combine.py
```diff
@@ -0,0 +1,103 @@
...
+        return '/* Content of %s */' % path
+    if isinstance(resource, FilesystemFile):
+        (directory, sep, filename) = path.rpartition('/')
+        return context.unrestrictedTraverse(directory).readFile(filename)
+    else:
+        try:
+            return resource.GET()
+        except:
+            return resource()
+
```
thet: Please leave a comment, what the difference of resource.GET() nad resource() is.

ebrehault: Done https://github.com/plone/Products.CMFPlone/commit/dba346005111ca999edf6f7148b2cf075f79080b
I have also replace the brutal try/except with a if/else.

thet:
Great work!

But I have some concerns to be discussed:

Wouldn't it have been possible to merge bundles with existing bundles instead of creating new meta-bundles? That would have allowed better flexibility - so one could create 3 or more production bundles, where other bundles are merged into - all having different set of rendering conditions.

Another thing is that the "expressions" become useless in production mode, don't they? How do they have an effect for plone-logged-in, which has an expression checking for logged in user but is merged into the logged-in meta bundle?

Same for conditional comments. They need also the "does not apply if the bundle is merged".  But this field might get obsolete soon (IE < 11 is dead!).

With the Plone < 5 approach where merging was automatically done, based on matching expressions and conditional comments, these issues would be gone.

ebrehault:
(maybe this should be discussed in the PLIP issue itself https://github.com/plone/Products.CMFPlone/issues/1277, but anyway...)

@thet I think the Plone < 5 bundling was too complex because of the conflicting usage of ordering and expressions: any resource having a different expression than the previous one resulted in a break in bundling, and at the end we got a long list of small bundles instead of one or two big ones, unless you manage the ordering very precisely (and everytime you install a new add-on, you need to re-order everything again).

So I propose a simple approach:
- for a big majority of resources, we replace the expression feature with only 2 "hard-coded" expressions: default and logged-in
- anything which needs a more precise matching will not be merged and will be injected into pages either using the expression mechanism, either using `add_bundle_on_request`

I think it is a more efficient approach but still very flexible.

Note: regarding conditional comments, you are right, we need to mention they won't be applied when merged. Thats' fixed.

thet:
@ebrehault - I thought I'll discuss this on the concrete implementation, therefore I wrote it down here.

I agree about the complexity of the previous approach with ordering. Although with declared dependencies we might could do a better ordering.

I'm OK with your approach, because it's flexible enough for the majority of usecases.

You say: "anything which needs a more precise matching will not be merged" - but only, if you do not set the "merge_with" parameter. If you do, it will be merged, regardless of the condition, no?

ebrehault:
> You say: "anything which needs a more precise matching will not be merged" - but only, if you do not set the "merge_with" parameter. If you do, it will be merged, regardless of the condition, no?

yes absolutely, the `merge_with` property has the priority over condition and expression.

ebrehault:
When the bundles are merged, I get an error in my browser:
 be removed.Instead, use pat-base and explicitly set parser to 'mockup' when calling extend.
```Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/bundle6/++plone++production/++unique++2016-01-27T20:11:01.825026/contenttypes-sprite.png Failed to load resource```

I cannot find this  `contenttypes-sprite.png` anywhere in Plone, but it is mentionned in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/static/plone-legacy-compiled.css
How can we refresh (=re-compile) this plone-legacy-compiled.css ?
@bloodbare any hints ? (you are the last person who committed this file :) )

**Comment by bloodbare** [comment]
As I remember the plone-legacy-compiled.css from CMFPlone was copied manually as its not used as its overwritten in the ResourceDirectory on lunch/access time. It was there for mockup propouses.

**Comment by ebrehault** [comment]
The robot tests are broken, but I cannot reproduce the problem when I do the same action manually in my browser.
@tisto, @gforcada, what is the procedure to launch one single robot test (so I can inspect and debug easilly) ?

**Comment by pbauer** [comment]
You can test one file with ``./bin/test -s Products.CMFPlone -t 'test_fields.robot'`` 
There is no way that I know of to only scenario, you'd have to comment-out the test you don't want to run.
BTW: You can also exclude tests, e.g. all robot-tests using ``-t !'.robot'``

**Comment by ebrehault** [comment]
@pbauer apprently we cannot use the `-t` option for robot tests:
```
./bin/test -s Products.CMFPlone -t 'test_folder_contents.robot'
Total: 0 tests, 0 failures, 0 errors, 0 skipped in 0.000 seconds.
```

I tried that way:
```
bin/robot-server --reload-path src/Products.CMFPlone/Products/CMFPlone/ Products.CMFPlone.testing.PRODUCTS_CMFPLONE_ROBOT_TESTING
bin/robot src/Products.CMFPlone/Products/CMFPlone/tests/robot/test_folder_contents.robot
```
but apparently Firefox is not loading the page (I miss a port config somewhere I guess)

**Comment by pbauer** [comment]
I forgot the ``--all`` switch:
``./bin/test --all -s Products.CMFPlone -t 'test_folder_contents.robot'``

**Comment by ebrehault** [comment]
@pbauer cool! thanks

**Comment by jensens** [comment]
this should work too:

    bin/test --all -s Products.CMFPlone -t 'Scenario Select All items'

**Comment by ebrehault** [comment]
I am confused, Jenkins says the build is still failing but there is no failure: http://jenkins.plone.org/job/pull-request-5.0/840/testReport/ 
The console output mentions an error about RAMCache: `AttributeError: 'RAMCache' object has no attribute '_cacheId'` but I do not understand which test it is related to.
@gforcada @tisto any hints?

**Comment by gforcada** [comment]
@ebrehault I would run it again and if that happens, we can investigate a bit further

**Comment by ebrehault** [comment]
@gforcada, still the same thing: http://jenkins.plone.org/job/pull-request-5.0/853/
The error occurs during tearDownEnvironment in plone.app.testing when it invalidates the RAM cache:
https://github.com/plone/plone.app.testing/blob/master/plone/app/testing/layers.py#L312

So it is not a failing test but a test error. Something removes this `_cacheId` attribute at some point but I do know when.

**Comment by gforcada** [comment]
@ebrehault no idea and no time right now, sorry. You can try adding a pdb with that condition on layers.py and see what's the last tests that run, or add a monitoring that prints if it's there or not on every test...

**Comment by ebrehault** [comment]
@gforcada good idea, thanks

**Comment by ebrehault** [comment]
@gforcada I made a fix in plone.app.testing https://github.com/plone/plone.app.testing/pull/24 which avoids this problem during tearDown.
My understanding is that a test cleans up the RAM cache object but does not destroy it, so next time p.a.testing tries to invalidate the cache, it breaks.
After many investigations, I haven't been able to identify the guilty test (and I still don't know why my PR makes this problem appear).
When you will have a moment (no hurry), please review my p.a.testing PR.

**Comment by ebrehault** [comment]
Jenkins after merging p.a.testing fix: http://jenkins.plone.org/job/pull-request-5.0/867/

**Comment by ebrehault** [comment]
Tests are green. @plone/framework-team : does anyone want to make a final review or shall I merge it?

**Comment by thet** [comment]
Again, I wanted to restate my concerns: This approach isn't flexible enough for some concrete client projects, where users are allowed to log in but do not need all the JavaScript which Site Admins would need. If the merge_with vocabulary would return the already existing bundles and the newly created meta bundle would use the conditional expression of the ``merge_with`` target bundle, that would solve my usecase. IMO that's also not too confusing - it only has the limitation that the conditional expression of the target bundle is used instead of any other possibly defined expressions of other merged bundles.

But then I recalled that I do not have to use the bundle merging mechanism in these cases. I still can 3 different bundles for anonymous, logged-in and site admins and declare all dependencies in there.

**Comment by ebrehault** [comment]
@thet, maybe I haven't been clear about what is the logged-in target: at the moment there is only one bundle in the logged-in metabundle, and that's plone-logged-in (see https://github.com/plone/Products.CMFPlone/commit/0231f2c1073b66e27294716da2b652a91a350a35).

All the site admin specific bundles (like thememapper, or resourceregistry for instance), are **not** in the logged-in metabundle (they are just loaded separately).

So this PR will not push unneeded site admin specific JS to regular users. It just pushes what is needed (=plone-logged-in bundle), and offers the ability to append your own custom bundles to this bundle.

Nothing specific has been done in this PR regarding site admin resources. We might consider creating a third meta bundle for them, but I am not sure it is a high priority (my concern was to fix the anonymous and logged-in resources loading, not site admin).

**Comment by thet** [comment]
Ah yes. Ok, didn't recall that although I could have known myself.

**Comment by frisi** [comment]
@ebrehault please review https://github.com/plone/documentation/pull/518 to make sure this is also documented

**Comment by ebrehault** [comment]
@frisi I have been discussing with @cewing today and he kindly proposed to re-phrase the documentation I made here https://github.com/plone/documentation/tree/plip-1277-meta-bundles-generation to fit the vocabulary he uses in https://github.com/plone/documentation/pull/518

**Comment by frisi** [comment]
@ebrehault and @cewing - good to hear your efforts are in sync ;-)

**Comment by ebrehault** [comment]
@tisto @gforcada I have a question about robot tests.
This PLIP was breaking some robot tests because the metabundles were generated during the test setup, so the `baseUrl` value in config.js was `http://foo/plone`, which breaks everything once the metabundles are used on localhost:55001.

I fixed it by making sure we re-generate the metabundles at run time at the begining of each robot test (I made a keyword which makes a post on `@@resourceregistry-controlpanel`, and I call this keyword in the `Test Setup` section).

I think it is ok, but I would appreciate if you could just review my last commit https://github.com/plone/Products.CMFPlone/commit/f04e376557d8b31426dee49877891e3e43c328dd before I merge this PR.

Thanks.

**Comment by gforcada** [comment]
@ebrehault I see that you applied the same fix  to all robot tests, does this mean that any robot test will need that once this PR is merged? That would be a bit unfortunate, but oh well.

Could this be moved as a ``setUp`` procedure before starting the robot tests? This way it may be faster and less verbose.

**Comment by ebrehault** [comment]
@gforcada it will be needed for any robot test involving JS features (so yes, pretty much every one), and we cannot replace it with a `setUp` procedure (the current `setUp` procedure already triggers metabundling just by calling the plone regular GS profile, but at this time, the current request domain is `http://foo/` so that's what we get in our config.js)

**Comment by gforcada** [comment]
@ebrehault it's late at night, so if there's no way around it, and must be done for each test, well it's unfortunate at best :-(

make sure is documented somewhere at least :-)

**Comment by ebrehault** [comment]
The ideal solution would be the PRODUCTS_CMFPLONE_ROBOT_TESTING layer `setUp` procedure is aware the current url is `http://localhost:55001` and not `http://foo/` but I don't know how to do that.

**Comment by vangheem** [comment]
one unfortunately side-affect of this when combining css resources is that it changes the relative path used so if you have background images, the relative location to specify them by gets messed up.

**Comment by ebrehault** [comment]
Oh, that's bad.
So either we fix the urls before merging (I guess prefixing all urls with '../' should be enough), either we change ++plone++production so it allows to traverse to other resources.

**Comment by jensens** [comment]
Shouldn't this better be a new issue labeled bug, so we can refer to it in an upcoming PR?

**Comment by ebrehault** [comment]
@jensens your are right
done

**Comment by ebrehault** [comment]
@vangheem that's fixed

**Comment by vangheem** [comment]
@ebrehault How does prefixing with ../ fix this?

If you have a css or less file in location:
 - ++plone++mybundle/styles.css 
 - with file reference to: url(images/myimage.png) -> ++plone++mybundle/images/myimage.png

When built urls become:
 - /++plone++production/++unique++2016-04-12T15:32:52.316863/default.css
 - /++plone++production/images/myimage.png -> this is wrong and I don't think there is any way to make it right

At least before you could have your build file in the same directory as where your original resources were. Since we don't have skins magic of fixing traversals this becomes a problem now.

I also don't understand the point of all the extra traversals in our resources. ++plone++production, ++unique++, this is a bit too much isn't it? Append a query string onto a gives you a unique url, we don't need traversals added in that screw up paths.

Building resources on the file system fixes most of these since building inlines the url() resources; however, if you build TTW or if you have fonts defined, the inliner does not inline all font definitions so for some browsers, fonts won't work. I have a suspicion that right now, font icons are broken in some browsers for Plone. It's possible they are unsupported browsers anyways.

I'm not sure I know a good solution with the way we are generating the resource URLs though.

A possible solution is to dynamically provide less variables for every ++plone++ static directory definition so we can interpolate urls correctly.

**Comment by ebrehault** [comment]
@vangheem I guess prefixing with ../ was just fixing a very basic case with Plone resources themselves, you are right, the issue is not fixed at all.
I have re-opened it.

I think the easiest solution is just to replace relative paths with full paths, so `url(images/myimage.png)` is just turned into `url(/++plone++mybundle/images/myimage.png)` once aggregated.

> ++plone++production, ++unique++, this is a bit too much isn't it? Append a query string onto a gives you a unique url

My understanding was querystring are not always recommended for caching management (but I have just asked my colleagues and they say I am wrong :), so I will fix it too).

**Comment by ebrehault** [comment]
@vangheem relative urls are fixed: https://github.com/plone/Products.CMFPlone/pull/1528

But regarding ++unique++, I remember now I did it this way because that's how it's done for bundles:
https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/resources/browser/scripts.py#L48
And @alecpm commit message https://github.com/plone/Products.CMFPlone/commit/67bb341a558c7ec1ffc65ebba3f5eecb5897fc79 mentions it is better for caching to use ++unique++ traversal instead of ?version=.

**Comment by vangheem** [comment]
I think my point is moot anyways. Since we have merging going on, the paths are never going to be what you want them to be.

Using absolute works; however, when in dev, it won't because you are at zope root(unless the ++plone++ traverser works from zope root, I'm not sure)

**Comment by ebrehault** [comment]
Sorry, my comment was incorrect:
It is not: `url(/++plone++mybundle/images/myimage.png)`
but: `url(++plone++mybundle/images/myimage.png)`
(so it is relative to the default.css url, which is always served from the portal root)

**Comment by vangheem** [comment]
nice, that works. Thank you.

**Comment by mauritsvanrees** [comment]
This pull request contains commit 608077966620fa8a094e51ac474dee1ad6e2e731 which adds the `Refresh JS/CSS resources` robot keyword. It calls a line like this:

```
Execute Javascript    $.post($('body') ...)
```

With `plip-zope4.cfg` on coredev 5.1 this still works fine locally but fails on Jenkins, saying:

```
WebDriverException: Message: $ is not defined
```

I have disabled it in commit c11527bf12093011d19fadd477a970c31ae4c118 of the `thet-zope4` branch. We'll see how Jenkins likes that in the Zope4 PLIP job.

If that fails, it would be interesting to have another way of doing the same thing.
Or maybe the Firefox on Jenkins needs to be newer. Locally I am using Firefox ESR 45.7.0.

ebrehault:
Strange, did you try to wait a little bit before call `$.post`?

Note: I do not recommend to upgrade Firefax as the newer versions are not properly supported by Selenium.

mauritsvanrees:
Yes, I tried a `Sleep  2` first, which did not help on Jenkins.

mauritsvanrees:
Tracking this in issue #1933.

PR #1305, opened 2016-01-03
4.3 write on read

jensens:
see plone/buildout.coredev#150 - merge together with others


mauritsvanrees:
Does anyone know why my commit 168a79420251e792615294fe1896fa69eed0e402 was cherry-picked? It is about emaillogin, which seems unrelated. But the strange thing is that without this commit on this branch, I get test failures in `emaillogin.txt` and with this commit I get test failures in `testEmailLogin.py`. Or the other way around, I don't remember.
Anyway, should be fine to include it. I have added fixes for the related test failures, plus a few more.

mauritsvanrees:
Superseded by https://github.com/plone/buildout.coredev/pull/176, from which the CMFPlone part (branch 4.3-write-on-read-orig-protect) does the same as the above pull request, except that it does not add plone4.csrffixes to the dependencies. If we want to, we can try that another time.

PR #1307, opened 2016-01-04
Fix url generation for tinymce when using virtual hosting

vangheem:


vangheem:
tests running: http://jenkins.plone.org/job/pull-request-5.0/671/

Products/CMFPlone/patterns/__init__.py
```diff
@@ -242,6 +241,8 @@ def tinymce(self):
 
         image_types = settings.image_objects or []
         folder_types = settings.contains_objects or []
+
+        site_path = generator.portal_url.replace(self.request['SERVER_URL'], '')
```
thet: I don't know about how safe it is to use SERVER_URL. but if you just want to strip out the domain including protocol, why not using a regex for this? This regex should do it:
``.*\:\/\/((?![\/?]).)*``

This matches any protol, which is delimited by a ``:\\`` sequence. Then it matches all characters until delimited by a path delimiter ``/`` or query string delimiter ``?``, using regex negative lookahead.

E.g:
```
>>> import re
>>> re.sub(r'.*\:\/\/((?![\/?]).)*', '', 'webdav://domainname.com/abc/xyz')
'/abc/xyz'
```

pbauer: Maybe off-topic since I did not look at how ``SERVER_URL`` changes with rewrite rules. But please do not use a regex here. I consider regular expressions as write-only code which is generally bad. Use ``urlparse`` instead:

```python
>>> from urlparse import urlparse
>>> o = urlparse('webdav://domainname.com/abc/xyz')
>>> o.path
'/abc/xyz'
```

thet: ha! urlparse. nice.
but what do you mean with write-only code?
i have no probs with regex except that they are hard to understand if not clearly documented.

vangheem: The point isn't only to strip off the domain. The point is to get the absolute path to the site. This is different depending on if it is using virtualhostmonster or not.

vangheem: So urlparse and or a regex won't work. We need something that knows about where the root of the site is.

thet: something like (replace getSite with something from ``plone.app.layout.globals.portal``, which gets the root portal):
```
>>> from zope.components.hooks import getSite
>>> from urlparse import urlparse
>>> site_path = urlparse(getSite().absolute_path()).path
```
?

thet: Ah, I think that doesn't help with VirtualHosting, when rooting to a subsection of the site. Hm...

vangheem:
Modified a bit. I'm going to merge as it seems like this is the best way to do it. Pretty nasty right now for deployed sites being broken.

PR #1313, opened 2016-01-13
Fix use of icons in search results

vangheem:


vangheem:
build: http://jenkins.plone.org/job/pull-request-5.0/684/

hvelarde:
I think those icons should respect the visibility configuration; have you though on that?

vangheem:
Probably. It should be taken care of by the body class applied and the theme--I think that's the way it's been done everywhere else. That isn't really part of this PR. This fixes the broken search results. Right now, search results show "False" in place of that icon.

PR #1329, opened 2016-01-25
Fix custom tinymce content styles not getting included correctly

vangheem:


vangheem:
tests running: http://jenkins.plone.org/job/pull-request-5.0/744/

PR #1333, opened 2016-01-25
Move p.a.discussion monkey patch for reindexing conversations

fredvd:
…logTool.py

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -22,6 +22,7 @@ from Products.CMFCore.permissions import AccessInactivePortalContent
 from Products.CMFCore.utils import _checkPermission
 from Products.CMFCore.utils import _getAuthenticatedUser
 from Products.CMFCore.utils import getToolByName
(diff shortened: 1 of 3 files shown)


Products/CMFPlone/CatalogTool.py
```diff
@@ -425,6 +427,21 @@ def indexObject(obj, path):
                     safe_callable(obj.indexObject)):
                 try:
                     obj.indexObject()
+
```
hvelarde: a few things about this implementation:

1. use the api: `api.portal.get_tool('portal_catalog')` and move that line inside the `if` block
2. is possible to be in here without access to the catalog?
3. when exactly that `StopIteration` could happen? is not obvious to me

fredvd: We're cleaning up the monkey patch, not refactoring code as this stage. 

Furthermore we can't use plone.api in Products.CMFPlone (circular dependencies)

tisto: @hvelarde plone.api usage is not allowed in Plone core.

PR #1347, opened 2016-01-29
fix: add barceloneta path in less configuration

Gagaro:
This is to fix https://github.com/plone/plone.app.theming/pull/92


Gagaro:
Tests should start there http://jenkins.plone.org/job/pull-request-5.0/789

vangheem:
Thanks!

PR #1356, opened 2016-02-01
small removal of Zope mention 

tkimnguyen:
...while logging out

### DIFF
--- a/Products/CMFPlone/skins/plone_login/logged_out.cpt
+++ b/Products/CMFPlone/skins/plone_login/logged_out.cpt
@@ -31,10 +31,10 @@
 
 <tal:notloggedoutafterall tal:condition="not: isAnon">
     <h1 class="documentFirstHeading"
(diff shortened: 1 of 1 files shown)


hvelarde:
why?

jensens:
I personally dont care if there Zope in or not. But for the average user its superfluous information and so this PR is imo ok.

gforcada:
:+1: there little to no benefit to know that it is Zope from a user interface point of view, just like we renamed the dexterity types control panel to be the types control panel, does not make sense to say that is dexterity (from plone 5 on)

hvelarde:
I see, +1 then

jensens:
@tkimnguyen please add a changelog entry and merge

jensens:
@tkimnguyen nudge

tkimnguyen:
Going to redo this (sorry for the delay) as a new PR with the changelog entry

tkimnguyen:
New PR is here https://github.com/plone/Products.CMFPlone/pull/1457

PR #1190, opened 2015-10-17
Zope 4

thet:
First part of a series of Zope 4 compatibility changes.
the "Control_Panel" was removed from the ZODB. Relevant code from App.ApplicationManager was copied over.
Did not add new tests, but old should pass.
Plus  other test fixes.

PR #1370, opened 2016-02-07
Add compatibility with zope master

pbauer:
replaces https://github.com/plone/Products.CMFPlone/pull/1190 (I f**\* that one up with rebases)

This pr has only Zope master compatibility changes:
- the "Control_Panel" was removed from the ZODB
- bobobase_modification_time is gone
- Products.Five.browser.metaconfigure has changed


gforcada:
@pbauer maybe this comes handy then: http://docs.plone.org/develop/coredev/docs/git.html :-)

Products/CMFPlone/controlpanel/browser/maintenance.py
```diff
@@ -58,10 +71,17 @@ def handle_shutdown_action(self, action):
                 default=u'You are not allowed to manage the Zope server.'
             )
             return
-        context = aq_inner(self.context)
-        cpanel = context.unrestrictedTraverse('/Control_Panel')
-        result = cpanel.manage_shutdown()
-        return result
+        try:
+            user = '"%s"' % getSecurityManager().getUser().getUserName()
+        except:
```
mauritsvanrees: No bare excepts please.
I guess the accepted 'exception' to this rule is that you can catch ConflictError and KeyboardInterrupt and raise them, and do a bare except after this. But I would expect catching ValueError and AttributeError is enough here.

Products/CMFPlone/controlpanel/browser/maintenance.py
```diff
@@ -72,18 +92,24 @@ def handle_restart_action(self, action):
                 default=u'You are not allowed to manage the Zope server.'
             )
             return
-        context = aq_inner(self.context)
-        cpanel = context.unrestrictedTraverse('/Control_Panel')
+
+        try:
+            user = '"%s"' % getSecurityManager().getUser().getUserName()
+        except:
```
mauritsvanrees: No bare excepts please.

Products/CMFPlone/controlpanel/browser/maintenance.pt
```diff
@@ -105,6 +105,10 @@
                             tal:condition="form_name"
                             tal:content="form_name">Form name</legend>
 
+                    <p>
+                        <strong i18n:translate="text_zope_database_name">Database name:</strong> <span tal:replace="view/dbName" />
```
mauritsvanrees: The `dbname` should be in an `i18n:name` for optimal translation in some languages (like right-to-left languages).

thet: Even though the dbName is the name you've choosen for your ZODB? The value comes from the view and is ``self.portal()._p_jar.db().database_name``

mauritsvanrees: Where the dbName comes from does not matter. It is part of a translatable string so it should get an i18n name. The lines after it, with the current database size, have an i18n:name too.

thet: Ah, I see. You want an ``i18n:name`` on it. I misunderstood that as an ``i18n:attribute``. But still: shouldn't the next paragraph changed instead? AFAIK you cannot translate text with html tags inside. Or how would you preserve the ``<strong>`` tag, when you apply the ``i18n:translate`` attribute at the surrounding paragraph, like below?

thet: Well - now I grep'd for ``<strong>`` and ``<span>`` in plone.app.locales and it's really done a couple of times. Will change as suggested, @mauritsvanrees !

jensens:
needs rebase

jensens:
There are still two bare excepts used, would be good to have them more specific.

jensens:
fixed bare except and squashed commits.

Products/CMFPlone/controlpanel/tests/test_controlpanel_security.py
```diff
@@ -29,13 +29,14 @@ def setUp(self):
...
         self.assertTrue(view())
 
     def test_plone_app_registry_in_controlpanel(self):
         self.controlpanel = getToolByName(self.portal, "portal_controlpanel")
-        self.assertTrue('plone.app.registry' in [a.getAction(self)['id']
-                                                 for a in self.controlpanel.listActions()])
+        self.assertTrue(
+            'plone.app.registry' in [
+                a.getAction(self)['id']
+                for a in self.controlpanel.listActions()])
```
gforcada: That can be changed to a ``self.assertIn``

jensens:
merged into master

Issue #974, opened 2015-09-14
"Plone 5 rocks!" message should be removed from the default theme


hvelarde:
According to @agnogueira this message is included in https://github.com/plone/plonetheme.barceloneta/blob/325e892db8f0f77ec37dd3ebbaa140c431197ade/plonetheme/barceloneta/theme/index.html#L36:L44

![selection_004](https://cloud.githubusercontent.com/assets/341393/9851209/e893683e-5acd-11e5-9cf3-d8a24f0a1c06.png)

pbauer:
It is here: https://github.com/plone/plonetheme.barceloneta/blob/master/plonetheme/barceloneta/theme/index.html#L36

It is included on the frontpage like this: 
```xml
 <drop css:theme="div.principal" css:if-not-content="body.template-document_view.section-front-page" />
```
(see https://github.com/plone/plonetheme.barceloneta/blob/master/plonetheme/barceloneta/theme/rules.xml#L36).
Personally I like it but I understand that it can be confusing since it is not editable ttw.

hvelarde:
IMO, this is a potential source of problems and must be removed

vangheem:
I agree. Move the saying to the body text if we want to keep it.

esteele:
My plan is to replace that with a link to the feature list. 

My other complaint with the text there is that it's not translatable since it's part of index.html.I'm kind of tempted to stick another portlet manager in above the content body and use that to insert this design element.

hvelarde:
@esteele I though we wanted to get rid of portlets, not add more; IMO, the footer portlet manager should be deprecated as soon as Mosaic is ready.

vangheem:
Less, please no more portlet managers. Can't we just put it in the body html? We can put more rich html in there--it won't get stripped.

polyester:
I concur, no more portlet managers. I do very much like the idea of having a 'hero' element (which this basically is) in the default Barceloneta theme; but issues of translation could make that difficult.

In a pinch, I would probably remove it entirely, but leave it well-commented (as comments) in Barceloneta. And then write a page "How to enable a Hero element on your front page" in the docs. I volunteer do that in Munich. It would serve as a perfect example of the powers of Diazo, without posing difficulties in the initial user experience especially of non-English speakers.

Our resources are minified before putting in production; so having comments in the non-minified version are acceptable.

esteele:
Fair enough. It's an element of this theme, not every Plone site.

[...]

PR #1374, opened 2016-02-11
move hero to content

gyst:
fixes https://github.com/plone/Products.CMFPlone/issues/974
only together with https://github.com/plone/plonetheme.barceloneta/pull/83


thet:
It would be awesome to have a ``hero`` format selection class for TinyMCE too! This setting has to be added to the registry.

Provided, this approach found consensus. But it's even translatable, isn't it? (that was @esteele objection in the original discussion)

gyst:
Yes it's translatable. And the corresponding barceloneta PR anticipates TinyMCE support in two ways:
- The actual styling classes are part of the theme. The content only has to flag a block '.hero'.
- The hero is supported on any page and auto-removes its boilerplate when not used on a page.

pbauer:
@thet the whole ``plone-frontpage.pt`` is translated and its content is set as the content of the fontpage in the target-language in https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/setuphandlers.py#L174-L195 The code is somewhat crazy but it works.

djay:
@gyst One enhancement would be allow it to be either in a portlet or in the main content. It might complicate the diazo but the advantage is that provides a way for the user to have a hero on a whole subsection of their site if they want. That would require documentation somewhere however.

I think its worth making the Hero style appear in the tinymce format menu. If the user removes the hero element they would have no way to add it back without directly editing the html.

gyst:
@djay The diazo rules I wrote should also support a hero in a portlet. They just take any '.hero' content anywhere and plunk it into the '#hero' slot above.

thet:
@gyst what @djay said about the diazo class in TinyMCE format menu is what I meant. You have to add this as an registry.xml entry.

jensens:
@gyst thank for caring! I reviewed it a bit:

- rebase is needed here.
- do we need an upgrade step? I dont think so, but I'am not sure.
- do we need to update docs? (if so please create a ticket at plone/documentation)
- do we need to update translations? (if so please create a ticket at collective/plone.app.locales)

This can be merged w/o TinyMCE changes. Would be cool to have them anyway.

This has to be merged with plone/plonetheme.barceloneta#83 (note to self)

gyst:
Hi @jensens 
- Rebase is needed here. Done.
- Do we need an upgrade step? No. The only effect on existing sites is that it removes the hero from an old front-page, in case you had not removed it already. That's a feature, not a bug, see community discussion. 
- Do we need to update docs? No. This makes it possible to remove the hero from the editor in an obvious way. Adding a hero should not be documented but enabled in TinyMCE.
- Do we need to update translations? Yes. https://github.com/collective/plone.app.locales/issues/102

Issue #1342, opened 2016-01-27
PLIP: Portal actions control panel


ebrehault:
## Abstract
Allow to manage portal actions from a Plone site setup control panel.

## Motivation
Everytime I show how to manage a Plone site to a customer, everthing looks nice and great, but when he/she asks if we can also manage the footer links, I am a little ashame to show the portal_actions ZMI interface.

## Assumptions
N/A

## Proposal & Implementation
I propose to implement a regular Plone control panel (based on z3c.form) that will reproduce the very same features as the portal_actions ZMI interface.
This PLIP is just about UI, it will not impact the action object model or implementation, it will not impact the way they are accessed in other packages, and it will not impact the way they are declared in a GS profile.

## Deliverables
New version of Products.CMFPlone.

## Risks
None

gforcada:
:+1:

mauritsvanrees:
+1

hvelarde:
+1

vangheem:
+1

frisi:
you might want to have a look at https://pypi.python.org/pypi/quintagroup.plonetabs/

tkimnguyen:
Could I second this, given that I created #1253? :) or does it have to be a FWT member?

ebrehault:
Sure you can :)

Everybody can second a PLIP (just like anybody can propose a PLIP). In the PLIP process, the FWT team has 2 major roles:
- at start, accepting or rejecting the PLIP,
- during the implementation, to provide support, and to review the code.

prakharjoshi:
@ebrehault  @tkimnguyen  Can this be done under GSOC for this summer ? This seems like a good feature to implement.

ebrehault:
@prakharjoshi sorry but I have already implemented it during the weekend :) (see #1342).
But don't be disappointed because it was probably too small for a GSOC anyway.

prakharjoshi:
@ebrehault  Oh its a good thing you finished it. :+1:

jensens:
Note to self: In order to apply this the following outcomes need to be merged:

- [ ] Implementation for PLIP 1342 portal actions #1385
- [ ] new vocabularies for portal actions plone/plone.app.vocabularies#26
- [ ] theme additions plone/plonetheme.barceloneta#87

gforcada:
@jensens should we use the milestones for keeping these things more or less organized?

jensens:
+1 - i can auto apply the missing ones tomorrow

[...]

PR #1385, opened 2016-02-14
Implementation for PLIP 1342 portal actions

ebrehault:
See #1342

@vangheem if you want to review it, here is the buildout:
https://github.com/plone/buildout.coredev/blob/5.0/plips/plip-1342-portal-actions.cfg

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,9 @@ Incompatibilities:
 
 New:
 
+- Control panel to mange portal actions
+  [ebrehault]
+
 - new less variable to configure the width of the toolbars submenu called
   ``plone-toolbar-submenu-width``.
   [jensens]
@@ -69,7 +72,6 @@ New:
 
 - Merge JS and CSS bundles into meta-bundles to reduce the number of requests
   when loading a page (PLIP #1277)
-  [ebrehault]
 
 Fixes:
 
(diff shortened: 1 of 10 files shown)


gforcada:
@ebrehault I just created this PLIP job so that you can run it as many times as you wish:

http://jenkins.plone.org/view/PLIPs/job/plip-1342-portal-actions/

ebrehault:
thank you :)

ebrehault:
@gforcada could you tell me if there anything wrong with my PLIP buildout https://github.com/plone/buildout.coredev/blob/5.0/plips/plip-1342-portal-actions.cfg
because I get a very strange error in Jenkins http://jenkins.plone.org/view/PLIPs/job/plip-1342-portal-actions/1/console
(it looks like if plone.portlets is not in the path, and the PLIP is not related to plone.portlets, it is not in the checkouts or anything...)

Note: is it mandatory to extends jenkins.cfg ? (I see that some PLIP buildout extends it, some don't)

gforcada:
@ebrehault I think that jenkins.cfg should not be extended... 

As for debugging why plone.portlets is not there, you can always run buildout annotate on the plip cfg file and check the output.

It's definitely something weird :-/

ebrehault:
Tests are ok: http://jenkins.plone.org/view/PLIPs/job/plip-1342-portal-actions/2/

PR #1409, opened 2016-02-25
fix: sort relateditems tree by sortable_title in tinymce

Gagaro:
Trees are sorted on is_folderish at the moment, but they also are filtered to only show is_folderish contents. This allow easier finding of the folder/content we want.

See also https://github.com/plone/plone.app.widgets/pull/132


Gagaro:
http://jenkins.plone.org/job/pull-request-5.0/960/

PR #1416, opened 2016-02-26
Add less variable for width of secondary toolbar column

jensens:
needs be merged together with PR plone/plone.app.upgrade#68 in plone.app.upgrade


jensens:
I will prepare a new PR with broader scope

PR #1417, opened 2016-02-26
Fix #637 return a JSON error instead of the Plone error page for non-HTML resources

### LINKED ISSUE #637: Return HTTP errors in proper format
Proposer: Eric Brehault
Seconder:

# Motivation
When a page does not exist, or has an error, or is not allowed for the user, Plone returns the appropriate HTTP error (404, 500, ...), and the response is a human readable page, properly skinned, which nice for the user.
And if the requested resource is not a page (an image, a JS file, an AJAX call, etc.), Plone also returns this human readable page.
It is useless because the page will not be displayed, and it produces many problems:
- the response is very heavy,
- it involves a lot of processing (Plone will render an entire page for nothing),
- for AJAX call, the response cannot be easily interperted,
- it might produce a cascade of errors (for instance: the regular response is not supposed to be rendered via Diazo, as it is not an HTML page, but the error is rendered by Diazo, and it might produce another error).

# Proposed solution
We could display the human readable error page only if the current request `HTTP_ACCEPT` parameter contains `text/html`, in other cases, we would just return a simple JSON error reponse.

# Proposed implementation
Test the `HTTP_ACCEPT` value in `Products/CMFPlone/skins/plone_templates/standard_error_message.py`, and call the existing template or make a JSON response accordingly.

# Risks
No identified risks.

ebrehault:


Products/CMFPlone/tests/redirection.txt
```diff
@@ -31,5 +31,16 @@ But it should not give an error while rendering the default error page:
 
   >>> "the following error occurred while attempting to render the standard error message" in browser.contents
   False
+
+As it is am image, it should return a JSON message
```
hvelarde: "As it is **an** image…"

Products/CMFPlone/tests/redirection.txt
```diff
@@ -31,5 +31,16 @@ But it should not give an error while rendering the default error page:
...
+  >>> '{"error_type": "NotFound"}' in browser.contents
```
hvelarde: IMO this should return `404` and not `NotFound`

ebrehault: well, at first I agreed, but:

- apparently this information is not available in `Products/CMFPlone/skins/plone_templates/standard_error_message.py`
- the HTTP status is already prviding this information, there is no need to repeat here, so I thought the error type might be interesting

Products/CMFPlone/skins/plone_templates/standard_error_message.py
```diff
@@ -27,6 +27,10 @@
 error_traceback = kwargs.get('error_traceback', None)
 error_value = kwargs.get('error_value', None)
 
+if "text/html" not in context.REQUEST.getHeader('Accept', ''):
+    context.REQUEST.RESPONSE.setHeader("Content-Type", "application/json")
+    return '{"error_type": "%s"}' % error_type
+
```
jensens: As a developer I would really love to have the traceback and other error values available. But OTOH it make the response large and we want small responses in production. Also we would need to take the permissions into account. 
What about adding the additional information only if zope is running in debug mode (foreground)?

Products/CMFPlone/skins/plone_templates/standard_error_message.py
```diff
@@ -27,6 +27,10 @@
 error_traceback = kwargs.get('error_traceback', None)
 error_value = kwargs.get('error_value', None)
 
+if "text/html" not in context.REQUEST.getHeader('Accept', ''):
+    context.REQUEST.RESPONSE.setHeader("Content-Type", "application/json")
+    return '{"error_type": "%s"}' % error_type
```
jensens: if we add new code we should use the ``.format`` style string templates, shouldnt we?

``'{"error_type": "{0:s}"}'.format(error_type)``

ebrehault: done

Products/CMFPlone/skins/plone_templates/standard_error_message.py
ebrehault: I know it would be nice, but that kind of things will be probably handled
by plone.restapi.
What I have done here is a very basic fix, it is in a Python script so we
cannot import many things (for instance I had to serialize the JSON message
manually, I could not do that with an entire traceback), and replacing
standard_error_message.py with a browser view could be very tricky, as it
is called by acquisition everywhere (in OFS.SimpleItem for instance), so I
decided to keep it the way it is, even if we cannot implement everything we
would expect.

ebrehault:
@jensens you might get a message from Jenkins because this merge will break the plone.app.theming tests.
I have merged my associated PR on plone.app.theming and I added it to checkout.cfg: https://github.com/plone/buildout.coredev/commit/d76eb7434f36c20a30dd2afe75caf522bbffd728
so it will be fixed.

gforcada:
@ebrehault @jensens merged this branch resulted in 4 errors http://jenkins.plone.org/job/plone-5.0-python-2.7/4188/ Eric already fixed three of them in p.a.theming http://jenkins.plone.org/job/plone-5.0-python-2.7/4190/, but there is still one left in CMFPlone itself. Do you mind fixing that one as well?

I see that the Pull Request job was not used before merging this, although 99% of pull requests merged by you use the Pull Request job, you forgot it for this one, just a little reminder of why is so important to use it always :-)

Pull requests jobs that have been started since this pull request was merged are failing :-/

jensens:
I'll look at it now.

ebrehault:
@jensens I reproduce the failure.
The tests were ok before my last change (=replace % with format()), I probably made a stupid typo or something.

ebrehault:
got it: with format(), curly brackets must be escaped :)

ebrehault:
The fix is pushed

jensens:
:+1:

gforcada:
@ebrehault thanks!

PR #1419, opened 2016-02-26
Registered post_handler instead of plone-final.

mauritsvanrees:
The plone-final import step now does nothing.

Instead, we redefined the old handler as a post_handler explicitly for
our main profile.  This is guaranteed to really run after all other
import steps, which was never possible in the old way.

The plone-final step is kept for backwards compatibility.

This requires Products.GenericSetup 1.8.2.

### DIFF
--- a/Products/CMFPlone/setuphandlers.py
+++ b/Products/CMFPlone/setuphandlers.py
@@ -172,14 +172,30 @@ def assignTitles(portal):
             setattr(aq_base(obj), 'title', title)
 
 
+def dummy_import_step(context):
+    """Dummy import step.
+
+    The plone-final import step used to call importFinalSteps below.
+    But plone-final was never guaranteed to be run as final step.  So
+    more and more import steps were added to its dependencies to let it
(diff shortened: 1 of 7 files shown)


Products/CMFPlone/setuphandlers.py
```diff
@@ -172,14 +172,30 @@ def assignTitles(portal):
...
+    run later and later.  Not nice.
+
+    With Products.GenericSetup 1.8.2, we can add a post_handler to a
+    profile (and a pre_handler).  We now do that.  So the plone-final
+    import step is no longer needed.  But others may depend on it, so we
+    keep it for now.  This dummy import step handler is meant for
+    that.
     """
-    Final Plone import steps.
+    pass
```
jensens: is there a way to issue a deprecation warning in some way here?

mauritsvanrees: Not really.
Well, we can issue a deprecation warning, but it would be displayed for every profile, because this import steps is still run for every profile.

PR #1425, opened 2016-02-29
better formatting of generator and generated file in orer to increase…

jensens:
… readability of Python code and generated Grundfile.js

### DIFF
--- a/Products/CMFPlone/_scripts/_generate_gruntfile.py
+++ b/Products/CMFPlone/_scripts/_generate_gruntfile.py
@@ -1,25 +1,20 @@
-from Products.CMFCore.FSFile import FSFile
-from Products import CMFPlone
-from Products.CMFPlone.interfaces import (
-    IBundleRegistry,
-    IResourceRegistry)
-
-from Products.Five.browser.resource import FileResource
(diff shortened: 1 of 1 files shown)


Products/CMFPlone/_scripts/_generate_gruntfile.py
```diff
@@ -134,13 +130,11 @@
                     plugins: [
                         new require('less-plugin-inline-urls'),
                     ],
+                    paths: {less_paths},
                     modifyVars: {{
-                      {globalVars}
+{globalVars}
                     }}
```
thet: why this unindent?

thet:
I guess, that's untested code, isn't it?
I'll let run a Pull-Request job anyways: http://jenkins.plone.org/job/pull-request-5.0/988/

jensens:
I dont see any test for it, anyway its also difficult to test. This would need quite a lot of refactoring and out of scope of this PR.

Products/CMFPlone/_scripts/_generate_gruntfile.py
jensens: because I indent manually and then it fits afterwards (before only first line was indented).

PR #1426, opened 2016-02-29
Minor toolbar cleanup:

jensens:
- more less and less css,
- typo corrected in less variable,
- use font fallback chain as in barcelonetta (works also w/o the theme),
- better readability with a darker background in submenu

This one needs to be merged together (or before) plone/plone.app.upgrade#69


thet:
LGTM

Issue #1430, opened 2016-03-01
Fix html validation


mauritsvanrees:
When validation the homepage of a fresh Plone 5 site, there are several errors and warnings. Let's fix them. This is from the validator at https://validator.w3.org:

- Error: Bad value menu for attribute role on element section.
- Warning: Element main does not need a role attribute.
- Warning: Section lacks heading. Consider using h2-h6 elements to add identifying headings to all sections.
- Warning: The banner role is unnecessary for element header.
- Warning: The contentinfo role is unnecessary for element footer.

Some are there multiple times. There can be others, for example when you change the portal_actions footer to show a portlet title (which is ugly by the way, but that is why it is switched off by default).

I have fixed these and will create pull requests and link them here.  There will still be failures on other pages, but this is a nice start.

The most controversial may be the warning 'section lacks heading'. We have various `section` elements with a `header` element in it, but that is not the same as a heading. Headings are `h1`, `h2`, etcetera. And in a lot of these cases it does not make sense to add a heading. My solution in most cases was to simply change the `section` into a `div`. Seems like that is fine for our current css. There were no glaring javascript errors due to this change either. Let's see what Jenkins has to say about that in a while.

Anyway, since it is only a warning, we could ignore it: adding a heading in a section is only a recommendation, not a requirement.

I have added https://github.com/plone/buildout.coredev/blob/5.0/plips/maurits-fix-html-validation.cfg with checkouts and source branches so you can try it out. I'll create the individual pull requests now and feed them to Jenkins in one go.

Pull requests:

~~https://github.com/plone/plone.app.contenttypes/pull/329~~ merged for 5.0 and 5.1
~~https://github.com/plone/plone.app.layout/pull/81~~ merged for 5.0 and 5.1
~~https://github.com/plone/plone.app.portlets/pull/64~~ merged for 5.1 only
~~https://github.com/plone/plone.app.theming/pull/98~~ merged for 5.0 and 5.1
~~https://github.com/plone/plonetheme.barceloneta/pull/90~~ merged for 5.0 and 5.1
~~https://github.com/plone/Products.CMFPlone/pull/1431~~ merged for 5.1 only

For easy copy-pasting in Jenkins:
```
https://github.com/plone/plone.app.contenttypes/pull/329
https://github.com/plone/plone.app.layout/pull/81
https://github.com/plone/plone.app.portlets/pull/64
https://github.com/plone/plone.app.theming/pull/98
[...]

PR #1431, opened 2016-03-01
Fix html validation

mauritsvanrees:
Part of #1430.


thet:
The ``<div>`` elements should also be changed to ``<aside>``, as done here plone/plone.app.portlets#64

thet:
I'll do...

thet:
http://jenkins.plone.org/job/pull-request-5.1/20/

PR #1447, opened 2016-03-08
Fix #1113: can't add content after adding Plone 5 site without Example content

### LINKED ISSUE #1113: can't add content after adding Plone 5 site without Example content
I created a Plone 5 site and unchecked the box "Example content".
This caused "Content-types (plone.app.contenttypes)" to not be installed.
Only after going to portal_quickinstaller and installing it manually I was able to add content pages, folders, etc. to the site.

jensens:
fixes #1113

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,10 @@ New:
 
 Fixes:
 
-- *add item here*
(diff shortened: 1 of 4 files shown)


jensens:
something is wrong, start at new branch

PR #1456, opened 2016-03-10
contact form fix

tkimnguyen:
Remove the encoding of the Reply-To header which was garbling the return address

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,9 @@ New:
 - Remove Zope mention in logout form 
   [tkimnguyen]
 
+- Do not encode reply-to email address for contact-info form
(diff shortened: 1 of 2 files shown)


jensens:
So, even if this one is green now.... but why exactly is this needed? In theory the old one should be fine too, while more modern RC6532 allows UTF8 too https://tools.ietf.org/html/rfc6532

tkimnguyen:
@jensens in two test cases in production we found that the encoding would result in reply-to email addresses that looked like this

somereal@mac.com@me.com 

or like

Reply-To: =?utf-8?q?kim=2Enguyen=40wildcardcorp=2Ecom?=

PR #1464, opened 2016-03-12
Fix robot tests

gforcada:
With the ongoing work on refactoring some HTML,
see https://github.com/plone/Products.CMFPlone/issues/1430,
this test was broken.


gforcada:
@thet @mauritsvanrees @jensens this should be the first to merge before moving to any other pull request (targeting 5.1).

PR #1476, opened 2016-03-17
Fixed versioning for File and Image.

iham:
After fixing the problem with plone/Products.CMFEditions#36, adding a File or Image failed by not having Locking-support.

This is fixed, as ilocking (if not already present - not at blobtypes) is also added when iversioning is applied to the dexterity type.

I also added the python package "mock" to the test requirements as it is used in the tests.


jensens:
@iham, please rebase.

iham:
see https://github.com/plone/Products.CMFPlone/pull/1587 instead.

PR #1490, opened 2016-03-24
Ensured front-page is English when creating an English site.

mauritsvanrees:
Previously, when creating an English site with a browser that prefers a different language, the body text ended up being in the browser language.

For languages without a front-page text translation the same happened:
they got the other language instead of English.

This pull request targets 5.1, but the same should be done on 5.0 and 4.3, possibly with minor changes. I can handle that once this is merged.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,13 @@ New:
 
 Fixes:
 
+- Ensured front-page is English when creating an English site.
+  Previously, when creating an English site with a browser that
+  prefers a different language, the body text ended up being in the
+  browser language.  For languages without a front-page text
+  translation the same happened: they got the other language instead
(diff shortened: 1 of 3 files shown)


gforcada:
Looks good (and with a test!)

gforcada:
Regarding backporting to 5.0 and 4.3, I'm not so sure if that effort is actually worth spent in other issues/pull requests. Is not something that people hit on a daily basis, and even if, they would be probably developers who most probably do not care much about the actual text per se :-)

But of course you are totally free to do the backports if you feel they would be important for both 5.0 and 4.3!

mauritsvanrees:
I have backported it to 4.3 and 5.0. Jenkins is happy. I have merged them myself.

Issue #1446, opened 2016-03-08
In the toolbar > add new > collection improperly indented 


pigeonflight:
The word "Collection" runs over the collection icon, all the other icons and names are aligned properly
![image](https://cloud.githubusercontent.com/assets/31827/13592283/b03f7b6e-e4be-11e5-87e9-aba22238f143.png)

drex44:
![snapshot1](https://cloud.githubusercontent.com/assets/5825883/13593136/c38f853c-e51e-11e5-9500-b25c91040952.png)


i don't see this in my site

pigeonflight:
@drex44 I'm using the latest Plone 5.0.x checkout (buildout.coredev).

davidjb:
Looks like @pigeonflight's screenshot for me in Plone 5.0.3, appears the same in FF 46a2 and Chromium 48. 

The cause is the CSS rule `.icons-on .contenttype-collection { padding-left: 20px!important}` and this is stemming from its inclusion in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/static/plone-legacy-compiled.css (see the very end of the line).  Removing these style definitions and enabling development mode for the plone-legacy bundle's CSS fixes the issue.

davidjb:
The CSS also affects the Dexterity Content Types page: 
![screenshot from 2016-03-18 13-58-26](https://cloud.githubusercontent.com/assets/1002811/13868150/c65a3ab4-ed11-11e5-92fb-3a2eaab55ccd.png)

(the missing Page/Document icon is another issue which I've already fixed in Dexterity https://github.com/plone/plone.app.dexterity/commit/7586ef1eda2f8e9e8a658880a9de3c2b35e353bd)

What's the process for updating `plone-legacy-compiled.css`?

davilima6:
@davidjb: I think it's this: https://github.com/plone/Products.CMFPlone/blob/master/DEVELOPING_BUNDLES.rst

thet:
Re-stating from https://community.plone.org/t/css-bug-fixing-i-am-about-to-remove-legacy-css-contemplating-if-it-would-be-better-to-override-the-property-in-a-different-file/1809

I had the impression, that ``plone-legacy-compiled.css`` ( https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/static/plone-legacy-compiled.css ) must have been compiled from other source files. But it looks like, it was done manually.
[...]

PR #1491, opened 2016-03-24
Collection CSS fixed  - bug #1446

jordnV:
The word "Collection" runs over the collection icon, all the other icons and names are aligned properly (fixed)


jensens:
- is the source of the problem really only in ``plone-legacy-compiled.css``? I would expect this css's canonical source to be somewhere else (something not named ``compiled`).
- does this apply to 5.1 branch as well? Would be good to have a PR for this branch too please.

jensens:
see https://community.plone.org/t/css-bug-fixing-i-am-about-to-remove-legacy-css-contemplating-if-it-would-be-better-to-override-the-property-in-a-different-file

jensens:
fixed by #1446

PR #1507, opened 2016-04-05
only encode css_body if unicode to avoid unicode error

rnixx:
^^


jensens:
Please add a note to ``CHANGES.rst`` - the change log file.

rnixx:
here you go ;)

PR #1512, opened 2016-04-06
Fix #1511 when aggregating bundles, use ++plone++static overrided versions if any

### LINKED ISSUE #1511: Resources from third party add-ons are not being included in compiled plone-legacy bundle
Seems JS resources registered in Plone 5 using old approach (`jsregistry.xml`) are not included in the final compilation: I installed an add-on and, even as I can see the JS resources listed in `default.js`, the source code is not present.

If I enable development mode, then I can see the source code included in `plone-legacy-compiled.js` and it's executed normally.

ebrehault:


hvelarde:
@esteele this fix must go into Plone 5.0.4

PR #1514, opened 2016-04-07
fix #1513 - Wrong portal_url used for TinyMCE in multilingual site

### LINKED ISSUE #1513: Wrong portal_url used for TinyMCE in multilingual site
When adding an image in the TinyMCE editor in a multilingual site the wrong navigation root is used. If a file is uploaded in the visual editor, and the content item is located at `/nl/agenda/lokroep-zonder-handen/edit` the `@@fileUpload` view is called on `/agenda/lokroep-zonder-handen/@@fileUpload`. The correct path for the file upload view would be  `/nl/agenda/lokroep-zonder-handen/@@fileUpload`, witht the language code included.

See this screenshot in the browser for the 404:

screen shot 2016-04-07 at 10 00 35

The portal_url attribute in [TinyMCESettingsGenerator.portal_url](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/patterns/__init__.py#L29) is pointing to the Plone site instead of the language specific folder. 

I think the fix is easy, use `self.context` when calling `get_portal_url` instead of using `self.portal`. When I try this in a pdb I get this result:

```
(Pdb) get_portal_url(self.portal)
'http://localhost:8100/avk'
(Pdb) get_portal_url(self.context)
'http://localhost:8100/avk/nl'
```

I can provide the fix, but I cannot see any tests related to the `TinyMCESettingsGenerator`.

jensens:
This includes refactoring of this crappy code and also moves stuff from plone.app.layout into this place.


jensens:
need to be merged together with with plone/plone.app.layout#85

Products/CMFPlone/patterns/settings.py
```diff
@@ -0,0 +1,130 @@
...
+from zope.component import getUtility
+from zope.interface import implementer
+
+import json
+
+
+@implementer(IPatternsSettings)
+class PatternSettingsAdapter(object):
+    """
+    This adapter will handle all default plone settings.
```
thet: should read:
"This adapter will handle all default pattern settings."

Products/CMFPlone/patterns/settings.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Right now, it only does tinymce
```
thet: Remove, as it also handles "mark_special_links".

Products/CMFPlone/patterns/settings.py
```diff
@@ -0,0 +1,130 @@
...
+        portal_url = portal_state.portal_url()
+        nav_root = getNavigationRootObject(folder, portal_state.portal())
+        nav_root_url = nav_root.absolute_url()
+        current_path = folder.absolute_url()[len(portal_url):]
+
+        image_types = settings.image_objects or []
+        folder_types = settings.contains_objects or []
+
+        server_url = self.request.get('SERVER_URL', '')
+        site_path = portal_state.portal_url()[len(server_url):]
```
thet: change to:
``ite_path = portal_url[len(server_url):]

jensens:
investigating ...

jensens:
fixed

PR #1522, opened 2016-04-12
Add lead imaging schema

bsuttor:
This PR is related to https://github.com/plone/plone.app.contenttypes/pull/339

It added some settings for plone app contenttypes behavior "leadimage".

I think I also have to add an upgrade step into plone.app.upgrade ?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,8 @@ Incompatibilities:
 
 New:
 
-- *add item here*
+- Add lead imaging schema.
+  [bsuttor]
 
 Fixes:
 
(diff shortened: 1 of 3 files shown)


jensens:
yes, i think so. usually its just an entry in registry.xml there

jensens:
but you're at the wrong branch, you need master for Plone 5.1 - 5.0.x is for bugfixes only. 
Please move this over. Also I'd not make a new schema, but put it into the existing IImagingSchema - see my other comment at plone/plone.app.contenttypes#339 - including proposed validation changes using an invariant.

PR #1523, opened 2016-04-12
Fix bundle aggregation when bundle has no CSS (or no JS)

ebrehault:


gforcada:
Changelog entry please :-)

PR #1525, opened 2016-04-13
Fix bundle aggregation when bundle has no CSS (or no JS) on 5.0.x

ebrehault:
I have already a PR for the same fix on master https://github.com/plone/Products.CMFPlone/pull/1523.

Here is a PR for 5.0.x.
(@plone/release-team, are we supposed to do that for all our important fixes?)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ Fixes:
 - Bundle aggregation must use ++plone++static overrided versions if any.
   [ebrehault]
 
+- Fix bundle aggregation when bundle has no CSS (or no JS)
+  [ebrehault]
+
 
 5.0.4 (2016-04-06)
 ------------------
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Yes, bug fixes should be done on Plone 5.0 as well, if that is what you meant.

I guess it is not always strictly necessary to create a separate pull request if they are small changes and unlikely to break anything: if tests pass for the master pull request, it might be fine to manually merge it into the 5.0 branch as well, as long as you keep an eye out for any Jenkins failures that may pop up unexpectedly.
But often an extra pull request is better and this is recommended. You never know what subtle differences there are.

ebrehault:
@mauritsvanrees well, there is always at least a difference regarding CHANGES.rst (as we will mention the fix in either 5.0.5, either in 5.1a2).

PR #1531, opened 2016-04-15
Fix #1530 compute requirejs config baseUrl on client-side

### LINKED ISSUE #1530: Merged bundles does not work when switching domains
Another issue I've encountered when using the merged bundle feature is that it includes generated requirejs configuration. Generated requirejs configuration includes a `baseUrl` setting which is specific to the url that you first created the site with.

Imagine you create the site, through a tunnel on localhost:8080, then switch the site over to production. The requirejs config has the baseUrl as `http://localhost:8080/Plone` and things start breaking.

\cc @ebrehault

ebrehault:


thet:
As long as we don't have a Plone 5.1 release, we should also include this fix (and related, if any) for Plone 5.0.

vangheem:
Yes, please bring this to 5.0.x

ebrehault:
Sure, I will do it.

PR #1605, opened 2016-05-26
Fixes #1602

### LINKED ISSUE #1602: modal login dialog ignores any redirects
Since Plone 5, the login link uses the `pat-plone-modal` mockup pattern, and by default this ignores any 302 redirects happening during login.  For example, `collective.onlogin` listens to `Products.PlonePAS.interfaces.events.IUserLoggedInEvent` or `Products.PlonePAS.interfaces.events.IUserInitialLoginInEvent` and sets a user-configurable redirect.  This works fine if using the classic `/login` or `/login_form` forms, but breaks when using the modal.

I have created an issue (martinschoel/collective.onlogin#3) describing the situation and two possible workarounds.  

I would suggest adding

```
, "actionOptions": {"redirectOnResponse": true}
```

to the `modal` property of `portal_actions/user/login`  (https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/profiles/default/actions.xml#L248).  This would fix the issue, by forcing the modal dialog to act on any 302 redirects it gets in the ajax call response.  I don't see any risks.

I can submit a PR if there are no immediate objections.

fulv:
Make the modal dialog pattern for login follow any redirects set while logging in.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,9 @@ Incompatibilities:
 
 New:
 
+- Make login modal dialog follow any redirects set while processing the login request.
+  [fulv]
+  
 - Add link to training.plone.org
   [svx]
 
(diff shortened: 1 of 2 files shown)


fulv:
Also fixes martinschoel/collective.onlogin#3.

Issue #1343, opened 2016-01-27
PLIP: assimilate collective.indexing


MrTango:
## Abstract

Move all functionality on existing packages (`Products.ZCatalog`, `Products.CMFCore` and `Products.CMFPlone`) and apply all monkey patches directly to the packages they belong to, like `Products.CMFCore` or `Products.Archetypes`.
## Motivation

Using `collective.indexing` has two main benefits:
- **performance**: indexing operations are queued, optimized and done at the transaction boundary. Only one indexing operation is done per object on any transaction. This means that plone is much faster, specially if lots of event handlers modify objects and keep reindexing the same object over and over again.
- **pluggable**: it creates a utility that enables third-party search indexers to be notified on any indexing operation (that's how collective.solr is plugged-in)

We think they are important and transparent enough that any Plone installation will benefit from it, specially for the performance part, although the pluggable makes even more easier to scale Plone as the site grows.
## Assumptions

Some tests in Plone core will need to be updated to work around the indexing optimizations (as tests expect that objects are indexed/reindexed/unindexed right away).

**No** refactorings will be done on either CMFPlone `CatalogTool` nor CMFCore `CatalogTool` nor ZCatalog `ZCatalog` base class (which both CMFPlone and CMFCore override), this cleanup/refactoring is out of scope.

We deemed more importantly to get the functionality first in rather than merging the catalog tools together. At the same time we encourage any individual/company to step up and propose a PLIP that goes in that direction.
## Proposal & Implementation
- [X] [plip config file on buildout.coredev](https://github.com/plone/buildout.coredev/blob/5.0/plips/merge-collective-indexing.cfg) to track the implementation and branches
- [X] [jenkins job](http://jenkins.plone.org/view/PLIPs/job/plip-collective-indexing) to test the plip config
- [x] move all functionality provided by collective.indexing to currently existing packages ([Products.CMFCore](https://github.com/plone/Products.CMFCore/tree/merge-collective-indexing) and [Products.CMFPlone](https://github.com/plone/Products.CMFPlone/tree/merge-collective-indexing))
- [x] the code of collective.indexing is now GPL and should be re-licensed by the PF-Board to ZPL before getting into CMFCore which are ZPL (@plone/plone-foundation-board)
- [x] add tests where the functionality moved
- [x] fix tests due to the way indexing is done with collective.indexing
- [x] documentation (describe the new feature, how to update tests, how to migrate from collective.indexing)
- [ ] collective.solr branch
[...]

PR #1608, opened 2016-05-28
Remove dependency on CMFUid.

jensens:
Seems unused. Let the tests decide.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,8 +12,11 @@ Incompatibilities:
 
 - Moved code around and deprecated old locations in ``Products/CMFPlone/patterns/__init__``.
   This goes together with same pattern settings changes in ``plone.app.layout.globals.pattern_settings``.
-  Also moved general usable ``./patterns/utils/get_portal`` to ``./utils/.get_portal``.
+  Also moved general usable ``./patterns/utils/get_portal`` to ``./utils/get_portal``.
   Depreacted ``./patterns/utils/get_portal`` and ``./patterns/utils/get_portal``.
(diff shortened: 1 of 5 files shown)


jensens:
see also #1343

gforcada:
At least needs an upgrade step and there are some traces of it in
cmfeditions
El dia 28/05/2016 6:46, "Jens W. Klein" <notifications@github.com> va
escriure:

gforcada:
I was looking into it, as the PLIP for collective.indexing got quite a lot of errors after rebasing my branch on top of CMFUid 2.3 instead of 2.2 and trying to clean CMFEditions from CMFUid traces lead me to ``portal_historyidhandler`` which is at least being used by plone.app.iterate.

I will stop here as that's a bit too much for a small cleanup, but it feels that CMFUid 2.3 branch is not ready at all and probably we stick with 2.2 and try to get rid of it instead of updating it.

I would need a separate PLIP though.

jensens:
Phew. CMFEditions should be changed to use ``plone.uuid``.  The tool in CMFEditions is deprecated anyway and can be removed. If anyone relies with addon code on CMFUid it can be installed anyway as an addon.

TODO:

- [ ] make this a plip
- [ ] Change CMFEditions to get its UIDs from ``plone.uuid`` (ensure old UUIDs are read)
- [ ] Remove CMFEditions old deperecated BBB uid tool.
- [ ] Conditional imports in plone.app.upgrade for CMFUid
- [ ] Upgrade step in plone.app.upgrade to remove all traces of CMFUid

Issue #1483, opened 2016-03-22
PLIP: Retina image scales


didrix:
**Proposer** : Diederik Veeze

**Seconder** : Maurits van Rees
# Abstract

Scale images in plone.scale such that they will look good on devices with pixel ratios higher than 1.
# Motivation

More and more devices these days can display images with higher resolution than the html/css resolution. Plone however often serves images that are scaled to the exact same size as html/css resolution, while a lot of devices can display twice the resolution.
# Assumptions

N/A
# Proposal & Implementation

I propose to add an option to the imaging control panel to set a pixel ratio for image scaling. Images generated in plone.scale will generate images whose height and width are multiplied by this pixel ratio. For instance an image in the site of 320x240 with a scaling factor of 2 should result in something like the following:
`<img src="someimage-640x480.jpeg" alt="Some Image" width="320" height="240">`
# Deliverables
- Scaling factor in plone.scale
- Option for pixel ratio in imaging control panel
- Documentation
# Risks

The image scales will get larger, but the image quality for these larger scales can be reduced while still giving a satisfactory result, for only a small increase in byte size.
# Participants

@mauritsvanrees 
# Note

~~I recently signed the contributor assignment agreement and am awaiting confirmation.~~

ebrehault:
That's interesting.
Two remarks:

- if I understand what you propose, if we enable the 2x option in the Image control panel, then all the images will be 2 times bigger to load, even for 1x devices (I agree the rendering will be correct because of the width and height attributes, but the network impact will be bad).

Don't you think we could use the HTML5 `srcset` attribute, which allows to declare several sources according the context:
```
<img src="images/logo-nope.png"
     srcset="images/logo-big.png 1x,
          images/logo-hd.png 2x"
     width="1824" height="499" alt="">
```
?
That way, any user will load the appropriate image considering her/his device, and there is no need to force a 2x scaling factor.

- your solution (but also mine) assumes the markup will contain the width and height of the image, but right now, when a user insert an image in TinyMCE, we do not have the width and height attributes on the img tag (we just get something like `data-scale="tile"`), and that's the same with many templates where images are inserted by plone (for instance in the navigation portlet, we get the news items image with just `class="image-icon"`).
So we need to change all the `img` markup generation in Plone.
It could be done if we use `tag()` instead of `tal:attributes="src blabla"` in all the considered templates (plus fixing the TinyMCE image plugin).

[...]

PR #1611, opened 2016-05-31
Plip 1483 retina image scales

didrix:
https://github.com/plone/Products.CMFPlone/issues/1483

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -122,6 +122,12 @@ Breaking changes:
 
 New features:
 
+- Make use of plone.namedfile's tag() function to generate img tags. Part of plip 1483.
+  [didrix]
+
+- Add retina scales settings in image handling. Part of plip 1483
+  [didrix]
+
 - Added ``ok`` view.  This is useful for automated checks, for example
   httpok, to see if the site is still available.  It returns the text
   ``OK`` and sets headers to avoid caching.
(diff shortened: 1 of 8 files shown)


ebrehault:
Tests are passing. The code looks ok.
Of course, the `getObject` thing is not ideal but we have no choice I guess.
@thet, any comment? (as you offered to review this PLIP)

mauritsvanrees:
We might want to introduce a catalog metadata column: boolean `has_image` which is True when at least one image field is filled in, or `images` with a list of image field names that have a filled in image. Plus a smart enough indexer to fill this. But that seems something for a different PLIP, out of scope for this one.

ebrehault:
@mauritsvanrees we already have the `getIcon` index, and @didrix still uses it in his modified templates. The only difference is now we have to wake up the objects to get the img tag (but we only do that when we know there is an image, so that's ok I guess).

jensens:
Ok, the current ``@@images``  implementation is kind of limited here. ``getObject`` is usually not what we want. I don't think I will give a +1 for the PLIP as long this is not solved.

So we need to know primary the current UIDs of the image scales. The ``plone.scale`` implementation already attaches this info to the object. Question is, if we may provide an central infrastructure where this is cached, so we can prevent waking up the object. I'd say, Yes:

The ``@@images`` may then adapt the brain itself and wake up the object only if there is no cached information.

There is nothing more needed than ramcache/alternative memcached mapping:
``object-UID+cachekey(height, width, scale, **otherkw) -> fieldname/Scale UID``.

I can help implementing this, its not that difficult, because the bits and pieces are already around.

ebrehault:
@jensens, I agree waking up objects in the advanced search template is quite bad (on the 2 others, I think it is probably harmless), but yes sure, if you think we can avoid it easily, that would be ideal!

PR #1614, opened 2016-06-02
Hide/show actions directly from the Actions control panel list

### LINKED ISSUE #1613: Allow to hide/show actions directly from the Actions control panel list
As @esteele mentionned in #1342

ebrehault:
Fix #1613


ebrehault:
@esteele this is a very basic change, but it's always good to have a review, so whenever you have time, have a quick look and merge it if you think it is ok.

PR #1616, opened 2016-06-02
Issue 1600 - Plone 4.3.x - Added property sort_on

rodfersou:


Products/CMFPlone/profiles/default/propertiestool.xml
```diff
@@ -157,5 +157,7 @@
   <property name="external_login_url" type="string"/>
   <property name="external_logout_url" type="string"/>
   <property name="external_login_iframe" type="boolean"/>
+  <property name="sort_on" type="string">relevance</property>
+  <property name="sort_reversed" type="boolean">False</property>
```
hvelarde: @rodfersou what's the purpose of this property?

Products/CMFPlone/tests/testPortalCreation.py
```diff
@@ -439,6 +439,11 @@ def testTTWLockableProperty(self):
         self.assertEquals(True,
                           self.properties.site_properties.lock_on_ttw_edit)
 
+    def testSortOnProperty(self):
+        # site_properties should have sort_on property
+        self.assertTrue(self.properties.site_properties \
```
hvelarde: this should be written in one line only or this way instead:

```python
self.assertTrue(
    self.properties.site_properties.hasProperty('sort_on'))
```

gforcada: While at it, even better if done like this:
```
self.assertTrue(
    self.properties.site_properties.hasProperty('sort_on'),
)
```
So diff's look better and you know when something ends, there's a nice flake8 plugin that controls this.

Although, this:
```
sort_on_property = self.properties.site_properties.hasProperty('sort_on')
self.assertTrue(sort_on_property)
```
is even better

PR #1629, opened 2016-06-08
Navroot subjects setting

jensens:
This one goes with plone/plone.app.vocabularies#35


jensens:
Upgrade step in plone.app.upgrade is missing.

PR #1637, opened 2016-06-14
Render an empty <dd> tag on folder_listing template when no description is available to avoid validation issues

hvelarde:


hvelarde:
CC @plone/framework-team

hvelarde:
I have no idea where this has gone in Plone 5; if we need to fix the same let me know.

thet:
@hvelarde should not be necessary in Plone5 / plone.app.contenttypes: https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/browser/templates/listing.pt#L102

PR #1642, opened 2016-06-20
Update form_tabbing.js

### LINKED ISSUE #1640: form_tabbing.js doesn't regard hashes in URL
Calling an edit-form with an URL containing a hash, like `front-page/edit#fieldset-dates` one still lands on the default-tab open, not the dates-tab. The fix is to change the line:

`var targetPane = item.find('.enableFormTabbing input[name="fieldset"]').val() || window.location.hash;`

To:

`var targetPane = window.location.hash || item.find('.enableFormTabbing input[name="fieldset"]').val();`

So a hash-URL preceeds the default-selected tab.

Adding this as an issue instead of a PR, because I'm not sure how the procedure is for Plone-4.3 fixes.

ida:
Fix #1640:
- Set correct tab-link-address, getting absolute href, instead of setting a relative link.
- If URL contains a hash of a form-tab-anchor, select that form-tab, instead of using default-tab.


davilima6:
Thank you. Let us know when it's ready for review.

Just one note: do not commit/push directly to the official 4.3x branch. That should only happen when a PR is merged.

davilima6:
Last thing: could you please squash all this PR's commits into one so it's easier to review and less extensive in the Git history? We always try to have just one commit per PR. You can do that with git rebase -i. I'm on IRC if you need a hand.

PR #1644, opened 2016-06-21
Update bower components

thet:
- Align bower components with newest mockup + documentation updates on mockup update process.

- Updated components directory, recompiled bundles.

- Toolbar fixes:
  - Autoformat with cssbrush and js-beautify,
  - Remove ``git diff`` in line 105, which broke compilation.
  - Use patternslib ``pat-base`` instead of ``mockup-patterns-base``.
  - Remove dependency on deprecated ``mockup-core``.

- Remove deprecated resource registrations for ``mockup-parser`` and ``mockup-registry`` from mockup-core. Use those from patternslib instead.

- Add missing ``jquery.browser`` dependency which is needed by patternslib.

- Ignore a bit more in ``.gitignores`` for CMPlones bower components.

- Fix select2 related robot test failures and give the test_tinymce.robot scenario a more unique name.

- Workaround a test problem with outdated Firefox 34 used at jenkins.plone.org. 
  This Workaround can be removed once https://github.com/plone/jenkins.plone.org/issues/179 was solved.


thet:
Must go together with: https://github.com/plone/plone.app.upgrade/pull/79

/cc @vangheem @jensens

thet:
Before merging, I have to fix this issue - the toolbar looks weired. I already got such toolbars, when compiling the plone bundle in an own project, where the plone.js resource file was changed to an own one. Until now, I just temporarily did my changes in CMFPlones plone.js bundle resource, compiled and it worked. But now, it looks like I have to finally fix this annoying problem.
Any hints?

screen shot 2016-06-21 at 16 43 07

thet:
I still would prefer to un-ignore bower_components in mockup. The update process would just be easier. /ref https://github.com/plone/mockup/issues/688

thet:
is it possible, that the plone bundle wasn't updated since quite a while? patternslib's utils.js has a dependency to jquery.browser, which wasn't fullfilled by our stack (it's fixed in this PR), the toolbar couldn't be compiled before @jensens toolbar cleanup because of a missing less variable, and after his fix he also introduced the toolbar layout issues shown in the screenshot above without anyone ever noticing it. The plone bundle must be around a year old, if all that is true.

or am I missing something? do you use a different approach on updating and compiling (via ``bin/plone-compile-resources -b plone``) than me, @vangheem ?

my ticket about JS cleanup ( https://github.com/plone/mockup/issues/688 ) would help getting this monster better manageable, IMO. /cc @bloodbare

jensens:
well, i'am sure this worked at the time of the commit.

thet:
ok - before @jensens PR it was compilable, as you changed the registry AND the less variables. so it's not SO long, that the plone bundle can't have been compiled....
sorry for the fuzz.
/ref https://github.com/plone/Products.CMFPlone/pull/1426

thet:
Fixed! Toolbar works! 
(A little) nicer than ever including @jensens optimizations (e.g. more contrast in submenus between selected and other list items).


screen shot 2016-06-22 at 00 38 11

thet:
@gforcada can you help here? have you ever seen a failure like this: http://jenkins.plone.org/job/pull-request-5.1/377/testReport/junit/Products.CMFPlone.tests.test_robot/RobotTestCase/Scenario_A_page_is_opened_to_edit_in_TinyMCE/
I can only reproduce locally if I force the browser window to be very tiny. I thought I could fix it by setting the window via the ``Set Window Size`` keyword from to something predefined, but apparently this didn't fix it: https://github.com/plone/Products.CMFPlone/pull/1644/commits/9c04155aaa6d976bbb792893743c54525ad4dbd8

thet:
Squashed commits, added @datakurre advice regarding robot test failure.
Restarting jenkins, crossing fingers.

thet:
@datakurre, can I ask for help again?

The statement ``Wait Until Element Is Not Visible  css=.plone-modal-footer .plone-btn-primary`` did not help, as the overlay ist still there and not being closed (See: https://github.com/plone/Products.CMFPlone/blob/3e202cf3d66281aa493a9c65f61889d96906d2a5/Products/CMFPlone/tests/robot/test_tinymce.robot#L73 ).

This is the screenshot produced: http://jenkins.plone.org/job/pull-request-5.1/378/robot/report/robot_test_tinymce_Scenario_A_page_is_opened_to_edit_in_TinyMCE_selenium-screenshot-1.png
and here is the xml output (search for: 20160622 15:36:01.689): http://jenkins.plone.org/job/pull-request-5.1/378/robot/report/robot_output.xml

I cannot reproduce locally. the ``.plone-modal-footer .pone-btn-primary`` element is clearly visible and should be unique on the page. Inserting a link works, but the image fails.

thet:
Maybe a JavaScript error? Is there a way to see the console? IIRC the firefox on jenkins is really old (~34).
Maybe a side effect of too old robot framework versions? I'm going to fix the one remaining test failure of the robot upgrade PR and see if this fixes this one as a side effect: https://github.com/plone/buildout.coredev/pull/224

jensens:
What a PITA! 

First latest selenium 2.53.3 does not like to play with its mate Firefox 47. Firefox 47+ now uses a complete new remote control system called "Marionette" and does not offer the old and busted WebDriver anymore. This goes with selenium 3 - but there is no selenium 3 released, there is just a beta from Github. I dont have time for such experiments.

So I installed a Firefox 46 on my machine, set the PATH for the test and let it run. All is fine, test is passing. Btw., the same in Chrome, also no problem.

FTR

I downloaded Firefox from https://ftp.mozilla.org/pub/firefox/releases/46.0.1/linux-x86_64/en-US/ and unzipped it in my workspace (renamed directory to ``firefox46``). Then

```
export PATH=/home/jensens/workspace/firefox46:$PATH
./bin/test --all -m Products.CMFPlone -t test_tinymce.robot
```

No idea whats wrong here, except that Jenkins Firefox may need an update? 

/cc @plone/testing-team - any ideas?

jensens:
Yes, we are on FF 34 https://github.com/plone/plone.jenkins_node/blob/5c14ae9205bc388c9161ef96fee18aa057096357/tasks/browsers.yml#L11

thet:
@jensens is currently hunting this bug and tracked it down to be Firefox 34 the cause of it, which produces an error with minified JavaScripts - but not with development mode on.

jensens:
Ok, installed FF34 locally and debugged into the robot test: There is a TypeError in the minified plone-loggedin.js with FF34 only. If I switch the whole into debug mode and all works fine.

What a great bug. Wasted time, because we're using an outdated Firefox nobody out there uses.

datakurre:
Oh, dear :(

I'd like to help, but unfortunately I'm AFK until early July. I also saw Firefox failing again with Selenium. PITA really if they just removed WebDriver. Has been driving me to phantom a where ever it works :(

jensens:
Not possible to fix because its not worth to spent another second on an outdated browser. Added workaround, remove if https://github.com/plone/jenkins.plone.org/issues/179 was solved.

thet:
Tnx for fixing!

thet:
New jenkins run with these PRs:
- https://github.com/plone/Products.CMFPlone/pull/1644
- https://github.com/plone/plone.app.upgrade/pull/79
- https://github.com/plone/buildout.coredev/pull/224

For the records - I'm preparing another update for this PR here. mockup-core is still around and used by the toolbar. I'm going to remove mockup-core, which should be fine for Plone 5.1 and all the "Usage of the mockup-patterns-base pattern is deprecated and it will eventually be removed." messages we got (probably from the toolbar) since mockup 2.0.
I'm wondering again, why I could build the plone bundle without mockup core physically on my hard disk, but still dependend by the toolbar. Though I got a missing mockup-core when building a plone bundle copy located in my project. Misterious things happening here...

thet:
Ready for Review!
/cc @vangheem @bloodbare @jensens

thet:
Tnx for merging @jensens !
And for the records, I removed the mockup-core references in CMFPlone and updated the toolbar accordingly.
I suspect no more "Usage of the mockup-patterns-base pattern is deprecated and it will eventually be removed."  warnings in JS console as a side effect, but let's see...

mauritsvanrees:
Thanks a lot!!

After merge there is one error... :-(
http://jenkins.plone.org/job/plone-5.1-python-2.7-robot/546/robot/Robot/Test%20Querystring/Scenario%3A%20Location%20query/

```
Element locator 'css=li.select2-result-selectable a' did not match any elements
after 30 seconds
```

I wonder if that needs a similar temporary fix like https://github.com/plone/Products.CMFPlone/pull/1644/commits/00a5e1849721b9219a64c3255b2b7373b3a67e50

thet:
Strange that this error didn't pop up in before merging. I'll have a look at this.

thet:
Not sure if that helps: https://github.com/plone/Products.CMFPlone/commit/c7e761a0316fdb6049e7f378fc6b429ccd916f0f
This time commited directly on master to avoid any PR testing side effects.

What's the process again to get firefox 34 to run on my local machine?

But anyways, we have to update firefox on jenkins. this is ridiculous, although i understand it was for the sake of stability.
Who has access to the jenkins servers?

mauritsvanrees:
The same error, except for the changed selector:

```
Element locator 'css=.pattern-relateditems-result-select' did not match any elements
after 30 seconds
```

This has the link to the used Firefox:
https://github.com/plone/plone.jenkins_node/blob/5c14ae9205bc388c9161ef96fee18aa057096357/tasks/browsers.yml#L11

Or one from here: https://ftp.mozilla.org/pub/firefox/releases/34.0.5/

mauritsvanrees:
I installed FF 34.0.5 on my Mac and the test passes...

```
bin/test --all -s Products.CMFPlone -t 'Location query'
```

thet:
OS specific behavior within one version of the browser - that's worse than it was with Internet Explorer.
(If this is really the cause)

Hmhm...

We could use ``Run Keyword And Ignore Error``, but we would also have to ignore the assert after that. And probably subsequent steps.

jensens:
process a documented above: https://github.com/plone/Products.CMFPlone/pull/1644#issuecomment-228013908

PR #1651, opened 2016-06-28
Fix robot tests on 5.0

mauritsvanrees:
Backport from 5.1.
See also https://github.com/plone/buildout.coredev/pull/227

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,7 +19,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Fix select2 related robot test failures and give the test_tinymce.robot scenario a more unique name.
+  [thet]
 
 
 5.0.5c2 (2016-06-22)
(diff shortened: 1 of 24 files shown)


mauritsvanrees:
Locally, with Firefox 46, this sometimes fails and sometimes passes:

```
bin/test --all -s Products.CMFPlone -m test_robot -t test_querystring.robot
...
AssertionError: ValueError: Element locator 'jquery=:focus' did not match any elements.
```

This is presumably in this test:

```
select index type ${INDEX}
    Input Text  jquery=.select2-drop-active[style*="display: block;"] input   text=${INDEX}
    Press Key  jquery=:focus  \\13
```

Adding a `Sleep  1` does not help.

Anyway, the pull request is probably good, but Jenkins cannot find firefox, as indicated in other pull requests.

mauritsvanrees:
This needs https://github.com/plone/plone.app.robotframework/pull/57 and https://github.com/plone/buildout.coredev/pull/227

PR #1654, opened 2016-06-29
Let npm install work on windows for plone-compile-resources

jensens:
On windows command ist ``npm.cmd`` instead of shell script ``npm``.

### DIFF
--- a/Products/CMFPlone/_scripts/compile_resources.py
+++ b/Products/CMFPlone/_scripts/compile_resources.py
@@ -17,7 +17,6 @@ package_json_contents = """{
     "grunt-contrib-requirejs": "~1.0.0",
     "grunt-contrib-uglify": "~1.0.1",
     "grunt-contrib-watch": "~1.0.0",
-    "grunt-debug-task": "~0.1.8",
     "grunt-sed": "~0.1.1",
     "less-plugin-inline-urls": "^1.1.0"
   }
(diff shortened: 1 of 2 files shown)


Products/CMFPlone/_scripts/compile_resources.py
```diff
@@ -129,13 +134,19 @@ def main(argv=sys.argv):
...
         subprocess.check_call(cmd)
 
     # Generate Gruntfile
-    grunt = os.path.join(base_path, 'node_modules', 'grunt-cli', 'bin', 'grunt')
+    grunt = os.path.join(
+        base_path,
+        'node_modules',
+        'grunt-cli',
+        'bin',
+        'grunt'
```
thet: here, no "grunt.cmd" on windows?

thet:
LGTM. test failures seem unrelated - does jenkins have currently some hickup?

mauritsvanrees:
Yes, Jenkins has multiple problems right now. See the latest issues on https://github.com/plone/jenkins.plone.org/issues

jensens:
I also removed the grunt-debug-task - which states itself: This plugin is no longer maintained, use node-inspector directly

jensens:
There are some issues left with the generated Gruntfile. Primary ``os.sep`` related. I can not test them now: no (lack of/unwilling to install) windows here. Yesterday at customer place we had 3 dev on win, so this was what we came up with. I'd like to merge this at least  (it fixes some of the issues) on then go on with this later in a second PR.

Products/CMFPlone/_scripts/compile_resources.py
jensens: grunt is grunt. do not ask why.

PR #1657, opened 2016-06-29
Filter members based on group membership

malthe:
This fixes an issue where all members would be shown, regardless of group membership.


jensens:

Issue #1656, opened 2016-06-29
Robot tests for linkintegrity and querystring are unstable


mauritsvanrees:
This is currently on coredev 5.0 and 5.1 with the robot tests.

```
First failure:
Robot.Test Linkintegrity.Scenario: After you fix linked page no longer show warning
StaleElementReferenceException: Message: Element not found in the cache - 
perhaps the page has changed since it was looked up
Stacktrace:
    at fxdriver.cache.getElementAt (resource://fxdriver/modules/web-element-cache.js:9454)
    at Utils.getElementAt (file:///tmp/tmpdwTdAw/webdriver-py-profilecopy/extensions/fxdriver@googlecode.com/components/command-processor.js:9039)
    at WebElement.getElementTagName (file:///tmp/tmpdwTdAw/webdriver-py-profilecopy/extensions/fxdriver@googlecode.com/components/command-processor.js:12141)
    at DelayedCommand.prototype.executeInternal_/h (file:///tmp/tmpdwTdAw/webdriver-py-profilecopy/extensions/fxdriver@googlecode.com/components/command-processor.js:12661)
    at fxdriver.Timer.prototype.setTimeout/<.notify (file:///tmp/tmpdwTdAw/webdriver-py-profilecopy/extensions/fxdriver@googlecode.com/components/command-processor.js:625)

Second failure:
Robot.Test Querystring.Scenario: Location query
Element locator 'css=.pattern-relateditems-result-select' did not match any elements after 30 seconds
```

See:
- http://jenkins.plone.org/job/plone-5.0-python-2.7-robot/4756/robot/
- http://jenkins.plone.org/job/plone-5.1-python-2.7-robot/561/robot/

These pull requests together passed the tests on 5.1, so they should have helped, but apparently not enough:
- https://github.com/plone/plone.app.robotframework/pull/57
- https://github.com/plone/Products.CMFPlone/pull/1655

More info, explaining those changes, is in this otherwise no longer relevant pull request: https://github.com/plone/Products.CMFPlone/pull/1652

Relevant recent change on Jenkins: Firefox was updated from 34 to 46, with hopes of fixing this or similar failures.

Locally on my Mac the tests pass for me. Well, not always... This should get you the same environment as Jenkins:
- Install https://ftp.mozilla.org/pub/firefox/releases/46.0.1/
[...]

PR #1658, opened 2016-06-30
Marked two robot tests as unstable, non-critical.

mauritsvanrees:
This is the single querystring test, and one of the three linkintegrity tests.

Refs https://github.com/plone/Products.CMFPlone/issues/1656

About half the pull request jobs of 5.0 and 5.1 are failing because of a few unstable robot tests. Best is to fix them. But for now I propose to mark them as non-critical.  Locally it then looks like this when running the linkintegrity and querystring tests, where one non-critical test is failing:

screen shot 2016-06-30 at 15 39 53

The critical tests pass, so Jenkins should be happy. But the non-critical failing tests are still visible, so those persons caring about tests (including me) may still be happy. Let's see how it looks on Jenkins.

This is how it looks on the console:

```
==============================================================================
Test Querystring                                                              
==============================================================================
Scenario: Location query                                              | FAIL |
Element locator 'css=.pattern-relateditems-result-select' did not match any elements
    after 30 seconds
------------------------------------------------------------------------------
Test Querystring                                                      | PASS |
0 critical tests, 0 passed, 0 failed
1 test total, 0 passed, 1 failed
==============================================================================
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,6 +68,9 @@ New features:
 
 Bug fixes:
 
+- Marked two robot tests as unstable, non-critical.
+  Refs https://github.com/plone/Products.CMFPlone/issues/1656  [maurits]
+
 - Use ``Plone Test Setup`` and ``Plone Test Teardown`` from ``plone.app.robotframework`` master.  [maurits]
 
 - Let npm install work on windows for plone-compile-resources.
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
I have added documentation in the failing tests so it is clearer why they have been marked as unstable.  At first I only marked one of the linkintegrity tests as unstable, but now the other two as well: all suffer from this.
- http://jenkins.plone.org/job/pull-request-5.1/411/ (original, two tests marked as unstable)
- http://jenkins.plone.org/job/pull-request-5.1/412/ (new, four tests marked as unstable)

PR #1664, opened 2016-07-07
the advanced_search input is present twice

Gagaro:
The second occurence is [there](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/templates/search.pt#L64-L65).

I think the first one shouldn't be there. We have the advanced_query parameter in double with both False and True as a value if we keep them both.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -156,6 +156,9 @@ Bug fixes:
 - Use zope.interface decorator.
   [gforcada]
 
+- Remove advanced_search input which is in double.
+  [Gagaro]
+
 5.1a1 (2016-03-31)
 ------------------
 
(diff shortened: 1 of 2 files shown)


davilima6:
Lgtm

CHANGES.rst
```diff
@@ -1,4 +1,4 @@
-.. This file should contain the changes for the last release only, which
+[6~].. This file should contain the changes for the last release only, which
```
gforcada: Is that maybe a typo? What is this ``[6~]`` for?

Gagaro: Sorry about that, it's my emacs.

gforcada: no problem, it's always a good idea to double check your commits  before pushing, but that's what pull requests for as well :-)

I started a jenkins job to test the pull request, let's see if jenkins agrees on your changes.

Gagaro: I usually do, but this one was literally 20 seconds before I had to left :yum: . Thanks for the review.

PR #1705, opened 2016-08-07
Fixed adding same resource/bundle to the request multiple times

vangheem:


gforcada:
@vangheem does this need future-porting to master as well?

Issue #1529, opened 2016-04-14
PLIP: Reset `define` in RequireJS after bundles loaded


vangheem:
## Abstract

Reset the requirejs `define` global variable in Plone so add-on developers do not have to learn how to use requirejs if they are including js that expect requirejs to be used.
## Motivation

We have integrators that face `Uncaught Error: Mismatched anonymous define` errors and do not know how to deal with it. While this is documented in RequireJS and there are answers on stackoverflow, it is still annoying and difficult to deal with as an integrator.

Also see: https://community.plone.org/t/third-party-js-add-ons-inside-one-template/1950
## Assumptions
- RequireJS still be used for js dependency resolution in plone
## Proposal & Implementation
- Set `define` to `undefined` after the built resources are included
- Implementation would be in CMFPlone, simply something like this would be added after the resources:

```
<script>
plone_define = define;
define = undefined;
</script>
```
## Deliverables
- a PR with this integrated into Products.CMFPlone
## Risks

This breaks behavior with normal requirejs usage. Developers will need to change how they are using requirejs in their projects if they are using it without the built bundles as `define` will no longer be supported OOTB outside of plone's bundles.

zopyx:
+1 for any solution that would allow us to use (in general arbitrary) JS add-ons in Plone.

rnixx:
There has been a similar workaround in plone with jquery by providing "jq" pointing to "jQuery". People then just started to wrap their JS in anonymous functions.

It might be an option to apply this pattern with require as well

    (function(define, require) {
      require([
        'mockup-pattern-autotoc'
      ], function(AutoTOC){
        // something
      });
    })(plone_define, plone_require);

It would even be possible to generate the wrapper around bundles if desired.

So there's no real code changes needed in recent bundles, no "require" and "define" by default and a clear way to write requirejs using code.

ebrehault:
Does it really have to be a PLIP?
In my opinion, it does not highly impact how Plone works, a PR should be enough.

pigeonflight:
+100

tisto:
+1

ebrehault:
@vangheem The PLIP has been approved by the Framework Team.

jensens:
Is there any implementation/branch already around?

rnixx:
If the title of this PLIP applies it's not what I was suggesting ;)

I was thinking about wrapping each bundle if desired but the "core" bundle, loading the JS basics and immediately undefining ``define`` and ``require``.

[...]

PR #1707, opened 2016-08-07
WIP: Rename the global `define` function, plip #1529

vangheem:
To implement plip: https://github.com/plone/Products.CMFPlone/issues/1529

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -80,6 +80,10 @@ Bug fixes:
 
 Breaking changes:
 
+- Rename the global `define` function after all plone resource registries are
+  added. This is to implement #1529 plip.
+  [vangheem]
+
 - Move toolbar resources to plone-logged-in bundle and recompile bundles.
   [davilima6]
 
--- a/Products/CMFPlone/resources/browser/scripts.pt
+++ b/Products/CMFPlone/resources/browser/scripts.pt
@@ -9,3 +9,19 @@
      </tal:wcondcomment
     ></tal:block
 ></tal:scripts>
+
+<tal:comment tal:replace="nothing">
+Plone resets the RequireJS global "define" function after all the bundles
+are delivered.
+
+The reason for this is to help integrators have less issues dealing with
+RequireJS and 3rd party scripts.
+
+The cases this mostly helps are:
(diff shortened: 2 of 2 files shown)


rnixx:
You simply reset global define AFTER ALL registered resource bundles have been delivered, right?
But then it does not solve our basic problem.

E.g.

https://github.com/collective/collective.js.datatables/blob/master/collective/js/datatables/profiles/plone5/registry.xml <- the JS is delivered as bundle as well...

So to make it work right now we implemented a really bad hack -> https://github.com/collective/collective.js.datatables/blob/master/collective/js/datatables/resources/media/js/jquery.dataTables.min.js#L5

define and require gets unset before the actual script is loaded and then reset to the original values.

That's why I was suggesting to implement a (generated) wrapper around each bundle if desired (except the basic JS, after which define and require gets reset) as described here -> https://github.com/plone/Products.CMFPlone/issues/1529#issuecomment-210304757

Or do I miss something in your approach?

rnixx:
So in order to achieve the flexibility with bundles, merged in one file or not, delivered JS structure may look somthing like:

// plone bundle
* jquery as is
* requirejs as is

// reset define and require

    plone_define = define;
    plone_require = require;

// iterate registered bundle resources

    (function(define, require) { // generated wrapper if defined for bundle in registry (default true)
      require([
        'mockup-pattern-autotoc'
      ], function(AutoTOC){
        // something
      });
    })(plone_define, plone_require); // generated wrapper

    // if no generated wrapper deliver JS as is

// all bundles delivered with addons
...

vangheem:
datatables is a unique case where the packaging is exceptionally difficult and not compatible. If you're packaging inside a bundle with requirejs, there is nothing else, no magic, we can do. Some js packages just suck to work with in packaging.

This PR is only to help those that have trouble including js from manual `<script>` tags or from inside their theme for example.

vangheem:
tests: http://jenkins.plone.org/job/pull-request-5.1/754/

vangheem:
This doesn't always work quite as expected as it can mess with other bundling in other unexpected ways.

I'm putting this PR on hold for now until we can figure out if there is a more safe way.

PR #1733, opened 2016-08-29
Prettify toolbar

thet:
# DONE

- Don't break workflow contentmenu header and better place the current status icon.
- Let the toolbar submenus be as wide as they need by setting a ``min-width`` and ``width: auto;``.
- Raise the width of the expanded toolbar from ``120px`` to ``180px``.
- Make menu hover background fit whole menu width.
- Don't show toolbar scroll bars for bigger screens and make scroll buttons full-width.

# TODO

- [x] Mobile styles are a bit more broken. Has to be fixed.

- [ ] Just checked the left toolbar option, not top one.

- [ ] On normal screens the scrollbars are hidden (no change so far for smaller screens). But now you can't scroll the overflown toolbar with mouse/touchpad.

/cc @jensens @hvelarde @agitator @zopyx 

BTW. any help is appreciated.

screen shot 2016-08-29 at 12 33 12

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,8 @@ New features:
 
   - Make menu hover background fit whole menu width.
 
+  - Don't show toolbar scroll bars for bigger screens and make scroll buttons full-width.
+
   [thet]
 
 Bug fixes:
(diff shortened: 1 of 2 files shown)


thet:
/cc @MatthewWilkes @alecpm 

updated this PR without the nasty autoformating and without the bundle compilation

thet:
dead end...

PR #1754, opened 2016-09-09
Remove LanguageTool layer

gforcada:
It is already removed when upgrading to Plone 5


mauritsvanrees:
I cherry-picked this to the 5.0.x branch in 271f2927e109f158dd8a6c912625fdd4a616a3e2.

I ran into an error on Plone 5.0 only when testing https://github.com/plone/plone.app.upgrade/pull/89 because Plone 5.0 still created this skin layer.

PR #1757, opened 2016-09-11
 Use fork of grunt-sed

gforcada:
The problem with released versions is that they are not compatible with newer grunt versions.

This grunt-sed fork in collective *only* differs from latest official release with the grunt version fixes. There's no new/removed functionality.

/cc @thet can it be merged together with https://github.com/plone/mockup/pull/702 ?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,6 +51,9 @@ Bug fixes:
 - Remove LanguageTool layer.
   [gforcada]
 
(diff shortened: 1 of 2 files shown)


thet:
yes, merge together.
but changelog missing :)

gforcada:
@thet thanks for the reminder, I added the changelog entry and started the pull request job on jenkins

PR #1765, opened 2016-09-16
Apply security hotfix 20160830 for z3c.form widgets. [5.1]

mauritsvanrees:
Small differences compared to the hotfix:
- We also set `form.ignoreRequest = True` instead of only setting this on the widget.
- When `form.ignoreRequest` is True, then we set `widget.ignoreRequest` to True. This avoids the longer taking tests for referrer and method.
- We first check the method, then the referrer, since the method check is cheaper.

Within the security team we were wondering about maybe patching/fixing the form class instead, but that seems hard. At least it seems hard to be sure that we have them all, since various classes inherit from the base z3c form, but override the `update` or `updateWidgets` methods without calling `super`. Well, [plone.z3cform extensible forms](https://github.com/zopefoundation/plone.z3cform/blob/0.9.0/src/plone/z3cform/fieldsets/extensible.py#L63) does call super. But [z3c.form group forms](https://github.com/zopefoundation/z3c.form/blob/3.2.10/src/z3c/form/group.py#L38-L63) don't, although those are sub forms so it might be okay. We could accept the current solution and revisit it later if we want.

Also, it looks like this does need to stay a patch in CMFPlone (or maybe in `plone(.app).z3cform`) and can't be done in `z3c.form` itself: that package should work outside of Plone and CMF too, so without a portal_url tool and without `getToolByName`.

CC @vangheem @davisagli

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,8 @@ New features:
 
 Bug fixes:
 
+- Apply security hotfix 20160830 for ``z3c.form`` widgets.  [maurits]
+
 - Fixed tests in combination with newer CMFFormController which has the hotfix.  [maurits]
 
 - Apply security hotfix 20160830 for ``@@plone-root-login``.  [maurits]
--- a/Products/CMFPlone/patches/__init__.py
+++ b/Products/CMFPlone/patches/__init__.py
@@ -29,3 +29,5 @@ import templatecookcheck        # Make sure templates aren't re-read in
 # production sites
 
 import publishing
+
+import z3c_form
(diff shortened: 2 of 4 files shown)


mauritsvanrees:
Okay, one weird test failure in `Products.CMFPlone.tests.PloneTestCase.FunctionalTestCase.runTest` because `self.runTest` is None:

```
Error in test runTest (Products.CMFPlone.tests.PloneTestCase.FunctionalTestCase)
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
TypeError: 'NoneType' object is not callable
```

On the master branch, when I run `bin/test -s Products.CMFPlone -t runTest` there are zero tests.  On the branch, I get the above failure locally.  The test file I have added is `test_z3c_form_widgets.py`. When I run `bin/test -s Products.CMFPlone -m test_z3c_form_widgets` all is well.

Any ideas?

mauritsvanrees:
Okay, the test setup needed a few changes, using proper layers. Should be fine now. Trying Jenkins again.

Issue #1340, opened 2016-01-26
PLIP: Get rid of portal_quickinstaller


mauritsvanrees:
## Abstract
No longer support installing via Extensions/install.py.

## Motivation
There are two ways of installing products: GenericSetup and CMFQuickInstaller. Checking if a package is currently installed, means having to check portal_setup and portal_quickinstaller, or having code that makes sure those two are in sync. The syncing is mostly in place, with some recent changes, but supporting both remains tricky. Cleanup would make this code much easier to understand.

## Assumptions
Some core packages need to actually get a GenericSetup profile instead of their current install.py (for example Products.Marshall). Some need to get an uninstall profile.

## Proposal & Implementation
- Update prefs_install_products_form browser view to only handle GenericSetup profiles.
- Register this view also as `@@installer` for easier remembering.
- Possibly create a utility instead that handles the basic operations.
- When installing a product, use its `default` profile, instead of using the first profile found.
- Warn before install when a product has no uninstall profile.
- Change any core code that uses the portal_quickinstaller (mostly plone.app.upgrade and tests).
- CMFQuickInstaller code should remain in Plone 5.x, but possibly stripped down and with deprecation warnings. Remove it in Plone 6. (plone.app.upgrade can get bare bones module patches to avoid possible breakage.)
- Document what third party add-ons should do.

## Deliverables
- CMFPlone branch
- plone.app.upgrade branch
- CMFQuickInstaller branch
- Branches for core packages that have no default or no uninstall profile yet.
- Documentation

## Risks
- Third party add-ons that have no profile, are no longer installable.
- We lose the automatic partial cleanup that CMFQuickInstaller does on uninstall (remove added skins, css, etc). The 'partial' part of this may be a benefit: a proper uninstall profile is now required instead of add-ons relying on the default cleanup being good enough.

## Participants
Maurits van Rees

Note that CMFQuickInstaller is Plone specific, in spite of the CMF in the name. So compatibility with CMF is not an issue.

See http://sourceforge.net/p/plone/mailman/message/34390699/ for small initial discussion in August 2015.

mauritsvanrees:
Working on this at the Alpine City Strategic Sprint 2016 in Innsbruck, Austria.
WIP on CMFPlone branch `get-rid-of-qi`:
https://github.com/plone/Products.CMFPlone/tree/get-rid-of-qi
The `prefs_install_products_form` on that branch no longer uses `portal_quickinstaller`.

[...]

PR #1772, opened 2016-09-20
Get rid of qi

jensens:
implementation of PLIP plone/Products.CMFPlone#1340

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,11 @@ Breaking changes:
 
 New features:
 
+- PLIP 1340: Deprecate portal_quickinstaller.
+  You should no longer use CMFQuickInstallerTool methods, but GenericSetup profiles.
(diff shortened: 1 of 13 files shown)


mauritsvanrees:
I (or you) can squash this if you like.

hvelarde:
@mauritsvanrees wouldn't be nice to add a version number on all those warnings to let people know when they have to fix their stuff in advance?

BTW, thank you, now I know how to add warnings to my packages :-)

mauritsvanrees:
@hvelarde What would you suggest as text? Something like this?

"listInstallableProducts is no longer supported since Plone 5.1. It will be removed in Plone 6.0."

mauritsvanrees:
I have updated the warnings along those lines, and squashed the commits.  I will merge all now.

gforcada:
@hvelarde you have http://docs.plone.org/develop/styleguide/deprecation.html as well

PR #1783, opened 2016-09-29
Issue 1600 - Plone 5.1.x - Added options to change default search order

rodfersou:


mauritsvanrees:
@rodfersou I sometimes see test failures on Jenkins for this. For example [this one](http://jenkins.plone.org/job/pull-request-5.1/1000/testReport/junit/Products.CMFPlone.tests.testSearch/TestSection/test_default_search_order_date/).

The `test_default_search_order_date` test seems unstable. I saw a test failure just now locally when running all CMFPlone tests (except robot). But when I ran `bin/test -s Products.CMFPlone -m testSearch` ten times all was well.

Also, the [test](https://github.com/plone/Products.CMFPlone/pull/1783/files#diff-d3abc295ae8a33b5db9470347cc93064R172) looks strange to me. The test layer sets up 100 pages. `my-page1` should be the oldest, `my-page99` the newest. The test sorts by `Date`, which apparently means newest first.

So why does the test expect that `my-page10` is the first result? Shouldn't that be `mypage-99`?

Since the test usually passes, the expected value is correct, but I don't understand why.
And on the failing Jenkins test, `my-page68` is first, followed by 67, etcetera.
When it failed locally, it started with page 74, then 73.

Do you know what is going on here?

Maybe waiting a short time between creating the pages would help, so that the Date index really has different values. That may need to be at least 1 second. With 100 pages this would slow down the test, but to me the 100 looks like overkill anyway. 10 pages seems enough. Or to better check batching, 12 or 15 seems fine. (I know the 100 pages were already there before your pull request.)

When I create 12 pages and wait 1.1 seconds in between, I consistently get `mypage-11` as the first one, followed by 10, etcetera. Well, waiting 0.1 seconds consistently does the trick too for me, so that may be fine.

rodfersou:
@mauritsvanrees good morning

> Also, the test looks strange to me. The test layer sets up 100 pages. my-page1 should be the oldest, my-page99 the newest. The test sorts by Date, which apparently means newest first.
> 
> So why does the test expect that my-page10 is the first result? Shouldn't that be mypage-99?
In this test I changed the date of one item just to make the result different from the relevance option.
At first this makes sense to me.. but now I don't think so.. maybe the best is to create a new set of items specific to these tests and rewrite the tests..

>  Since the test usually passes, the expected value is correct, but I don't understand why.
> And on the failing Jenkins test, my-page68 is first, followed by 67, etcetera.
> When it failed locally, it started with page 74, then 73.
> 
> Do you know what is going on here?
Maybe something change on batch logic? I really don't know..

mauritsvanrees:
I have created a pull request to implement the ideas about improving the tests, as discussed above:
https://github.com/plone/Products.CMFPlone/pull/1865

rodfersou:
@mauritsvanrees thank you

Issue #1742, opened 2016-09-02
PLIP: remove support for old style resource registries in Plone 5.2


zopyx:
## Abstract

Remove/disable support for old-style resource registries (cssregistry.xml and jsregistry.xml) completely.
## Motivation

With Plone 5.1 the support for old-style CSS and Javascript registrations (cssregistry.xml and jsregistry.js) should be dropped completely. Any code/add-on trying to register or update a GenericSetup profile containing such configurations should raise an exception. Besides a speaking error message why the exception occurs ("your using an old-style configuration...") the message should indicate how to proceed instead (e.g. a link to the new-style registry docs). An explicit failure is better here than silently ignoring those files or just logging of the failure.
## Reasons

Dealing with old-style registrations in the context of the new Plone resource registry implementation in terms of the "plone-legacy" bundle has been broken throughout the complete Plone 5.0 releases. Many of these issues are documented on Github and various postings of mine  on community.plone.org. The overall impression is that there is no reasonable solution for fixing the handling of old-style registrations for Plone 5. Fixes are more workarounds and appear like duct-type fixes. Every Plone 5.0.x brought new problems with the support of legacy registrations.

Integrators and maintainers of legacy code need a clear story how to deal with their old code when they need to move forward Plone 5. Javascript and in particular support for arbitrary thirdparzy Javascript modules is essential and crucial nowadays. Although the support for thirdparty Javascript modules in Plone 5 through the new resource registries implementation is still tricky, partly fragile and not consistent, removing another fragile way doing things in Plone is in the interest of every integrator (saving money, resources and motivation).
## Proposal & Implementation

There is no reference implementation for this PLIP. As a first step it should be sufficient to add checks to the exisiting resource registry code base that deals with the plone-legacy bundle or deals with the import of the related XML configuration files. In a later step the codebase could be cleaned up by removing all plone-legacy bundle related code.
## Risks

This PLIP will clearly break all-ons that would possibly work with Plone 5.0 with the plone-legacy bundle approach. However there is no way to make the legacy bundle approach work under all circumstances. Stability and consistent behavior of the Plone installation is king - both for end-users, developers and integrators - in particular a must for an enterprise-grade CMS.
## Deliverables
- PR removing/disabling cssregistry.xml and jsregistry.xml
- Documentation how migrate old code and old-style registrations to the new Plone 5 RR implementation. The documention partly exists.
[...]

PR #1784, opened 2016-09-30
Fix plone-legacy RequireJS errors in development mode

thet:
Reset and re-enable ``define`` and ``require`` for the ``plone-legacy`` bundle in development mode.
Fixes issues with legacy scripts having RequireJS integration in development mode.
In Production mode, resetting  and re-enabling is done in the compiled bundle.

Please note: This does only work for ``plone-legacy`` and not for other "legacy" bundles with "compile" set to False.
I couldn't find a way quickly, to reliably check the compiled attribute of the bundle object in resources/browser/scripts.py - looks like it always returns False in development mode - so this doesn't help much.

~~I just recognized, that I do not check if the plone-legacy bundle itself is in development mode - I just check if the resourceregistry is. That has to be fixed I guess.~~ ... it is checked....

But for the general fix: Comments welcome!

/cc @jensens @vangheem 

This addresses: 
https://github.com/plone/Products.CMFPlone/issues/1529
https://github.com/plone/Products.CMFPlone/issues/1742

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,10 @@ Bug fixes:
 - Add utf8 headers to all Python source files. [jensens]
 
 - Add default icon for top-level contentview and contentmenu toolbar entries [alecm]
+- Reset and re-enable ``define`` and ``require`` for the ``plone-legacy`` bundle in development mode.
+  Fixes issues with legacy scripts having RequireJS integration in development mode.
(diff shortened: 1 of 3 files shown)


thet:
@vangheem this now redefines define and require for all bundles in develop javascript mode, if compile is set to false. can we also do that for bundles, which are actually not legacy bundles and just built outside plone?

PR #1786, opened 2016-09-30
add utf8 headers to all python files

jensens:
This should not break existing pull requests since its easy to merge, but is a first little step towards following in Products.CMFPlone our own code style guides.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,8 @@ New features:
 
 Bug fixes:
 
+- Add utf8 headers to all Python source files. [jensens]
+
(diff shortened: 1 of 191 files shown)


gforcada:
@jensens thanks for doing that! Did you use any generic script that could be used for other packages as well?

jensens:
Just a tiny script 
https://gist.github.com/jensens/43965a2d25c26dd691fd549ba492fd48


Am 2. Oktober 2016 11:20:13 vorm. schrieb Gil Forcada Codinachs 
<notifications@github.com>:

PR #1793, opened 2016-10-16
avoid dependency on plone.app.imaging

davisagli:
We don't want to depend on plone.app.imaging, because it requires Archetypes.

This copies a permission and the implementation of 2 util functions from plone.app.imaging so we can stop depending on it.

See also https://github.com/plone/plone.app.vocabularies/pull/40

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,8 @@ New features:
 
 Bug fixes:
 
+- Avoid dependency on plone.app.imaging. [davisagli]
+
 - Add utf8 headers to all Python source files. [jensens]
 
 - Apply security hotfix 20160830 for ``z3c.form`` widgets.  [maurits]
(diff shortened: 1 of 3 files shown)


thet:
The test failures seem unrelated.
I run the test myself after merging the other related PR's.

thet:
Still failing with the same errors...

davisagli:
Missing zcml prefix. Trying again.

thet:
Outch, CMFPlone still imports from ``plone.app.imaging``. But it really should be the other way around:

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/utils.py#L670 for ``getAllowedSizes`` and ``getQuality``.

They import from:
https://github.com/plone/plone.app.imaging/blob/master/src/plone/app/imaging/utils.py

davisagli:
That's what I'm trying to fix here, but I missed a few things. Will revisit later today.

David Glick
(mobile)

davisagli:
Now it passed.

PR #1799, opened 2016-10-20
[WIP] Move PasswordResetTool to CMFPlone

tomgross:
Move PasswordResetTool to CMFPlone since there are interdependencies anyway.
This PR additionally does:

 - Move ZMI / DTML configuration form into a Plone page
 - Move CMFFormController to BrowserViews
 - Replace djange_randomstring with plone.uuid generator

With this PR merged Products.PasswordResetTool is no longer necessary.


jensens:
I know that some addons are (ab)using PasswordResetTool for other similar tasks. But I can not find code importing from there, just using the tool instance. So this should be not a problem to merge once finished.

Products/CMFPlone/PasswordResetTool.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def manage_toggleUserCheck(self, REQUEST=None):
+        """ZMI method for toggling the flag for checking user names on return.
+        """
+        self.toggleUserCheck()
+        m = self.checkUser() and 'on' or 'off'
+        return self.manage_overview(manage_tabs_message="Returning username check turned %s" % m)
+
+    def __init__(self):
+        self._requests = {}
```
davisagli: While we're moving the password tool, let's change this to a BTree. We had a customer with a site that had self-registration turned on and was getting spam accounts created, and the database grew rapidly because each new user saved a new copy of this entire dict.

Products/CMFPlone/RegistrationTool.py
```diff
@@ -422,9 +422,10 @@ def registeredNotify(self, new_member_id):
         # don't need to worry about 'UseMailHost' permissions).
         registry = getUtility(IRegistry)
         encoding = registry.get('plone.email_charset', 'utf-8')
-        mail_text = self.registered_notify_template(
-            self, self.REQUEST, member=member, reset=reset, email=email,
-            charset=encoding)
+        registered_notify_template = getMultiAdapter(
```
davisagli: This sort of lookup would be more readable if we had a "getView" utility function in CMFPlone.

Products/CMFPlone/profiles/default/toolset.xml
```diff
@@ -31,7 +31,7 @@
  <required tool_id="portal_migration"
            class="Products.CMFPlone.MigrationTool.MigrationTool"/>
  <required tool_id="portal_password_reset"
-           class="Products.PasswordResetTool.PasswordResetTool.PasswordResetTool"/>
+           class="Products.CMFPlone.PasswordResetTool.PasswordResetTool"/>
```
davisagli: Remember this will require an upgrade step in plone.app.upgrade.

Products/CMFPlone/tests/test_functional.py
```diff
@@ -10,7 +10,8 @@
 from Testing.ZopeTestCase import ZopeDocFileSuite
 
 
-UNITTESTS = ['messages.txt', 'mails.txt', 'emaillogin.txt', 'translate.txt']
+UNITTESTS = ['messages.txt', 'mails.txt', 'emaillogin.txt', 'translate.txt',
+             'pwreset_browser.txt']
```
davisagli: Are you up for converting the doctest to unit tests?

Products/CMFPlone/RegistrationTool.py
tomgross: On the first glance this sounds tempting but getting the multi-adapter is well documented and changing it would lead to different patterns all over the place.

davisagli: Yeah, I realize the suggestion is outside the scope of this pull request. But I think it's something we should consider changing in all those places. :)

tomgross:
Superseded by #1809

PR #1802, opened 2016-10-21
get rid of completly (in whole core) unused plone_scripts

jensens:


gforcada:
@jensens  plone_log is still used at https://github.com/plone/Products.CMFPlone/blob/85475602e9e6d0f0861137bdea7b485ae2406e7b/Products/CMFPlone/skins/plone_form_scripts/content_status_modify.cpy

PR #1809, opened 2016-10-23
Move PasswordResetTool from own product to here

tomgross:
Supersedes #1799

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,8 @@ New features:
 - Include mockup 2.4.0.
   [thet]
 
+- PasswordResetTool moved from its own package to here
+  [tomgross]
 - Prevent workflow menu overflowing in toolbar [MatthewWilkes]
 
 - Add default icon for top-level contentview and contentmenu toolbar entries [alecm]
--- a/Products/CMFPlone/PasswordResetTool.py
+++ b/Products/CMFPlone/PasswordResetTool.py
@@ -1,0 +1,247 @@
+"""PasswordResetTool.py
+
+Mailback password reset product for CMF.
+Author: J Cameron Cooper, Sept 2003
+"""
+from Products.CMFCore.utils import UniqueObject
+from Products.CMFCore.utils import getToolByName
+from OFS.SimpleItem import SimpleItem
+from App.class_init import InitializeClass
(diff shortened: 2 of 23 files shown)


tomgross:
Tests are green (https://github.com/plone/buildout.coredev/pull/261): Please merge until it fails again ;-)
It needs 

 - [ ] #1809
 - [ ] plone/plone.app.upgrade#92
 - [ ] https://github.com/plone/buildout.coredev/pull/261 (only 3e646ef8647b89142df99c8fc27a3dddbac2fc47) 

Thanks
Tom

jensens:
cool, like to merge, but plone/plone.app.upgrade#92 is not green

tomgross:
@jensens Thanks. The PR in plone/plone.app.upgrade and plone/Products.CMFPlone depend on each other and fail therefore individually. There reference is the PR in plone/buildout.coredev#261, which combines both PRs and turned now green.

gforcada:
@tomgross as Jens already said on another PR you can test multiple pull requests in a single job, just put each PR URL on a line, the text between the textarea and the button says that already :-)

tomgross:
@gforcada I know but this didn't work for the buildout itself, just for the products. I guess this is an edge case. The problem is that PasswordResetTool is referenced in the test part of buildout and the tests fail but with this PRs merged the product shouldn't be there alltogether.

jensens:
I see, this is indeed an edge case. Well, it was green and the p.a.upgrade is green in 5.0 branch as well, so I think we're safe to merge

jensens:
rebased, merged p.a..upgrade and removed from sources.cfg

I hope jenkins stays green after all 3 are run.

tomgross:
@jensens @gforcada Thanks for the review, the tipps and the merge. The build indeed broke because of the reference to PasswordResetTool in the buildout. I removed it in https://github.com/plone/buildout.coredev/commit/8c168f7f6434633bab234caa1fa0d316a4bb8d81 and triggered the build http://jenkins.plone.org/job/plone-5.1-python-2.7/1005/
I expect it to return to green since the only failure was in PasswordResetTool.

jensens:
Ok, I've overseen the reference in tests.cfg.thanks for the merge.

PR #1812, opened 2016-10-23
Styling fixes related to plone.app.contentmenu Display menu re-org

alecpm:
Related to https://github.com/plone/plone.app.contentmenu/pull/19


alecpm:
I think this fix along with the plone.app.contentmenu changes should be considered for backport to 5.0.x.

Issue #1801, opened 2016-10-21
list of plone_scripts to get rid of


jensens:
I just checked the state of skins/plone_scripts and it looks there are some low hanging fruits around.

- [ ] ``add_ext_editor.py`` seems completly outdated, I doubt this one is used anywhere
- [ ] ``author_find_content.py`` this is not used on core at any place
- [ ] ``browserDefault.py``

    - more complex to fi- [ ] ``gure out if this is used as a script, because the name is used in different context

- [ ] ``canSelectDefaultPage.py``

    - plone.app.content/plone/app/content/browser/templates/select_default_view.pt
    - Products.CMFPlone/Products/CMFPlone/skins/plone_scripts/canSelectDefaultPage.py
    - Products.CMFPlone/Products/CMFPlone/tests/testBrowserDefaultScripts.py
    - Products.CMFPlone/Products/CMFPlone/tests/testPloneView.py

- [ ] ``check_id.py``

    - various places in AT world

- [ ] ``createMultiColumnList.py``

    - Products.CMFPlone/Products/CMFPlone/browser/templates/search.pt

- [ ] ``create_query_string.py``

    - this is not used on core at any place
    
- [ ] ``ate_components_support.py``

    - Products.Archetypes/Products/Archetypes/browser/templates/calendar_macros.pt

- [ ] ``displayContentsTab.py``

    - only place this string appears is Products.CMFPlone/Products/CMFPlone/browser/interfaces.py where it is defined on IPlone interface.
      fishy.

- [ ] ``enableHTTPCompression.py``

    - plone.app.layout/plone/app/layout/viewlets/httpheaders.py

- [ ] ``externalEditorEnabled.py``

    - plone.app.blob/src/plone/app/blob/profiles/file-replacement/types/ATFile.xml
    - plone.app.blob/src/plone/app/blob/profiles/image-replacement/types/ATImage.xml
    - plone.app.blob/src/plone/app/blob/profiles/sample/types/Blob.xml
    - Products.ATContentTypes/Products/ATContentTypes/profiles/default/types/Document.xml
    - Products.ATContentTypes/Products/ATContentTypes/profiles/default/types/Event.xml
    - Products.ATContentTypes/Products/ATContentTypes/profiles/default/types/File.xml
    - Products.ATContentTypes/Products/ATContentTypes/profiles/default/types/Image.xml
    - Products.ATContentTypes/Products/ATContentTypes/profiles/default/types/Link.xml
[...]

PR #1816, opened 2016-10-27
Move script ``date_components_support.py`` to Archetypes 

jensens:
see also #1801


jensens:
ready, merge with plone/Products.Archetypes#68

PR #1823, opened 2016-11-02
pwresettool cleanup and removal of unused security declaration

jensens:


Products/CMFPlone/PasswordResetTool.py
```diff
@@ -134,33 +137,26 @@ def resetPassword(self, userid, randomstring, password):
...
-
+    @security.public
     def getExpirationTimeout(self):
         """Get the length of time a reset request will be valid.
         """
         return self._timedelta
 
-    security.declarePublic('checkUser')
-
+    @security.declarePublic
```
hvelarde: @jensens ssee? here you used `@security.declarePublic`

davisagli: This should be `@security.public`

jensens: Oh yes, this is by mistake and probably the reason why the tests are failing. Thanks for the hint.

Products/CMFPlone/PasswordResetTool.py
```diff
@@ -134,33 +137,26 @@ def resetPassword(self, userid, randomstring, password):
...
     def checkUser(self):
         """Returns a boolean representing the state of 'user check' as described
         in 'toggleUserCheck'. True means on, and is the default."""
         return self._user_check
 
-    security.declarePublic('verifyKey')
-
+    @security.public
```
gforcada: @hvelarde ``security.declarePublic`` and ``@security.public`` are the same with some extra bonus as not being able to make typos (what if you write security.declarePublic('fao') and as @jensens already said, it works on python 3 as well

Issue #1758, opened 2016-09-13
Usage of canSelectDefaultPage Script


tomgross:
In the plone_scripts folder of CMFPlone (5.0.6) there is a script **canSelectDefaultPage**. A grep shows it is only used in two tests. What's the purpose of it? Can it be removed?

ale-rt:
I did a grep in my eggs cache.

It seems it was used in `plone_forms/select_default_view.cpt` at least until Products.CMFPlone-5.0a1:
```
ale@emily:~/.buildout/eggs$ /bin/grep -R canSelectDefaultPage --include="*.cpt" | sort
Products.CMFPlone-4.1.6-py2.7.egg/Products/CMFPlone/skins/plone_forms/select_default_view.cpt:                    <p tal:condition="context/canSelectDefaultPage">
...
Products.CMFPlone-5.0a1-py2.7.egg/Products/CMFPlone/skins/plone_forms/select_default_view.cpt:                    <p tal:condition="context/canSelectDefaultPage">
```

Indeed it is still used:

- https://github.com/plone/plone.app.content/blob/33ea6e38f3393c80583eaac206b25087a795d18b/plone/app/content/browser/templates/select_default_view.pt#L49

But it would be great to remove (or at least deprecate it) it for 5.1.

PR #1832, opened 2016-11-12
remove canSelectDefaultPage script

davisagli:
Fixes https://github.com/plone/Products.CMFPlone/issues/1758

Needs to be tested and merged along with https://github.com/plone/plone.app.content/pull/113


davisagli:
oops, please ignore the 5.0 test run; for 5.0 the change in plone.app.content is fine on its own

PR #1850, opened 2016-11-24
Add support for TinyMCE content CSS set by theme to point to a remote site

datakurre:
Fixes issue where it was not possible to configure tinymce-content-css to remote (e.g. CDN) resource.


datakurre:
FWIW tinymce-styles-css has similar limitation, but I just did not need it here.

PR #1851, opened 2016-11-24
Remove ZopeTestCase, especially ZopeDocFileSuite

mauritsvanrees:
This is work started at the Plone Conference in Boston by @ivanteoh and me, with help from @gforcada.

Note that even after this pull request, lots of tests still use `plone.app.testing.bbb.PloneTestCase`, which means they still ZopeTestCase via the back door. But this removes lots of direct usage of ZopeTestCase.

I have moved `translate.txt` to Products.Archetypes. See pull request https://github.com/plone/Products.Archetypes/pull/71, which I will test together with this one.

@ivanteoh Note that since your last commit I have rebased the branch on master.


mauritsvanrees:
Jenkins tests in combination with https://github.com/plone/Products.Archetypes/pull/71 pass

mauritsvanrees:
BTW, it is fine to merge them independently if wanted. They only belong together because this pull request removes a test and the Archetypes pull request copies it there so it does not get lost.

gforcada:
Great job (as usual)!!

PR #1857, opened 2016-11-30
Load Products.CMFPlone.patches earlier [5.0]

mauritsvanrees:
This is part of PloneHotfix20161129.
It is also subtly related to our removal of the SecureMailHost patch.


mauritsvanrees:
Replaced by #1860.

PR #1858, opened 2016-11-30
Load some patches earlier, instead of in our initialize method.

mauritsvanrees:
This is part of PloneHotfix20161129.

@vangheem Can you check the sanity/security of this? On Plone 5.0 and 5.1 I have made pull requests to simply import *all* patches at the bottom of `CMFPlone/__init__.py`. On 4.3 this gave failures during test setup. So now I only import part of the fixes there. It fixes the z3c.form widget problem from the hotfix.


mauritsvanrees:
In pull request #1857 for 5.0 I simply load all patches at the end of `CMFPlone/__init__.py` but this leads to test failures in plone.app.referenceablebehavior and plone.formwidget.recurrence: http://jenkins.plone.org/job/pull-request-5.0/1635/

So on Plone 5.0 I may want to do the same as in this pull request.
And maybe on 5.1 too, although the tests do pass there and the pull request was merged.

PR #1859, opened 2016-12-01
Fix ResourceRegistry in development mode

jensens:
This one fixes a very annoying behavior of ResourceRegistry (RR) and some related issues. Developing JS and LESS/CSS on the file system and making this changes show up in Plone.

First sorry for mixing cleanup and changes. But it happened...

So this fix:

- makes RR respecting ``debug-mode`` (``zope.conf``, set via ``buildout,cfg`` instance section): like with templates it does not deliver a production mode bundle, but always delivers a fresh bundle. 
- In ``debug-mode`` do not ignore development mode for anonymous users.
- last_compilation date make sense if used with ZODB-stored TTW compiled bundles. If one compiles with ``bin/plone_compile_resources`` its just annoying if one has to reset the RR TTW.

This PR refactors parts of the code. I do not expect anybody has subclassed the existing code, if so it may need minor adjustments. I do not consider this part of Plone as API. 

Besides that all is as it was before, No upgrade step needed, nothing breaks.


thet:
When sending the resource registry in development mode and activating the "develop javascript" mode on a bundle, I now get the following error in the console of the web browser:

```
++resource++plone.js:18 Uncaught TypeError: define is not a function
    at ++resource++plone.js:18
```

Can it be, that you somehow treat the non-legacy/compilable bundle resources like as if they were legacy resources? For legacy resources, ``define`` and ``require`` are unset before inclusion and re-set afterwards.

thet:
^^ fixed in https://github.com/plone/Products.CMFPlone/pull/1883

PR #1868, opened 2016-12-08
WorkflowTool: Include Inactive Content In Worklists

sebasgo:
`WorkflowTool.getWorklistsResults` uses the catalog to gather its results.
By default the catalog tool excludes inactive content from the search
results if the current user lacks the permission to view inactive
content in the current context. This can result in incomplete results,
since the method is supposed to return the work list results for the
whole site.

This commit bypasses the the catalog's inactive content filter.


sebasgo:
@jensens: Same PR with minimal changes (location of `CHANGES.rst` has changed) for master is [here](https://github.com/plone/Products.CMFPlone/pull/1873)

PR #1880, opened 2016-12-21
Don't minify already minified resources

thet:
Don't minify CSS or JavaScript resources if they end with .min.css resp. .min.js.

I think it's safe to assume, that resources with the ending .min.css or .min.js are already minified. There is no need to minify them again. Also, this could be a way to add resources, which cannot be successfully minified, like AngularJS.

Speaking of Angular 1.5.x, slimit minifies the expression
``(0).constructor;`` to ``0.constructor;``, which is nonsense and leads to runtime errors.
So this is a way to include the already minified version of AngularJS without getting this minification error.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,9 @@ Breaking changes:
 
 New features:
 
+- Don't minify CSS or JavaScript resources if they end with ``.min.css`` resp. ``.min.js``.
+  [thet]
+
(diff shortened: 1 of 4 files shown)


thet:
It would actually be better to have an option per resource if it should be minified or not.
But this has some problems:
- The option would only work for legacy/non-compile bundles (AFAIK resources from compile bundles are not minified and cooked like that?)
- Individual CSS files within one resources could just be cooked all together or not at all.

ebrehault:
Looks good to me!

PR #1883, opened 2016-12-22
fix reset/re-enable of RJS for legacy bundles in develop_javascript m…

thet:
…ode. was broken with last code cleanup.

/cc @jensens


thet:
No Changelog entry required, as this is a fix for a feature added in this current unreleased version of Plone.
The feature it refers to is:
```
- Reset and re-enable ``define`` and ``require`` for the ``plone-legacy`` bundle in development mode.
  Fixes issues with legacy scripts having RequireJS integration in development mode.
  In Production mode, resetting  and re-enabling is done in the compiled bundle.
  [thet]
```

PR #1884, opened 2016-12-23
new mockup release

thet:
Refs: https://github.com/plone/mockup/pull/726


thet:
Don't mind the changelog error. there is already a changelog entry presen, which belongs to this change.

PR #1902, opened 2017-01-11
Added `ok` view.

mauritsvanrees:
This is useful for automated checks, for example `httpok`, to see if the site is still available.  It returns the text `OK` and sets headers to avoid caching.

Calling `/ZopeTime` can work too, which is what I currently do as a simple status check, instead of calling the homepage, which could be expensive or could get a cached response even when the Plone Site is down.  But if you use `experimental.publishtraverse`, this will either give you a `NotFound` error, or log a warning each time it is called, due to lacking permissions.  And this may happen in core Plone in the future too.

If this is accepted: would this be acceptable for 5.0 and 4.3 too?

### DIFF
--- a/Products/CMFPlone/browser/okay.py
+++ b/Products/CMFPlone/browser/okay.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from Products.Five.browser import BrowserView
+
+
+class OK(BrowserView):
+    """Returns OK.
+
+    Useful for automated checks, for example httpok, to see if the site
+    is still available.  For this you don't want to query a possibly
+    expensive page.  And you don't want to query anything that may be
+    stored in an intermediate caching server.
+
+    Calling /ZopeTime can work too.  But if you use
+    experimental.publishtraverse, this will either give you a NotFound,
+    or log a warning each time it is called, due to lacking
+    permissions.  And this may happen in core Plone in the future too.
+    """
+
+    def __call__(self):
+        # Make really sure this response is not cached.  This is what
+        # plone/app/caching/operations/utils.py does in the doNotCache
+        # function.
+        set_header = self.request.response.setHeader
+        set_header('Expires', 'Sat, 1 Jan 2000 00:00:00 GMT')
+        set_header('Cache-Control', 'max-age=0, must-revalidate, private')
+        # Return a short and simple message.
+        return u'OK'
--- a/Products/CMFPlone/tests/test_okay.py
+++ b/Products/CMFPlone/tests/test_okay.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+from plone.testing.z2 import Browser
+from Products.CMFPlone.testing import PRODUCTS_CMFPLONE_INTEGRATION_TESTING
+
+import unittest
+
+
+class OkayTest(unittest.TestCase):
+    """Test the OK simple status view."""
+
+    layer = PRODUCTS_CMFPLONE_INTEGRATION_TESTING
+
+    def setUp(self):
+        self.app = self.layer['app']
+        self.portal = self.layer['portal']
+
+    def test_okay_browser(self):
+        browser = Browser(self.app)
+        app_url = self.app.absolute_url()
+        portal_url = self.portal.absolute_url()
+        # Try a couple of urls that should return the same.
+        urls = (
+            app_url + '/@@ok',
+            app_url + '/ok?hello=1',
+            portal_url + '/@@ok',
+            portal_url + '/ok?hello=1',
+        )
+        for url in urls:
+            browser.open(url)
+            self.assertEqual(browser.contents, u'OK')
+            get_header = browser.headers.getheader
+            self.assertEqual(
+                get_header('Expires'), 'Sat, 1 Jan 2000 00:00:00 GMT')
+            self.assertEqual(
+                get_header('Cache-Control'),
+                'max-age=0, must-revalidate, private')
+            # Getting it with a browser gives some more headings than accessing
+            # the view directly.
+            self.assertEqual(get_header('content-length'), '2')
+            # content-type has a charset, but we don't really care about that.
+            self.assertTrue(
+                get_header('content-type').startswith('text/plain'))
+
+    def test_okay_view(self):
+        for page in (self.app, self.portal):
+            view = page.restrictedTraverse('@@ok')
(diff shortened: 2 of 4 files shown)


gforcada:
Seems harmless enough to be backported to 5.0 and 4.3 to me

mauritsvanrees:
BTW, I was curious how fast this was, so I used Apache Bench:

```
ab -n 1000 -c 4 http://127.0.0.1:8080/Plone/@@ok
```

That gave about 300 requests per second on my machine.
Calling it on the Zope root works too: that is about 900 requests per second.
Calling `/ok` instead of `/@@ok` is marginally slower due to the different lookup in the Zope publisher.

Products/CMFPlone/browser/okay.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from Products.Five.browser import BrowserView
+
+
+class OK(BrowserView):
+    """Returns OK.
+
+    Useful for automated checks, for example memmon, to see if the site
```
hvelarde: not memmon but httpok.

Products/CMFPlone/browser/okay.py
```diff
@@ -0,0 +1,25 @@
...
+    Calling /ZopeTime can work too.  But if you use
+    experimental.publishtraverse, this will either give you a NotFound,
+    or log a warning each time it is called, due to lacking
+    permissions.  And this may happen in core Plone in the future too.
+    """
+
+    def __call__(self):
+        # Make really sure this response is not cached.
+        set_header = self.request.response.setHeader
+        set_header('Expires', 'Wed, 11 Jan 2017 22:00:00 GMT')
```
hvelarde: why don't use the same as in other pages: `Sat, 1 Jan 2000 00:00:00 GMT`?

Products/CMFPlone/browser/okay.py
```diff
@@ -0,0 +1,25 @@
...
+        set_header('Cache-Control', 'max-age=0, must-revalidate, private')
```
hvelarde: this should be set to `no-cache, no-store, must-revalidate`, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

Products/CMFPlone/tests/test_okay.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+from plone.testing.z2 import Browser
+from Products.CMFPlone.testing import PRODUCTS_CMFPLONE_INTEGRATION_TESTING
+
+import unittest
+
+
+class OkayTest(unittest.TestCase):
+    """Test the OK simple status view.
```
hvelarde: this one-line docstring must be written `"""Test the OK simple status view."""` instead and have a clean line above it.

Products/CMFPlone/tests/test_okay.py
```diff
@@ -0,0 +1,53 @@
...
+            # the view directly.
+            self.assertEqual(get_header('content-length'), '2')
+            # content-type has a charset, but we don't really care about that.
+            self.assertTrue(
+                get_header('content-type').startswith('text/plain'))
+
+    def test_okay_view(self):
+        for page in (self.app, self.portal):
+            view = page.restrictedTraverse('@@ok')
+            self.assertTrue(view(), u'OK')
```
hvelarde: `self.assertEqual`?

Products/CMFPlone/tests/test_okay.py
davisagli: Empty line above it? https://www.python.org/dev/peps/pep-0257/#one-line-docstrings says no blank line above or below.

Products/CMFPlone/browser/okay.py
mauritsvanrees: You are right, I am mixing up memmon and httpok. Will fix.

Products/CMFPlone/browser/okay.py
mauritsvanrees: I thought it was nice to return the date and time that I created this. But for clarity and easy recognition I'll change it.

Products/CMFPlone/browser/okay.py
mauritsvanrees: I took the value that [plone.app.caching](https://github.com/plone/plone.app.caching/blob/d8c5ae4115a4b06872a3d7092141c540338b398e/plone/app/caching/operations/utils.py#L106-L119) sets. I first had a header similar to what you suggest, but according to `plone.app.caching` it can trip up Internet Explorer in some cases.
Not that this is meant to be used by IE or any other browser, but it seems best to stick to what we already do. I will add a note.

Products/CMFPlone/tests/test_okay.py
mauritsvanrees: I sometimes see a blank line above it, and I really do not like it.
But I will make it one line indeed.

Products/CMFPlone/tests/test_okay.py
mauritsvanrees: Ah, correct. I copied another test file that has this same mistake. Will fix.

mauritsvanrees:
I have force pushed changes to address the feedback so far. Thanks!
Previous Jenkins run had one error, but that was an unrelated error in getting a browser. I will run again.

Products/CMFPlone/tests/test_okay.py
hvelarde: @davisagli this is a class, not a method:

> Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.

davisagli: Okay, I don't have a problem with that. But that's a blank line below the docstring, and you said above, which is what I didn't like.

mauritsvanrees: I *did* look at that missing blank line **after**, thinking it might be nicer to add one. Done.

jensens:
LGTM. I'll merge. And I am ok to backport this.

jensens:
This may need some documentation too, right?

sneridagh:
+1

mauritsvanrees:
I made a pull request for [documentation](https://github.com/plone/documentation/pull/752).

PR #1903, opened 2017-01-12
Use correct assertion methods in sitelogo tests.

mauritsvanrees:
We were using assertTrue where assertEqual was obviously meant.

I didn't include a changelog entry, because that is irrelevant here I would say.


hvelarde:
ups, sorry… seems I merged it before tests finished; I hope they'll pass

mauritsvanrees:
Locally they were fine at least. :-)

mauritsvanrees:
And Jenkins has finished successfully too, so all is well. :-)

PR #1912, opened 2017-01-18
Added security checks for str.format. [5.0]

mauritsvanrees:
Part of PloneHotfix20170117.

### DIFF
--- a/Products/CMFPlone/resources/browser/mixins.py
+++ b/Products/CMFPlone/resources/browser/mixins.py
@@ -1,4 +1,5 @@
 from Products.CMFPlone.interfaces import IResourceRegistry
+from Products.CMFPlone.utils import safe_format
 from Products.Five.browser import BrowserView
 from plone.registry.interfaces import IRegistry
 from urlparse import urlparse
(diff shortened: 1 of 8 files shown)


mauritsvanrees:
@vangheem This merges the hotfix into 5.0.

mauritsvanrees:
Note that this adds the formatting function and classes from the hotfix into `utils.py`. So third party code might start relying on this. If we do not want this, then we had better change that now, before merging.

jensens:
Well, utils.py is probably the right location. if we want to indicate this a thing not to use outside we could prefix all with ``_`` - but I am not sure if this is needed/ what we want.

jensens:
Since nobody has an opposite opinion on this I will merge it.

Products/CMFPlone/resources/browser/mixins.py
```diff
@@ -67,9 +68,10 @@ def __call__(self):
             less_vars_params[name] = value
 
         for name, value in registry.items():
-            t = value.format(**less_vars_params)
+            t = SafeFormatter(value).safe_format(**less_vars_params)
```
thet: @mauritsvanrees you have imported ``safe_format`` but using ``SafeFormatter``. This leads to an error mentioned here: https://github.com/plone/Products.CMFPlone/pull/1938#issuecomment-277922063

Products/CMFPlone/resources/browser/mixins.py
```diff
@@ -115,7 +117,7 @@ def __call__(self):
...
```
thet: Same here.

mauritsvanrees:
Huh, strange that this is not caught by the tests.
I have added pull request https://github.com/plone/Products.CMFPlone/pull/1944 which fixes the import and hopefully makes things a bit clearer.

PR #1915, opened 2017-01-20
include new mockup version

thet:
/cc @tomgross


thet:
.... No Changelog needed, as this unreleased version still has a changelog for a new mockup version included.

PR #1922, opened 2017-02-01
Added security checks for str.format.  Part of PloneHotfix20170117. [5.1]

mauritsvanrees:
Port of #1912 for 5.1.

### DIFF
--- a/Products/CMFPlone/resources/browser/mixins.py
+++ b/Products/CMFPlone/resources/browser/mixins.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from plone.registry.interfaces import IRegistry
 from Products.CMFPlone.interfaces import IResourceRegistry
+from Products.CMFPlone.utils import safe_format
 from Products.Five.browser import BrowserView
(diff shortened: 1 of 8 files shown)


Products/CMFPlone/resources/browser/mixins.py
```diff
@@ -68,7 +69,7 @@ def __call__(self):
             less_vars_params[name] = value
 
         for name, value in registry.items():
-            t = value.format(**less_vars_params)
+            t = SafeFormatter(value).safe_format(**less_vars_params)
```
thet: @mauritsvanrees you have imported ``safe_format`` but using ``SafeFormatter``. This leads to an error mentioned here: https://github.com/plone/Products.CMFPlone/pull/1938#issuecomment-277922063

Products/CMFPlone/resources/browser/mixins.py
```diff
@@ -117,7 +118,7 @@ def __call__(self):
...
```
thet: Same here.

PR #1923, opened 2017-02-02
Simplify (DRY) combine bundles and make error handling more explicit.

jensens:
Also explicit raise error if headers are modified.


jensens:
wrong approach, start next simpler

PR #1924, opened 2017-02-02
Do not modify the Content-Type header on bundle combine

jensens:


mauritsvanrees:
I knew I fixed something like this last year. That was in the import step for combining resources. See https://github.com/plone/Products.CMFPlone/pull/1438
There I simply safe the original `Content-Type` header and restore it afterwards.

rnixx:
Maybe using a subrequest to call the resource would be a more clean way to fix this, just as reminder for future

Issue #1931, opened 2017-02-03
Creating a content called 'layout' breaks the folder that contains it


erral:
### What I did:

1. Create a Plone site
2. Add a folder
3. Add a document with the title 'layout'
4. Go to folder's folder_contents 

### What I expect to happen:

I expect to see the contents of the folder

### What actually happened:

maximum recursion depth exceeded

### What version of Plone/ Addons I am using:

Clean Plone 5.0.6 version.

The same happens if you go to edit the recently created folder.

erral:
See the referenced pull request for a robot-test showing how it fails.

jensens:
Partially fixed in Plone 6: "layout" was added to reserved IDs in check_id (plone.base 4.0.0a6, Feb 2026). However, the root cause in CMFDynamicViewFTI fti.py line 123 (getattr(aq_base(context), "layout")) still exists. Content created via API, import, or rename can still trigger the bug.

PR #1932, opened 2017-02-03
Creating a document called 'layout' breaks the parent folder

erral:
See the bug report #1931


erral:
The Jenkins build here is expected to fail

sunew:
@erral So what would be the solution? Making 'layout' a 'reserved word' (making users unable to use it as id), or adding something to the (from title generated) id ("layout_" or similar) ?

("layout" is the zope property used for setting a view for an object)

sunew:
yes and the tools too

davisagli:
Hmm, I thought this check was done by the check_id script https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/skins/plone_scripts/check_id.py but it already includes `layout` in a list of reserved names so I wonder why that's not working.

erral:
The check_id script is also called when renaming a content object through the actions menu or folder_contents.

jensens:

PR #1936, opened 2017-02-04
Test fix (Zope 4 related): More General test if controlpanle back lin…

jensens:
…k URL is ok.


jensens:
unstable social icon robot tests are failing. merge anyway

PR #1945, opened 2017-02-07
Fixed SafeFormatter import in mixins.py. [5.1]

mauritsvanrees:
Clarified safe_format, which is renamed to _safe_format to discourage importing it outside CMFPlone.

Same as #1944 but now for 5.1.


mauritsvanrees:
Note: no changelog entry added, because this changes something that was added but not yet released.

thet:
LGTM

PR #1792, opened 2016-10-13
Restore ability to include head when loading via ajax

displacedaussie:
When loading a page using `ajax_load`, `ajax_main_template.pt` is used instead of `main_template.pt`. The ajax template loads a simplified version of the page.

Previously it was possible to set `ajax_load_with_header`/`ajax_include_head` to include the CSS/JS for the page, but this ability is now missing.

This change restores the ability to load everything in `<head>` (plone.htmlhead, head_slot, javascript_slot, css_slot etc.) if desired.

PR #1948, opened 2017-02-13
Load patterns when using AJAX/iframe, improve AJAX template

displacedaussie:
This pull request follows on from https://github.com/plone/Products.CMFPlone/pull/1792/ which restored the ability to include the `<head>` block when loading a page via AJAX.

The following issues are fixed with this change:

 * JS Patterns are now loaded when a page is loaded via AJAX
 * JS Patterns are now loaded when a page is loaded in an iframe
 * The `<header>` tag is now included in the AJAX template
 * The `visual_portal_wrapper` id is now included in the AJAX template
 * `structure` is now used when loading global status messages (as in main_template.pt)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -153,6 +153,7 @@ New features:
 
 Bug fixes:
 
+- Include JS Patterns when loading a page via ajax or an iframe [displacedaussie]
 
 - Restore ability to include head when loading via ajax [displacedaussie]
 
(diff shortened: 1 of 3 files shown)


ebrehault:
Looks good to me, but I am wondering about the `visual_portal_wrapper` thing, as this is usually how we enable general Diazo rules or not. I do not think it is a big deal because `?ajax_load=1` disables Diazo anyway.
@thet @vangheem any opinion?

Issue #1950, opened 2017-02-16
Circular import for getToolByName patch


mauritsvanrees:
In coredev 5.1 this fails: `bin/test -s plone.app.portlets`:

```
  File "/Users/maurits/shared-eggs/cp27m/zope.configuration-3.7.4-py2.7.egg/zope/configuration/config.py", line 179, in resolve
    mod = __import__(mname, *_import_chickens)
  File "/Users/maurits/community/plone-coredev/5.1/src/plone.protect/plone/protect/auto.py", line 50, in <module>
    from plone.app.blob.content import ATBlob
  File "/Users/maurits/community/plone-coredev/5.1/src/plone.app.blob/src/plone/app/blob/content.py", line 10, in <module>
    from plone.app.blob.markings import markAs
  File "/Users/maurits/community/plone-coredev/5.1/src/plone.app.blob/src/plone/app/blob/markings.py", line 8, in <module>
    from Products.ATContentTypes.interface import file as atfile
ZopeXMLConfigurationError: File "/Users/maurits/community/plone-coredev/5.1/src/plonetheme.barceloneta/plonetheme/barceloneta/configure.zcml", line 11.2-11.41
    ZopeXMLConfigurationError: File "/Users/maurits/shared-eggs/cp27m/plone.app.theming-1.3.5-py2.7.egg/plone/app/theming/configure.zcml", line 19.4-19.44
    ZopeXMLConfigurationError: File "/Users/maurits/shared-eggs/cp27m/plone.app.registry-1.5-py2.7.egg/plone/app/registry/configure.zcml", line 12.4-12.34
    ZopeXMLConfigurationError: File "/Users/maurits/shared-eggs/cp27m/plone.app.registry-1.5-py2.7.egg/plone/app/registry/browser/configure.zcml", line 6.4-6.43
    ZopeXMLConfigurationError: File "/Users/maurits/shared-eggs/cp27m/plone.app.z3cform-2.2.1-py2.7.egg/plone/app/z3cform/configure.zcml", line 10.2-10.41
    ZopeXMLConfigurationError: File "/Users/maurits/community/plone-coredev/5.1/src/plone.app.widgets/plone/app/widgets/configure.zcml", line 12.2-12.41
    ZopeXMLConfigurationError: File "/Users/maurits/community/plone-coredev/5.1/src/Products.CMFPlone/Products/CMFPlone/configure.zcml", line 15.2-15.46
[...]

PR #1951, opened 2017-02-16
Moved getToolByName early patch to the later patches.

mauritsvanrees:
This fixes a circular import.
See #1950.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -153,6 +153,12 @@ New features:
 
 Bug fixes:
 
+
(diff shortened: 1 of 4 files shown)


jensens:
hmm, what was the original reason we had it in the early patches?

mauritsvanrees:
It was in `patches/security.py` and we wanted to make sure the security patches were loaded early. Otherwise some tests in `testSecurity.py` were failing. So we moved over the entire file to `earlypatches`.
For this `getToolByName` patch it doesn't matter exactly when it is loaded.

PR #1952, opened 2017-02-22
AccessInactivePortalContent in catalog queries

thet:
Checked for ``AccessInactivePortalContent`` for each path in a catalog query.
This solves a problem, where Editors couldn't see inactive content, even though they had the required permission on a subpath of the portal (e.g. a subsite).

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -1,13 +1,7 @@
 # -*- coding: utf-8 -*-
-import logging
-import re
-import time
-import urllib
-
 from AccessControl import ClassSecurityInfo
 from AccessControl.PermissionRole import rolesForPermissionOn
-from AccessControl.Permissions import (
-    manage_zcatalog_entries as ManageZCatalogEntries)
+from AccessControl.Permissions import manage_zcatalog_entries as ManageZCatalogEntries  # noqa
 from AccessControl.Permissions import search_zcatalog as SearchZCatalog
 from Acquisition import aq_base
 from Acquisition import aq_inner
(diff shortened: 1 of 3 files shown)


Products/CMFPlone/CatalogTool.py
```diff
@@ -379,6 +379,51 @@ def getCounter(self):
...
+        Conservative check: as soon as one path is disallowed, return False.
+        If a path cannot be traversed, ignore it.
+        """
+        allow_inactive = _checkPermission(AccessInactivePortalContent, self)
+        if allow_inactive:
+            return True
+
+        paths = query_kw.get('path', False)
+        if not paths:
+            return []
```
jensens: Why an empty list here?

Products/CMFPlone/CatalogTool.py
```diff
@@ -379,6 +379,51 @@ def getCounter(self):
...
+                # When no object is found don't raise an error
+                pass
+
+        if not objs:
+            return False
+
+        allow = True
+        for ob in objs:
+            allow = allow and\
+                _checkPermission(AccessInactivePortalContent, ob)
```
jensens: If you have more than one paths given, one not allowed will result in not showing the other path as well. Is it intended? Overall I think it's impossible (or at least difficult to do in a fast way) to implement.

jensens:
@mauritsvanrees since your also deeper into security, what do you think of this change?

Products/CMFPlone/CatalogTool.py
thet: Yes, that's indented - I think it's better to be conservative here instead of showing too much content which shouldn't be shown.
Also, it's quite uncommon that two paths are used.

Products/CMFPlone/CatalogTool.py
thet: That's a leftover from refactoring this method a couple of times... I'll change to ``False``.

thet:
As discussed with @jensens this might need a broader discussion and therefore a PLIP.

As a prerequisite, that this change has any effect to editors is, that the editor role gets the ``AccessInactivePortalContent`` permission set. That way, they can see contents from other editors, which are expired or inactive. AFAIK, the Owner role has the ``AccessInactivePortalContent`` permisson. This has following impacts:

- Editors can see inactive portal content from other editors in the ``folder_contents`` view (which is actually very handy. Imagine someone creating a content or an automated import where the creator is set to the user who imported and another editor is editing the content and setting the publication date to the future. That another editor isn't able to find the content anymore).

As soon they have it, they are able to see expired or not-effective contents from other editors in ``folder_contents`` ().

- Inactive content will show up in listings (custom listings, collections, folder_summary_view, etc). If the number of items is limited (e.g. 10 batch items) it might push other content out of the view. The resulting listing will not correspond to the listing other persons will see, e.g. anonymous visitors to a page. This is actually a problem.

- Also, inactive content should be visually marked as such - like with workflow states.

Part of the solution could be to provide an additional parameter to ``searchResults`` to show inactive content, if the user has permission for it and use it in plone.app.content for the folder_contents listing.

jensens:
While the overall new behavior is correct we have some side effects in existing projects needing some more discussion. 

So please do NOT merge.

thet:
Ideas to visually mark inactive content in folder_contents:
- not-effecive content: In parentheses ``*Title of the content item*``
- expired content: Striked Through ``~~Title of the content item~~``

jensens:
Overall after merging this, the only behavioral change is that editors in local roles are seeing in their section the same as editors in global roles which I think is a good thing. 

The other problems are still there and we already have them for global roles editors. 

I propose to fix these problems in a different PR.

mauritsvanrees:
I didn't have time previously to look at this, but I had a look now, and I agree. Thanks for the fix!

PR #1954, opened 2017-02-25
call processQueue in CatalogTool.search too

davisagli:
A failing test on the thet-zope4 branch (which I rebased) made me notice that this is missing.


davisagli:
(I think this is covered by the existing changelog entry for PLIP 1343, so I didn't add a new one.)

PR #1961, opened 2017-03-07
Legacy bundle cooking improvements

thet:
Resource registry legacy bundle cooking: Exit early with a warning, if preconditions to build are not given (no compilation paths).
Allow cooking CSS, even if no JS is defined.
Log all important steps of the cooking process.

- does the fix, that bundles with only a css resource can be cooked properly documented? or was the inability to do so a bug and a failure regarding this an unexpected behavior?


thet:
@hvelarde good question - i had myself reading code and trying out to make it clear myself.

The best solution IMO is to clear set last_compile of the legacy bundle in registry.xml:
```
  <records
      prefix="plone.bundles/plone-legacy"
      interface="Products.CMFPlone.interfaces.IBundleRegistry"
      purge="False">
    <value key="last_compilation"></value>
  </records>
```
After each profile import, the first time the plone-legacy bundle is accessed, it is re-cooked.
After cooking, the ``last_compilation`` attribute of the bundle is set and for further requests, the already cooked bundle is delivered.

If it's not already documented, this should go into the plone docs.

A button to manually re-cook would be a nice addition to the resource registry control panel.

I'm copying this to https://community.plone.org/t/updating-static-files-with-legacy-bundle-in-plone-5/3579/6

PR #1964, opened 2017-03-09
Wrong use of assertTrue in testResourceRegistries

jensens:


mauritsvanrees:
Well spotted. Change looks good.
The test failure suggests that `insert-before` is not working.
Indeed the existing code treats before and after [the same way](https://github.com/plone/Products.CMFPlone/blob/5.1b2/Products/CMFPlone/resources/exportimport/resourceregistry.py#L133-L144)...

jensens:
Indeed, now it works local. Lets see what Jenkins thinks.

PR #1983, opened 2017-03-28
Add constraint to avoid filling twitter_username field with strings starting with a "@" character

### LINKED ISSUE #1981: No validation is made on Twitter username
### What I did:

I added a `@` char to the Twitter user name in the Social Media configlet.

### What I expect to happen:

An validation error should be shown.

### What actually happened:

I was able to save the invalid user name.

screen shot 2017-03-28 at 13 41 17

Plone 5.1b2

hvelarde:
fixes #1981

### DIFF
--- a/Products/CMFPlone/interfaces/controlpanel.py
+++ b/Products/CMFPlone/interfaces/controlpanel.py
@@ -5,6 +5,7 @@ from Products.CMFPlone import PloneMessageFactory as _
 from zope import schema
 from zope.interface import implementer
 from zope.interface import Interface
+from zope.interface import Invalid
 from zope.schema.vocabulary import SimpleTerm
 from zope.schema.vocabulary import SimpleVocabulary
 
@@ -1457,6 +1458,13 @@ class IUserGroupsSettingsSchema(Interface):
     )
 
 
+def validate_twitter_username(value):
+    if value and value.startswith('@'):
+        raise Invalid(
+            u'Twitter username should not include the "@" prefix character.')
+    return True
+
+
 class ISocialMediaSchema(Interface):
 
     share_social_data = schema.Bool(
@@ -1467,10 +1475,15 @@ class ISocialMediaSchema(Interface):
         default=True)
 
     twitter_username = schema.ASCIILine(
-        title=_(u'Twitter Username'),
-        description=_(u'To identify things like Twitter Cards. Do not include the \'@\' prefix character.'),
+        title=_(u'Twitter username'),
+        description=_(
+            u'To identify things like Twitter Cards. '
+            u'Do not include the "@" prefix character.'
+        ),
         required=False,
-        default='')
+        default='',
+        constraint=validate_twitter_username,
+    )
 
     facebook_app_id = schema.ASCIILine(
         title=_(u'Facebook App ID'),
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Add constraint to avoid filling ``twitter_username`` field with strings starting with a "@" character.
+  [hvelarde]
 
 
 5.1b3 (2017-04-03)


tkimnguyen:
we don't need something like https://github.com/plone/Products.CMFPlone/issues/1981#issuecomment-289981449 ?

hvelarde:
@tkimnguyen no

hvelarde:
@plone/framework-team ready for review and merging.

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -1457,6 +1458,12 @@ class IUserGroupsSettingsSchema(Interface):
     )
 
 
+def validate_twitter_username(value):
+    if value and value.startswith('@'):
+        raise Invalid(u'Twitter username should not include the "@" prefix character.')  # noqa: E501
```
gforcada: you can easily break this into two lines, so no need to add a noqa, exceptions to the rules are for that, for exceptions, not just because I don't want to split a sentence in two.

gforcada: And actually while at it, why do you raise Invalid and rather just remove the ``@`` character if there is any, which would be more helpful.

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -1467,10 +1474,12 @@ class ISocialMediaSchema(Interface):
         default=True)
 
     twitter_username = schema.ASCIILine(
-        title=_(u'Twitter Username'),
-        description=_(u'To identify things like Twitter Cards. Do not include the \'@\' prefix character.'),
+        title=_(u'Twitter username'),
+        description=_(u'To identify things like Twitter Cards. Do not include the "@" prefix character.'),  # noqa: E501
```
gforcada: As said above, splitting a string into two lines is not that much of a problem, rather than adding noqa everywhere.

hvelarde:
@gforcada IMO, this doesn't worth a test.

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -1457,6 +1458,13 @@ class IUserGroupsSettingsSchema(Interface):
     )
 
 
+def validate_twitter_username(value):
+    if value and value.startswith('@'):
+        raise Invalid(
+            u'Twitter username should not include the "@" prefix character.')
```
gforcada: Would you mind changing this to actually remove a leading ``@`` if there is any?

hvelarde: no, because that's not what a validator is supposed to do: a validator must validate, not change the value of an entry; also, what happens if the user adds 2 leading "@" chars? then you will have a lot more logic on it.

tkimnguyen: doesn't lstrip('@') do the trick?

hvelarde: again: a validator must validate, not change the value of an entry.

jensens: Indeed, in our form framework validation and manipulation/extraction of values are two different domains. Do not change the value in a validator. It may introduce some harm.

davisagli:
For a twitter handle field I think automatically fixing the extra @ during save is better UX than showing the user a validation error.

David Glick
(mobile)

hvelarde:
I would have skipped this validator if I had imagined all the discussion in advance ;-)

tkimnguyen:
Hehe this is why I just changed the help text ;)

gforcada:
Computers are supposed to help people do their job rather than tell them to do extra work...

mauritsvanrees:
The new validator is an improvement. Fixing the input would be a nice extra, but not something to do in a validator.

jensens:
Well, everyone is free to create an PR...

hvelarde:
@gforcada no, computers are supposed to take over people's jobs for the sake of the Bezos and Musks of the world.

tkimnguyen:
Well !@$# me, I *did* make the code change but it was not merged https://github.com/plone/plone.app.layout/pull/106

gforcada:
@tkimnguyen thanks for the pull request! Now you need a new one on Castle to remove it as it is done in Plone core itself 😄

tkimnguyen:
Ah no I think Castle uses a git checkout of that tree with my code change...

Issue #1982, opened 2017-03-28
Inline validation fails on Plone 4.3.12 due to bad base tag


mauritsvanrees:
### What I did:

- Plone coredev 4.3 or a plain 4.3.12 (or 4.3.13 pending) instance.
- Create a Plone Site.
- Edit the front page.
- Move the cursor out of the title field.

### What I expect to happen:

- Inline validation kicks in, but does not complain.

### What actually happened:

- Inline validation makes the Title field red, but does not show an error.

Alternatively, if you make the Title field empty and move the cursor out of the field, the field also turns red, but does not show an error.

### Analysis

It goes wrong because [`inline_validation.js`](https://github.com/plone/Products.CMFPlone/blob/4.3.12/Products/CMFPlone/skins/plone_ecmascript/inline_validation.js#L45) calls `<base tag url>/at_validate_field`. Since Plone 4.3.12 this means `http://localhost:8080/Plone/front-page/edit/at_validate_field` instead of `http://localhost:8080/Plone/front-page/at_validate_field`. So we actually get the html of the edit page back via ajax, instead of a json response with a possible validation error.

### Options

- Revert the `base` tag fixes from https://github.com/plone/plone.app.layout/pull/87. But that seems a bad idea and will reintroduce other problems. cc @rodfersou who made this change.
- Fix `inline_validation.js` to not use the `base` tag. Or probably: use it, but strip `/edit` from it. Probably: find the right most slash and remove it and anything after it. I worry that there may be corner cases where we should *not* strip anything.

rodfersou:
Something like this would work: https://github.com/collective/collective.liveblog/blob/master/src/collective/liveblog/browser/templates/update.pt#L136-L140

mauritsvanrees:
Pull request made: https://github.com/plone/Products.CMFPlone/pull/1984

rodfersou:
@mauritsvanrees @gforcada we can close this issue?

mauritsvanrees:
Ah, yes we can.

PR #1984, opened 2017-03-29
Fixed inline validation for Archetypes content. [4.3]

mauritsvanrees:
Every field was shown as having an error, without showing an error message.
Fixes https://github.com/plone/Products.CMFPlone/issues/1982

The wrong behavior was a side effect of the base tag changes in Plone 4.3.12.

Notes:
- dexterity is not affected, because it uses z3c form inline validation
- I did not see errors when viewing PloneFormGen forms, it does not have the `blurrable` class that would be needed for inline validation to kick in.

### DIFF
--- a/Products/CMFPlone/skins/plone_ecmascript/inline_validation.js
+++ b/Products/CMFPlone/skins/plone_ecmascript/inline_validation.js
@@ -29,6 +29,8 @@ jQuery(function ($) {
             $field = $input.closest('.field'),
             uid = $field.attr('data-uid'),
             fname = $field.attr('data-fieldname'),
+            url = '',
+            index = -1,
             value = $input.val();
 
         // value is null for empty multiSelection select, turn it into a [] instead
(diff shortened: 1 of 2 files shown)


rodfersou:
@mauritsvanrees your implementation works, but maybe it is better to use `location.href` instead of `<base>` tag.

We keep the `<base>` tag at Plone 4.3.x just to minimize the problems, but the ideal was to remove this tag the same way we do at Plone 5.x

mauritsvanrees:
True. But `location.href` can contain query parameters, so we would need to strip those off as well. That goes fine with my code as long as the url actually contains 'edit', which would usually be the case for `location.href` as well, but not always.

Actually, I think we can do one better. The formlib and z3c.form validations that are in this same file, use the url of the form action. I have updated the pull request to use this for archetypes as well.

And I have added an extra check: if the response we get has an undefined error message, as happens when you get html back instead of a json message, then we don't display an error.

rodfersou:
Just for the sake of curiosity, this gist show how to parse the URL, then you can remove the query parameters easily https://gist.github.com/jlong/2428561

mauritsvanrees:
Ah, you pointed to something similar in the issue, but I didn't catch on to how I could use that. I'll try to remember this one for the next time I need something like this. Thanks!

Issue #1863, opened 2016-12-06
Putting space in "Path Identifier" allows site to  be created; however, site does not function


rileydog:
1) when setting up a plone site, if you put a space in the Path identifier the site will be created,
but nothing really works right (too many issues to mention). I know above the Path Identifier entry box it states "No special characters are allowed."
but I didn't think a space was. This could be mitigated by adding "no spaces" to the statement. Best solution: don't allow names with spaces to be saved.

nikandroff:
Hello! I tried to fix this issue today. I found the file and the line responsible for the name validation process. This is 
``` bad_id=re.compile(r'[^a-zA-Z0-9-_~,.$\(\)# @]').search ```
at ObjectManager.py

After testing the behavior I found the following points, that complicate this issue:
1. Not only spaces in the name are harmful for the website creation. Other allowed characters lead to similarly broken website. The characters include ```~.,()#@```

2. Some of (or all) these characters cannot be simply removed from 
``` bad_id=re.compile(r'[^a-zA-Z0-9-_~,.$\(\)# @]').search ```
Apparently they are allowed for a reason, since removal of white space, for example, would lead to inability to create any website, throwing an error for every name:
``` Error Value: The id "Discussion Item" contains characters illegal in URLs. ```

That is my first experience with Plone, so I do not quite understand this behavior. Could you please suggest me why other strings (such as ```Discussion Item```) checked for validity at the same place? What would be a good way to ban space character (and potentially other allowed characters) without breaking validation of other strings?

mikkohu:
The validation done in ObjectManager.py doesn't really seem to be fit for checking path identifier. I'm rather new to Plone too, so I'm not aware of all the id validation rules. It seems however that either ObjectManager should have a separate expression the check path identifier against. Some sort of form validation could also be used on the Create a Plone Site -page to go along the guide text.

tkimnguyen:
It could simply be that no one had tested this in the site creation form and so no one realized we were using the same validator for different purposes.

nikandroff:
@tkimnguyen do you think there is a simple and good solution for this behavior? At this point, I am not sure how this name validation should be dealt with. If that would require some sort of restructuring, maybe the priority of this bug is not high enough to do it just yet.

rileydog:
I realize it is hard to prioritize bugs because we care more about bugs that impact us. This is human 
  I posted this bug and therefore I think it's important ;). 
[...]

PR #1991, opened 2017-03-31
add "no spaces" to help for site ID field

tkimnguyen:
as per https://github.com/plone/Products.CMFPlone/issues/1863#issuecomment-290524823


hvelarde:
I think you must fix the validator also; it has to be easy.

tkimnguyen:
If it were so easy do you think I wouldn't have done it? I'm lazy. Read the thread in the issue. Not going there myself.

jensens:
I would say this is better than nothing and we need a help text anyway. There can be a second PR adding a validation.

PR #1996, opened 2017-04-07
Fix issue where catalog search with path failed when path had inaccessible (private) levels

datakurre:
restrictedTraverse from the navigation root should be used with care, because there might be inaccessible levels for the current user between the root and target.

Please, see how catalog brains do their restricted getObject by doing unrestrictedTraverse to the parent of the target and then restrictedTraverse to the target: 

https://github.com/zopefoundation/Products.ZCatalog/blob/master/src/Products/ZCatalog/CatalogBrains.py#L66

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,10 @@ New features:
 
 Bug fixes:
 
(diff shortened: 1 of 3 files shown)


thet:
Looks good to me.

PR #2000, opened 2017-04-09
new catalog metadata column `MimeType` (dexterity and at) https://git…

fgrcon:
…hub.com/plone/Products.CMFPlone/issues/1995

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,10 @@ New features:
 - Update mockup to latest version.
   [thet]
 
+- new metadata catalog column mime_type
+  https://github.com/plone/Products.CMFPlone/issues/1995
+  [fgrcon]
+
 Bug fixes:
 
 - Adapt tests to render social metadata only if you are anonymous.
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -285,6 +285,11 @@ def getIcon(obj):
 
 
 @indexer(Interface)
+def mime_type(obj):
+    return aq_base(obj).getPrimaryField().getContentType(obj)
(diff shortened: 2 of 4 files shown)


CHANGES.rst
```diff
@@ -17,6 +17,10 @@ New features:
 - Update mockup to latest version.
   [thet]
 
+- new metadata catalog columń MimeType
```
davisagli: typo -- there's an accent on the n

Products/CMFPlone/CatalogTool.py
```diff
@@ -285,6 +285,16 @@ def getIcon(obj):
 
 
 @indexer(Interface)
+def MimeType(obj):
+    if obj.portal_type in ['File','Image']:
```
davisagli: Why limit it to File and Image? Custom types could have a file or image field too. I would check to see if the object has a primary field, and if so use the mimetype from that field's value.

Products/CMFPlone/profiles/default/catalog.xml
```diff
@@ -135,4 +135,5 @@
  <column value="portal_type"/>
  <column value="review_state"/>
  <column value="start"/>
+ <column value="MimeType"/>
```
davisagli: I know we don't really have a convention for whether to use camel case or underscores for catalog metadata columns. But the only ones that are capitalized are named after CMF accessor methods. I would probably call this `mime_type`

fgrcon:
@davisagli  @jensens :   thanks for reviewing  ....

fgrcon:
@jensens  thanks for reviews, done  http://jenkins.plone.org/job/pull-request-5.1/1487

PR #2006, opened 2017-04-11
[WIP] Remove last mockup

gforcada:


gforcada:
@thet so seems I was right, reverting your changes makes Jenkins happy

PR #2021, opened 2017-04-17
Fix default value for "robotx.txt" to avoid issues with content containing "search" in the id

### LINKED ISSUE #1994: robots.txt contradicts sitemap.xml.gz for urls containing the string "search"
### What I did:

Installed Plone 5.0.6 and created a folder called `Research`

### What I expect to happen:

I expected Google would index the contents.

### What actually happened:

Google search console threw an error: that the urls in that folder were in the
`sitemap.xml.gz`, but was Disallowed by `robots.txt`.

Plone 5.0.6

See https://github.com/plone/Products.CMFPlone/commit/03a7670544add6c889ce72391f71f4775929418e

hvelarde:
fixes #1994

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,9 +23,12 @@ New features:
 
 Bug fixes:
 
+- Fix default value for ``robots.txt`` to avoid issues with content containing "search" in the id.
+  [hvelarde]
+
 - Remove references to Products.CMFDefault on meta.zcml
   [gforcada]
-  
+
 - Adapt tests to render social metadata only if you are anonymous.
   [bsuttor]
 
--- a/Products/CMFPlone/interfaces/controlpanel.py
+++ b/Products/CMFPlone/interfaces/controlpanel.py
@@ -39,7 +39,7 @@ Disallow: /*folder_factories$
 Disallow: /*folder_summary_view$
 Disallow: /*login_form$
 Disallow: /*mail_password_form$
-Disallow: /*search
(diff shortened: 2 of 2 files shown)


hvelarde:
@plone/framework-team this is ready.

CHANGES.rst
```diff
@@ -23,9 +23,12 @@ New features:
 
 Bug fixes:
 
+- Fix default value for ``robotx.txt`` to avoid issues with content containing "search" in the id.
```
jensens: typo?

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -39,7 +39,7 @@
 Disallow: /*folder_summary_view$
 Disallow: /*login_form$
 Disallow: /*mail_password_form$
-Disallow: /*search
+Disallow: /@@search
 Disallow: /*search_rss$
```
jensens: are you sure we do not need ``/search`` as well?

hvelarde: I don't think so, because AFAIK all references are to the view are using `@@`; also this is only a list of things to include by default, anybody can make changes to it later.

I saw the `sitemap` view is also available everywhere and I think that's a mistake.

vincentfretin: If you're sure that we use /@@search everywhere, this is fine.

PR #2022, opened 2017-04-17
Fix default value for "robotx.txt" to avoid issues with content containing "search" in the id

hvelarde:
refs. #1994

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,9 @@ New features:
 
 Bug fixes:
 
+- Fix default value for ``robots.txt`` to avoid issues with content containing "search" in the id.
+  [hvelarde]
+
 - Show version of products in Add-ons control panel configlet.
   This fixes https://github.com/plone/Products.CMFPlone/issues/1472.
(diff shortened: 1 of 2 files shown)


hvelarde:
@plone/framework-team this is ready.

CHANGES.rst
```diff
@@ -19,6 +19,9 @@ New features:
 
 Bug fixes:
 
+- Fix default value for ``robotx.txt`` to avoid issues with content containing "search" in the id.
```
vincentfretin: robots.txt you mean?

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -38,7 +38,7 @@
 Disallow: /*folder_summary_view$
 Disallow: /*login_form$
 Disallow: /*mail_password_form$
-Disallow: /*search
+Disallow: /@@search
```
vincentfretin: To be safe, you should also add
Disallow: /search

PR #2025, opened 2017-04-18
Use HTML5 meta charset

malthe:
Using http-equiv is no longer the only way to specify the character set of an HTML document.

### DIFF
--- a/Products/CMFPlone/browser/templates/main_template.pt
+++ b/Products/CMFPlone/browser/templates/main_template.pt
@@ -23,7 +23,7 @@
     <metal:cache tal:replace="structure provider:plone.httpheaders" />
 
   <head>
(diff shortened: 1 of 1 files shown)


Products/CMFPlone/browser/templates/main_template.pt
```diff
@@ -23,7 +23,7 @@
     <metal:cache tal:replace="structure provider:plone.httpheaders" />
 
   <head>
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta charset="UTF-8" />
```
thet: I prefer having ``UTF-8`` written in lower case. It doesn't matter technically, but it's a matter of style.

hvelarde: also, a changelog entry is missing.

jensens:

PR #2045, opened 2017-05-04
update mockup

thet:


Products/CMFPlone/static/plone-compiled.js
```diff
@@ -18928,5 +18937,5 @@ require([
 
 });
 
-define("/home/_thet/data/dev/plone/buildout.coredev/src/Products.CMFPlone/Products/CMFPlone/static/plone.js", function(){});
+define("/home/_thet/data/dev/fhnw/plone/src/Products.CMFPlone/Products/CMFPlone/static/plone.js", function(){});
```
tomgross: I wonder what this absolute path does ...

PR #2054, opened 2017-05-19
Include TinyMCE 4.6.1

frapell:
Include latest TinyMCE


frapell:
Included in mockup in https://github.com/plone/mockup/pull/764

bloodbare:
As @thet merged the update to 4.5.6 now there is a huge conflict. @frapell is there any reason why you need 4.6.1 ?

frapell:
Not really... Just to try and be up to date with tinymce. In fact we
already have tinymce_builded 4.6.2 so this can be thrown away.

I'm seeing the same problems with tests with 4.6.2 so I won't be pushing
that for a while...

thet:
So, I'm closing this one.
Looking forward for  the 4.6.x series to land in Plone, though.

PR #2058, opened 2017-05-24
Review context url for @@render-portlet and @@updateSharingInfo ajax calls

### LINKED ISSUE #2053: @@sharing is broken on Page objects in Plone 4.3.12 and 4.3.14
### What I did:

1. Create vanilla Plone 4.3.14 site.
2. Add private Page with title Test at the top of the site
3. Navigate to sharing tab for that page
4. Type some characters into the Search box
5. Kaboom: exception

### What I expect to happen:

List of potential users as search results

### What actually happened:

Large python back trace because the search form AJAX accessed:

  http://localhost:8080/Plone2/test/@@sharing/@@updateSharingInfo

rather than the following used by Plone 4.3.11:

  http://localhost:8080/Plone2/test/@@updateSharingInfo
 
The root cause appears to be:

https://pypi.python.org/pypi/plone.app.layout/2.3.17

  2.3.15 (2016-06-28)
   Fixes:
   _Fix base tag differs from actual URL (fixes [86](https://github.com/plone/plone.app.layout/issues/86)). [rodfersou]_

which was actually made **after** plone.app.layout 2.3.15 was released, (December 2016): that comment is placed incorrectly in the README file. I'm happy to make a bug report there as well.

Vanilla Plone 4.3.14, which uses  plone.app.layout 2.3.17. So does Plone 4.3.12, and I see exactly the same problem there.

(NB: Pinning plone.app.layout 2.3.15 in Plone 4.3.14 resolves the problem).

Update: appears to be the same issue discussed in: https://github.com/plone/Products.CMFPlone/issues/2051

rodfersou:
Fixes #2053

### DIFF
--- a/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js
+++ b/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js
@@ -1,5 +1,18 @@
 (function($){
 
+function createURL(view){
+    // using this method we keep the get parameters used with plone protect
+    // https://gist.github.com/jlong/2428561
+    if (typeof view === 'undefined') {
+        view = '';
+    }
+    var parser = document.createElement('a');
+    parser.href = location.href;
+    parser.pathname = parser.pathname.replace(/\/@@.*/, '');
+    parser.pathname += view;
+    return parser.href;
+}
+
 function refreshPortlet(hash, _options){
     var options = {
         data: {},
@@ -9,7 +22,7 @@ function refreshPortlet(hash, _options){
     $.extend(options, _options);
     options.data.portlethash = hash;
     ajaxOptions = options.ajaxOptions;
-    ajaxOptions.url = $('base').attr('href') + '/@@render-portlet';
+    ajaxOptions.url = createURL('/@@render-portlet');
     ajaxOptions.success = function(data){
         var container = $('[data-portlethash="' + hash + '"]');
         var portlet = $(data);
@@ -103,7 +116,7 @@ $(document).ready(function(){
 
     $('#content-core').delegate('#sharing-search-button', 'click', function(){
         $.ajax({
-            url: $('base').attr('href') + '/@@updateSharingInfo',
+            url: createURL('/@@updateSharingInfo'),
             data: {
                 search_term: $('#sharing-user-group-search').val(),
                 'form.button.Search': 'Search'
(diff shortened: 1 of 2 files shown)


hvelarde:
can someone in the @plone/framework-team take a look on this?

rodfersou:
@mauritsvanrees could you please help reviewing this code?

mauritsvanrees:
Seems good. Thanks!

mauritsvanrees:
@ezvirtual I cannot reproduce the problem on 4.3.18.
The exact code from this PR is not in that Plone version, but that is because the implementation has changed a bit. See the [history](https://github.com/plone/Products.CMFPlone/commits/4.3.x/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js).

The [new version of the code](https://github.com/plone/Products.CMFPlone/blob/4.3.18/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js#L119) seems okay.
If it does not work for you, then maybe you have some custom changes that prevent setting the `data-base-url` on the `body` tag [here](https://github.com/plone/Products.CMFPlone/blob/4.3.18/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js#L6-L14).

adaugherity:
I just hit this after upgrading a site from 4.3.11 to 4.3.18.  I could not remove group rights from the folder they were defined on, and when assigning new rights, they got stored at the site root rather than the intended location.

With browser dev tools we were able to track this down to the `data-base-url` being set to the site root rather than the current page.  This is due to a naming conflict with plone.protect, which [sets a global var base_url to the site root](https://github.com/plone/plone.protect/blob/3.x/plone/protect/protect.js#L9), which overrides the setting from plone_javascript_variables.  However, even _that_ setting is wrong, as it's the referer rather than the page itself.

I tried disabling CSRF protection (via the environment variable `PLONE_CSRF_DISABLED true`), and then `data-base-url` got set to the referring page rather than the site root.

I eventually found [this](https://github.com/4teamwork/opengever.core/issues/4672#issuecomment-410232485) which fully explained the problem, and mentioned needing to copy the `data-base-url` attribute in your Diazo theme.  This was in fact the solution for us -- after adding that attribute (and data-portal-url for good measure) to the body copy rule in our theme, sharing is now fixed!

**To sum up:** a default installation is not broken, but if you Diazo themes, you must copy these attributes.  There are two bugs in the fallback setting: using HTTP referer rather than the page itself, and a name conflict with plone.protect.

PR #2059, opened 2017-05-29
Remove TinyMCE options from the body.

thet:
Remove TinyMCE pattern options from the body, as these are always set on the richtext fields mimetype selector or - if not there - on the textfield itself.

regarding the mimetype selector managing tinymce options see: https://github.com/plone/plone.app.z3cform/blob/master/plone/app/z3cform/widget.py#L615 (I don't like how the options are managed (mainly because I've written the code myself), but this allows for switching between different editors - plain for .rst, .txt, .md, etc. and tinymce for html mimetypes)

Merge together:
- https://github.com/plone/Products.CMFPlone/pull/2059
- https://github.com/plone/plone.app.contenttypes/pull/403


thet:
Let's see if tests break.

PR #2068, opened 2017-06-12
Cleanup of PLIP #1734

thet:
Followup of https://github.com/plone/Products.CMFPlone/issues/1734

To be merged together:

https://github.com/plone/Products.CMFPlone/pull/2068
https://github.com/plone/plone.app.content/pull/131
https://github.com/plone/plone.app.contenttypes/pull/410
https://github.com/plone/plone.portlet.collection/pull/17
https://github.com/plone/plone.app.event/pull/258
https://github.com/plone/plone.app.portlets/pull/99
https://github.com/plone/mockup/pull/774

PR #2073, opened 2017-06-14
Thet updmockup

thet:


thet:
Changelog is provided by a previous mockup update. This one is necessary for the PR done here: https://github.com/plone/Products.CMFPlone/pull/2068 which was already approved.

PR #2079, opened 2017-06-17
Update mockup

thet:


fgrcon:
@thet Wow  - just tested it manually  - works fine again. BTW Thanks for the immediate merges!

jensens:
changelog not needed - mockup-js changes are documented in mockup.

PR #2082, opened 2017-06-22
Updated jqtree to 1.4.1 to autoscroll for items when dragging.

b4oshany:


datakurre:
Depends on https://github.com/plone/mockup/pull/781

thet:
LGTM

seanupton:
@thet @datakurre @b4oshany does this require that we rebuild compiled mockup assets (and which bundle?) to use this outside of development mode?  I get the sense that this might be a remaining loose end, but I am not sure.  Let me know, I am happy to help if so.

datakurre:
@seanupton I believe that this is only used in thememapper bundle. Oshane needs this for drag and drop support in filemanager pattern used by thememapper.

seanupton:
@datakurre thanks for clarifying.

PR #2085, opened 2017-06-23
Fix default image scale sizes, to fit default theme better

MrTango:
This updates the ols images scale sizes to more useful values and fixes
actually the current Plone 5 default theme like the album view.
Changed scales are:
large 1200:1200
preview 768:768
mini 400:400
thumb 200:200


MrTango:
This has no effect on exiting Plone sites. It will only set the new defaults for new Plone sites.

hvelarde:
quick question, why 1200 and not other value? where did you get that scale from?

jensens:
I am overall +1 for such a change, but I also think it is more complex than it seems.
What are the rationales behind exact these values? 

Scales for a field are stable unless the modified date of an item was changed. 
So that's what happens if the registry values are changed: The old scales are still in place. If an item changes its modified date, the new scale settings will be used. A migration step could walk through all items and directly empty the annotated scales. 

The question is also if someone relies on the default scales for some reason. We should at least add a warning to the migration guide. Integrators should always define scales explicit for their project. I doubt in the real world out there most don't do so.

MrTango:
The most important part is the scale size for thumb which now fits to the default theme and also is more useful for todays screen resolutions. The 1200px size for large is just one example, but everything is better then the very old default of 768px, because 768 is not large at all. Large for me is the big zoomed view of an image. So it's something like 1200px or even 1600px.

The change will have no effect on existing sites nor will it change the existing scales during an upgrade. Except someone is reloading the default profile of Plone. If you someone want to have different sizes, she could always set this in her theme specific profiles.

MrTango:
But of course it makes sense to tell people about it in the Upgrade guides and release notes.

hvelarde:
I understand the motivation, but changing it without and upgrade step will cause problems to people with content already in place; also, my concern is about what size we should use, 768px was selected in the past for some reason: I don't remember which one, but it should be the same reason standard ads were 768px wide.

currently the most commonly used display resolution is 1366x768px and the second one is 1920x1080px; so probably a better size should be 1080px and not 1200px, but the new standard in the ads world is 970px wide.

on the other side, the transition to device-independent pixel makes things more complicated, because a 768x768px image on a screen with pixel density of 3 would be in fact 256x256px if not handled correctly.

so this is the kind of change that seems pretty simple at the beginning, but is in fact way more complex, and that's why I would ask to stand by until things are clearer.

MrTango:
We can also fix the thumb scale size in barceloneta theme, because this actually relies on it.
The current Plone 5 default is broken with the default settings, because it relies on 200px width for thumb (see album_view). If we change thumb to 200px which I would prefere we should also change mini, preview and large so that it makes sense again. But if we woudl change the settings in Barceloneta, it's almost like change it where I did it now, except for installations which are not using the default theme as a base.

I also don't know if the ads size matters here. This is not for adds. The large scale is usually the biggest size you want to show to the user and therefor this deppends on the screen resolutions and typical image sizes. If you want a smaller image, the preview scale would have the same size as large before, so you could use that. The bigger size can be almost fill the screen, which would be proably something bigger than 1280px.
But as I said I don't care much about the exact size of the large scale, as long as it's a lot bigger then 768px.

What I don't understand is, why and what for do you want an upgrae stept for?
I only what to set a new default for new Plone site, not for existing ones.
Or am I missing somthing here?

hvelarde:
an upgrade step will be needed to regenerate the scales in place; I insist this is a breaking change and I don't think is a good idea to do it in Plone 5.x series, ao I would suggest to leave it for Plone 6.

you can anyway change it in your sites safely and easily but messing around other people settings is a bad idea.

I mentioned ads because I don't know how the original scales were chosen on first place, but I suspect it has to do with "standard" sizes and those "standard" sizes only exist in the web advertising world.

we should ask the people involved on the first implementation of image scales why they chose those sizes in the first place.

jensens:
As far as I can think back in Plone time these scales are around, from a time where 800x600 screens where a thing. I am sure it is a good idea to rethink them.

MrTango:
@hvelarde why do you thing it will breake anything?
The change only has an effect for new Plone sites, thats why I don't want to have an upgrade step.
I don't want to change existing sites. the only problem one can have, might be if one creates a new site and uses a theme which expects the old default scales. But this can and should be fixed in the theme. The theme can either use the bigger scales or set them as wanted.

About the scales, I wouldn't overthing this. It 's ws a time where 800x600 was the way after that, it was 1024x768 resp. 960x768 what you wanted to do. Today even this is a bit to small.

For me the most important fix is, that we fix the default theme either in the theme or with the default scales.

hvelarde:
IMO, more people needs to participate on this discussion, as @jensens already said: "is more complex than it seems".

jensens:
Is this still a thing?

thet:
We still have the old image scales... I think this is still a thing.

agitator:

agitator:

agitator:
merge with https://github.com/plone/plone.restapi/pull/757

jensens:
Can we have this in Plone 6 please?

ale-rt:
> merge with
> [plone/plone.restapi#757](https://github.com/plone/plone.restapi/pull/757)

Still open

> [plone/plone.app.contenttypes#527](https://github.com/plone/plone.app.contenttypes/pull/527)

Already merged

> [plone/plone.formwidget.namedfile#40](https://github.com/plone/plone.formwidget.namedfile/pull/40)

Tests running

> [plone/plone.outputfilters#35](https://github.com/plone/plone.outputfilters/pull/35)

Tests running

jensens:
Meanwhile outdated, right?

thet:
This is outdated. We already have updated image sizes in Plone 6 via: https://github.com/plone/Products.CMFPlone/issues/3279

PR #2086, opened 2017-06-26
fix: translate image scales in patterns

Gagaro:


jensens:
Plone 5.0 test must fail since this is Products.CMFPlone master for Plone 5.1 only.

PR #2098, opened 2017-07-06
refactor filter control panel and tests

MrTango:


setup.py
```diff
@@ -37,6 +37,7 @@
             'lxml',
             'mock',
             'plone.app.robotframework>0.9.16',
+            'robotframework-debuglibrary',
```
tisto: @MrTango I don't think this should be declared as a dependency here. I am sure this is already declared somewhere in plone.app.robotframework. cc @datakurre

datakurre: @tisto It's not a default dependency. "Debug" keyword in p.a.robotframework is a bit fancy. It is dependency with [debug] extras, but that can only be defined in buildout.

1. open debuglibrary console when that's available
2. pause with dialog when that dialog is supported by python (it's optional Python build dep)
3. as last fallback stops with pdb

That said, it's not a perfect dependency. A better option would have been to put it somewhere into buildout.coredev so that's its included in modules of bin/test, but we didn't immediately find the correct testrunner recipe option for that.

gforcada:
@MrTango as the branch needed to be rebased, I did it while removing those extra commits of merging with master (pro tip: rebase rather than pull *please!*).

Before force pushing the branch I double checked that your tree and my rebased tree had exactly the same code bit by bit.

PR #2100, opened 2017-07-07
Added ‘Show Toolbar’ permission.

agitator:


datakurre:
Pull is missing description, but this was supposed to be usable with

https://github.com/plone/plone.app.layout/pull/128 and
https://github.com/plone/plone.app.layout/pull/129

to be able to optionally limit toolbar to show e.g. only for Contributors/Editors/Reviewers while only showing a Plone 4 style membertools dropdown for authenticated users (this does not change the default behvior).

thet:
LGTM.
Some documentation on how to use this would be nice, but that's not a blocker.

PR #2101, opened 2017-07-08
* Add data-base-url attribute in body tag closes #2051

### LINKED ISSUE #2051: add data-base-url attribute to HTML body tag in Plone 4
Since plone 4.3.12, the `<base href` attribute in HTML generated by Plone no longer always points to the context URL as it used to prior to the change. This change broke Plone and some add-ons. More breakage may still surface. Fixes have varied because no alternative was provided when the change was made.

For a lengthy background, see the [discussion](https://community.plone.org/t/how-to-get-context-url-in-js-on-plone-4-3-12/4031). 

Rather than rolling back the change which was done to support some other things and would require reverting them, I suggest providing a future-proof alternative (thanks @rodfersou for suggesting using a new attribute):

Plone 5 has removed `<base href` completely. Instead Plone 5 has added a `data-base-url` attribute to the HTML `body` tag. Which points to the context URL.

So, I suggest same be done for Plone 4. That way, anything in Plone core and/or add-ons needing context URL in Javascript have a future-proof way of getting it from here on.

### LINKED ISSUE #2053: @@sharing is broken on Page objects in Plone 4.3.12 and 4.3.14

### LINKED ISSUE #2061: @@plone_lock_operations called for wrong URL
### What I did:

1. Create or edit a Document (or any other type).
2. Wait 5 minutes till the AJAX request @@plone_lock_operations/refreshlock occurs.

### What I expect to happen:

The result of the AJAX request refreshes the lock.

### What actually happened:

The result of the AJAX request is the view of the Document.

Plone version 4.3.14.

--

The main problem is, that url from the base-tag is being used which now points to the actual url instead of the context's url. I guess it can be easily fixed by applying the proposal from #2051 and rewriting [this line](https://github.com/plone/Products.CMFPlone/blob/4.3.14/Products/CMFPlone/skins/plone_ecmascript/unlockOnFormUnload.js#L48).

rodfersou:
* Fix @@plone_lock_operations fixes #2061
* Review @@sharing fixes #2053

### DIFF
--- a/Products/CMFPlone/browser/jsvariables.py
+++ b/Products/CMFPlone/browser/jsvariables.py
@@ -7,6 +7,7 @@ from Products.CMFPlone import PloneMessageFactory as _
 
 TEMPLATE = """\
 var portal_url = '%(portal_url)s';
+var base_url = '%(base_url)s';
 var form_modified_message = '%(form_modified)s';
 var form_resubmit_message = '%(form_resubmit)s';
 var external_links_open_new_window = '%(open_links)s';
@@ -36,6 +37,7 @@ class JSVariables(BrowserView):
 
         props = getToolByName(context, 'portal_properties').site_properties
         portal_url = getToolByName(context, 'portal_url')()
+        base_url = self.request['HTTP_REFERER']
 
         # the following are flags for mark_special_links.js
         # links get the target="_blank" attribute
@@ -54,6 +56,7 @@ class JSVariables(BrowserView):
 
         return TEMPLATE % dict(
             portal_url=portal_url,
+            base_url=base_url,
             open_links=open_links,
             mark_links=mark_links,
             form_modified=form_modified,
--- a/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js
+++ b/Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js
@@ -1,18 +1,5 @@
 (function($){
 
-function createURL(view){
-    // using this method we keep the get parameters used with plone protect
-    // https://gist.github.com/jlong/2428561
-    if (typeof view === 'undefined') {
-        view = '';
-    }
-    var parser = document.createElement('a');
-    parser.href = location.href;
-    parser.pathname = parser.pathname.replace(/\/@@.*/, '');
-    parser.pathname += view;
-    return parser.href;
-}
-
 function refreshPortlet(hash, _options){
     var options = {
         data: {},
@@ -22,7 +9,7 @@ function refreshPortlet(hash, _options){
     $.extend(options, _options);
     options.data.portlethash = hash;
     ajaxOptions = options.ajaxOptions;
-    ajaxOptions.url = createURL('/@@render-portlet');
+    ajaxOptions.url = $('body').attr('data-base-url') + '/@@render-portlet';
     ajaxOptions.success = function(data){
         var container = $('[data-portlethash="' + hash + '"]');
         var portlet = $(data);
@@ -116,7 +103,7 @@ $(document).ready(function(){
 
     $('#content-core').delegate('#sharing-search-button', 'click', function(){
         $.ajax({
-            url: createURL('/@@updateSharingInfo'),
+            url: $('body').attr('data-base-url') + '/@@updateSharingInfo',
             data: {
                 search_term: $('#sharing-user-group-search').val(),
                 'form.button.Search': 'Search'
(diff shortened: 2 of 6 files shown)


docs/CHANGES.rst
```diff
@@ -9,7 +9,19 @@ Changelog
 4.3.16 (unreleased)
 -------------------
 
-- Nothing changed yet.
+Bug fixes:
+
+- Add data-base-url attribute in body tag.
+  Closes `issue 2051 <https://github.com/plone/Products.CMFPlone/issues/2051>`
```
mauritsvanrees: The RST link is missing an underscore at the end.

docs/CHANGES.rst
```diff
@@ -9,7 +9,19 @@ Changelog
...
+  [rodfersou]
+
+- Fix @@plone_lock_operations.
+  Fixes `issue 2061 <https://github.com/plone/Products.CMFPlone/issues/2061>`
```
mauritsvanrees: Here too, the RST link is missing an underscore at the end.

rodfersou:
I understand your concerns.. I'll try to create a fallback code then, maybe next week, this week I'm on vacation :wink:

Products/CMFPlone/skins/plone_ecmascript/kss-bbb.js
```diff
@@ -1,17 +1,34 @@
...
-    parser.pathname = parser.pathname.replace(/\/@@.*/, '');
-    parser.pathname += view;
-    return parser.href;
-}
+
+    // Remove views that we know Plone has from the URL
+    var knownViews = [
+        /\/folder_contents/,  // Plone folder_contents view
+        /\/edit/,             // Plone edit page
+        /\/@@.*/              // All other browser views
```
rodfersou: Someone remember other important views to list here?

rodfersou:
@mauritsvanrees done

pgrunewald:
Good job, @rodfersou! :tada: I would like to see this in production, because we are using several workarounds on our site already. I would like to drop this extra maintenance.

pgrunewald:
I happened to notice, if the current URL has an hash, the polyfill will not strip this part. This could lead to problems, when composing it with this schema:

```javascript
$('body').attr('data-base-url') + '/@@viewname'
```
(as it currently happens in many of your PRs)

This results in URLs like: http://localhost:8080/Plone/some/document/#anchor/@@viewname

rodfersou:
@pgrunewald you are totally right, we could add a parser and remove that hash part.

Products/CMFPlone/browser/jsvariables.py
```diff
@@ -36,6 +37,7 @@ def __call__(self, *args, **kwargs):
 
         props = getToolByName(context, 'portal_properties').site_properties
         portal_url = getToolByName(context, 'portal_url')()
+        base_url = self.request['HTTP_REFERER']
```
rodfersou: Can someone please validate this for me, if it is safe to use this request variable? Thank you!

rodfersou:
Oh, now it looks much better without the guessing game :smile:

Products/CMFPlone/browser/jsvariables.py
hvelarde: no, the referrer identifies the URI that linked to the resource being requested; that is, where the request was originated. that's not the `base_url`, AFAIK.

rodfersou: @hvelarde it looks as in our context this works fine (`@jsvariables` is called by jsregistry).

gforcada:
@rodfersou I merged all plone github organization related pull requests, you need to have someone over at the zopefoundation github organization to merge https://github.com/zopefoundation/Products.CMFDefault/pull/1 and then it will be finished 😃

rodfersou:
@gforcada it was already merged AFAIK

rodfersou:
thanks

Issue #1800, opened 2016-10-20
Replace cssmin for something else


gforcada:
It is no longer maintained and it is no python 3 compatible.

A quick search on pypi shows https://pypi.python.org/pypi/rcssmin/1.0.6 as python 2 and 3 compatible... any other solution?

tomgross:
https://pypi.python.org/pypi/cssutils/1.0.1 is another one.

loechel:
I have already tried it, it is actually Python 34-3.6 compatible or needed just a few minutes to fix it, I think I still have it on my notebook. But it is also so old that it has never been flagt as such. but we should replace it totaly, it does not make any sense anymore, if we switch to extrenal JavaScript tools like webpack.

If we want to keep JS and CSS minifing in Plone, then a SASS element would be much better, we might should look if https://pypi.python.org/pypi/pyramid_scss could be integerated

gforcada:
@loechel I would be more than happy if we first to a minimal cssmin to *something* replacement and later on decide if we drop it all together, as that would involve a much more longer and, possibly, heated community discussion.

*Let's keep it simple as we have so much to do* 🙂

loechel:
Ok, I will make a fork in the plone org and make a Python 3 compatible version tomorrow, we might need to look for a new name or release rights.

loechel:
I have also tested to directly replace it by PyScss which works very smooth, but yes one step after each other.

gforcada:
@loechel does it make sense to fork an unmaintained project? Physiologists will call this "Plone syndrome" 😆 keep unmaintained projects alive and kicking 😄 

I would rather just replace it by a maintained *and* python 3 compatible right away. As long as we don't move to python 3, cssmin is still usable and works, so if we want to change something there, let's do it future proof ™️

loechel:
then I would suggest to move to PyScss which could do that easily and some additional things.

gforcada:
Ptscss would not do the trickas our current CSS is on LESS, or does it support LESS as well? On the pypi page it does not say anything

loechel:
@gforcada first of all, PyScss should be the replacement for cssmin, which it would do perfectly, see https://gist.github.com/loechel/a8d037908d0158f14b1841e1bb3d72a8 
But also SCSS is a superset of CSS, SASS and LESS, it could handle our less files. We don't use the edge cases where the compatibility is low.

PyScss support an uptodate version of SCSS and therefor we could benefit from that.

[...]

PR #2107, opened 2017-07-16
replaced cssmin with PyScss for css compression.

loechel:
Replaced cssmin with PyScss to ensure Python 3 compatibility and maintainability. 
Removed dependency to cssmin, so could break dependency for third party addons that depend on it.
Introduced PyScss as a drop in replacement that could also do more things.
Discussion on that at https://github.com/plone/Products.CMFPlone/issues/1800
Pull request requested by @gforcada


loechel:
@gforcada did you see why the Jenkins build failed? there is not enough scroll back, but it seems like the PyScss module is not installed during buildout.

gforcada:
@loechel I hoped you had an idea already, I just saw the error but did not dig further.

As for the full log, if you look closely, at the very top of http://jenkins.plone.org/job/pull-request-5.1/299/console there is a link to http://jenkins.plone.org/job/pull-request-5.1/299/consoleFull

loechel:
Might it be that the pull request did not see the branch at buildout time? so that we need something like a plip.cfg in buildout.coredev to run the buildout against?

loechel:
@gforcada base problem fixed, but there seems to be more errors.

I don't see why the Pull request also tries to load tests for pyScss.
Also there is a UnicodeDecodeError for an mdash which I don't see.

gforcada:
Couldn't be that this dash is coming from the CSS and pyscss needs to have the data encoded?

loechel:
ok @gforcada I have found the error and fixed it. It was actually the other way around pyscss accepts unicode, but also returns unicode and the cooked_css only accepts ascii strings

tests for css minification run locally. 
But test suite is failing again on Products.CMFPlone.tests.testMigrationTool.TestMigrationTool and Products.CMFPlone.tests.test_factory.TestFactoryPloneSite. Should we ignore that?

loechel:
@gforcada anything else on that pull request? Did you need something more to merge?

Products/CMFPlone/resources/browser/cook.py
```diff
@@ -90,28 +92,30 @@ def cookWhenChangingSettings(context, bundle=None):
                 css_url = siteUrl + '/' + css_resource
                 response = subrequest(css_url)
                 if response.status == 200:
-                    logger.info('Cooking css {0}'.format(css_resource))
+                    logger.info('Cooking css %s', css_resource)
                     css = response.getBody()
-                    cooked_css += '\n/* Resource: {0} */\n{1}\n'.format(
+                    if css_resource[-8:] != '.min.css':
+                        if sys.version_info == (2):
```
gforcada: This conditional returns ``False`` on my python 2 prompt, but the comment on the line below implies that it should be true... maybe you wanted to write ``sys.version_info[0] == 2`` ?

loechel:
@gforcada you are right, with the general translation into unicode that check wasn't even necessary anymore. I did remove it. I hope it is now ready for merge.

gforcada:
*while tests are running*

@loechel would you mind adding the version pins needed on buildout.coredev? I see

```
# Required by:
# pyScss==1.3.5
pathlib = 1.0.1

# Required by:
# Products.CMFPlone==5.1b5.dev0
pyScss = 1.3.5
```

on the jenkins output

loechel:
@gforcada done in https://github.com/plone/buildout.coredev/pull/355

Issue #1570, opened 2016-05-03
toolbar should not be shown when user  authenticated but has no edit, contribute or review role for the current page


fgrcon:
Plone 5.x
This is a regression  compared to 4.x

There are many use cases  where users authenticate just in order to be able to see shared content.  The toolbar is useless in such situations, is confusing and is ugly for most themes. Instead of the toolbar a simple dropdown <logged in user name> with user actions like in Plone 4.x  needs to be shown. 

See also http://stackoverflow.com/questions/32824931/how-to-disable-plone-5-toolbar-for-authenticated-user-group/36979732#36979732

jensens:
keep in mind: the personal bar is part of the toolbar.

djay:
I suspect we might need a way to re-enable the old style personal bar for these type situations. I can see lots of situations where you want to use plone for login/logout/change pw etc, but not for letting people edit/workflow and not for your own custom actions.

Possibly just via diazo and theming. but certainly It should be easy to to turn off the toolbar so you can use another form of personal tools. I'm not sure if there should be an automatic, simpler personal tools in such a case though.

fgrcon:
- showing  the toolbar to authenticated users for pages where they dont have edit/add/publish roles seems inacceptable to me.
- so when i hide the toolbar i will  have to provide the personalbar (or just the users name, and links to dashboard (which is still missing) and to preferences.
- I created a **workaround ** by slightly tweaking plone.app.layout the rest is done by some css rules / no diazo modifications needed.
````
remove plone/app/layout/viewlets/anontools.pt
--------------------------------------------------------
added:
+++ b/plone/app/layout/viewlets/personal_bar.pt
@@ -0,0 +1,49 @@
+<div id="personaltools"
+     i18n:domain="plone">
+
+<p class="hiddenStructure" i18n:translate="heading_personal_tools">Personal tools</p>
+
+<dl class="actionMenu deactivated" id="portal-personaltools"
+    tal:condition="python:view.user_actions and not view.anonymous">
+  <tal:block condition="not: view/anonymous">
+  <dt class="actionMenuHeader">
+    <a id="user-name"
+       tal:attributes="href view/homelink_url"
+       tal:content="view/user_name">
+        John
+    </a>
+  </dt>
+  <dd class="actionMenuContent">
+    <ul>
+        <li tal:repeat="action view/user_actions"
+            tal:attributes="id string:personaltools-${action/id}">
+            <a href=""
[...]

Issue #957, opened 2015-09-11
Horizontal toolbar issues


hvelarde:
- [ ] Personal tools overflow (#946)
- [ ] Missing personal toolbar when expanding the toolbar (#950)
- [ ] More… menu completely misplaced (#952)

vangheem:
We've noticed these problems as well. We ended up going with a toolbar on top and bottom. There is just no way to fit everything you need on the toolbar otherwise.

Here is an example:

![screen shot 2016-05-06 at 9 30 35 am](https://cloud.githubusercontent.com/assets/524413/15076081/6d394c4a-136d-11e6-9e84-ad057dd7772e.png)

Then you also have more room to have breadcrumbs and other things in the toolbar.

djay:
I think that part of the problem here is that plugin authors don't have much options with regard to where to make their items available on the toolbar. Greenbar/toolbar tabs has always been easiest. But from a UX perspective this is not ideal. Not everything is top level and needs to be one click away. For example, in that list Sydication, aliases, Analytics and even history I think wouldn't be used often enough. 
I think we can do a few things

1. Do a vote/survey on what are the top 3-4 things people want to be one click away. Make people priotise. That should give us data to work from. Include end users in that as well.
2. Have a way to easily put more forms into the edit view. For example, aliases could be available from edit, since it is conceptually about working with the url of the content. Syndication might be within edit to. Perhaps within edit > settings there could be a menu of extra settings.
3. Have 1-2 generic dropdowns similar to Actions. The idea of having metadata/history dropdown has been suggested. Analytics might fall under that menu. 
4. I've mentioned before how sharing and state are tightly related and can be combined.

rioksane:
I mostly agree with vanhgheem and djay here. After using Plone 5 in various sites in the last months, I agree that there's room for improvement, even though I do like Plone 5 and its new toolbar very much. Right now I'm working on a site that has 11 links on the toolbar - and they all look as important compared to each other. 
<img width="190" alt="plone5-toolbar-crowded" src="https://cloud.githubusercontent.com/assets/5127672/16976299/9d154f76-4e56-11e6-81f6-e7079b3d581a.png">

It doesn't or shouldn't have to be major changes - just make the most important links bold (Add, Edit, Contents) and put some horizontal lines and/or marging to separate primary and secondary function sets. 
[...]

PR #2108, opened 2017-07-20
fix role map for `Show Toolbar Permission`

fgrcon:
see end of discussion of https://github.com/plone/Products.CMFPlone/issues/1570

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,10 @@ New features:
 
 Bug fixes:
 
+- fix role map for `Show Toolbar Permission` 
+  see end of discussion of https://github.com/plone/Products.CMFPlone/issues/1570
+  [fgrcon]
+
 - add :focus class on toolbar for keyboard users  (https://github.com/plone/Products.CMFPlone/issues/1620)
   [polyester]
 
--- a/Products/CMFPlone/profiles/default/rolemap.xml
+++ b/Products/CMFPlone/profiles/default/rolemap.xml
@@ -219,7 +219,12 @@
     </permission>
     <permission name="Show Toolbar"
                 acquire="False">
-      <role name="Authenticated"/>
+        <role name="Manager"/>
+        <role name="Site Administrator"/>
+        <role name="Owner"/>
+        <role name="Reviewer"/>
+        <role name="Editor" />
+        <role name="Contributor" />
     </permission>
   </permissions>
 </rolemap>


hvelarde:
@fgrcon be careful with this: the toolbar must be fixed as a whole and not just to solve your specific use case.

as mentioned before, I agree with @jensens: all authenticated users must see at least a toolbar with the personal bar or we'll be creating new basic usability problems like user doesn't know she is logged-in and user doesn't know how to log out.

if you want to hide the toolbar on your projects, do that as part of your policy.

jensens:
yes, please add Authenticated to the defaults - if you then remove it in your policy thats fine.

fgrcon:
absolutely sure about that. ;-)

fgrcon:
Bye

fgrcon:
Bye

fgrcon:
after a while (but still  a little bit annoyed ...)
@hvelarde: I really would appreciate that you start reading stuff more carefully before commenting. This is a regression in ux/ui. Untill Plone 4.x  an authenticated user only saw a toolbar (edit bar) when he had the appropriate locale roles. 

@jensens:  adding authenticated renders all of  @agitator s changes useless. But if you prefer to force plone users to create policies  for something which really should be default behavior - well ok for me  ..

agitator:
@fgrcon you are right about the "regression"

uiwise there is the problem that the personal actions would switch between top-right and bottom left depending on the users permission, which would be inconsistent.

in my usecase, an intranet, i'm adding additional edit-tools within the content area and never show the toolbar for that role, so that the user doesn't have that inconsistency.

hvelarde:
@fgrcon of course we have a regression and many other problems related to the toolbar: previously we had different elements: an edit bar and a personal bar, among them; today we have only one big toolbar with huge UI/UX problems and I've being pointing to them even before the release of Plone 5: #957.

you are messing around the `Show Toolbar` permissions, not just the edit part of it; you were just about to create more problems, because you're not thinking the whole picture but only in solving your problem.

so start by being careful yourself and be also open to criticism, @jensens had already rose the concern before me and you just ignored both of us.

jensens:
The problem of a more consistent editing UI is bigger than a bunch of small fixes introducing new problems. I hope Pastanaga will help in future. We may also improve the current toolbar, but this need to be done carefully with discussions, probably best at a sprint or conference with a wide variety of Plonistas attending.

Issue #1376, opened 2016-02-11
portal_controlpanel/manage_editActionsForm permissions lost on saving


thet:
When I go to `MYSITE/portal_controlpanel/manage_editActionsForm` - the ZMI portal_controlpanels form - and click on save, the permissions of all configlets are set to "(none)".
This makes it impossible to edit something here.

mauritsvanrees:
This works fine in Plone 4.3, but fails in 5.0. After saving, the `permissions` property of an action is no longer a tuple, but a string. We always take `permissions[0]`, so instead of `Manage portal` we now only get `M`.

Oops, that was me in commit af2bdc30ef. That was just a pyflakes/pep8 commit, but I managed to negate some condition checks... I will fix.

PR #2115, opened 2017-08-02
Fixed removing permissions when saving portal_controlpanel in ZMI. [5.0]

mauritsvanrees:
Fixes https://github.com/plone/Products.CMFPlone/issues/1376 for Plone 5.0.

(Plone 4.3 is not affected.)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,7 +26,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
An obvious fix, and Jenkins passes. I will merge it myself.
Well, after removing the extra text rom the changelog. I never understand how git can possibly think that those lines are part of what I am cherry-picking...

Issue #2105, opened 2017-07-13
sendto_form errors on email_from_name


cdw9:
### What I did:

Trying to use the sendto_form throws an error. Either click the "Send this" link if available, or add `/sendto_form` to the end of a url to get the form. The form loads, but after clicking Send, the page displays the error "Unable to send mail."

### What I expect to happen:

Email should be sent

### What actually happened:

```
2017-07-13 17:26:21 INFO Plone Unable to send mail: 'email_from_name'

 - Expression: "context/email_from_name"
 - Filename:   ... g/Products/CMFPlone/browser/templates/sendto_template.pt
 - Location:   (line 28: col 29)
 - Source:     ... al:replace="structure context/email_from_name" />
                                         ^^^^^^^^^^^^^^^^^^^^^^^
 - Arguments:  repeat: {...} (0)
               template: <ViewPageTemplateFile - at 0x11627dc10>
               utool: <ImplicitAcquisitionWrapper portal_url at 0x114ae28c0>
               views: <ViewMapper - at 0x1172ca790>
               modules: <instance - at 0x10eb184d0>
               args: <tuple - at 0x11691fc90>
               here: <ImplicitAcquisitionWrapper customer-care at 0x114ae2d20>
               user: <ImplicitAcquisitionWrapper - at 0x114ae25a0>
               nothing: <NoneType - at 0x10e04c668>
               charset: utf-8
               dummy: <NoneType - at 0x10e04c668>
               container: <ImplicitAcquisitionWrapper customer-care at 0x114ae2d20>
               request: <instance - at 0x113cccc68>
               wrapped_repeat: <SafeMapping - at 0x1176b2d60>
               traverse_subpath: <list - at 0x11769a050>
               default: <object - at 0x10e13bc50>
               loop: {...} (0)
               context: <ImplicitAcquisitionWrapper customer-care at 0x114ae2d20>
               view: <SendToForm sendto_form at 0x116e5a290>
               translate: <function translate at 0x1176961b8>
               root: <ImplicitAcquisitionWrapper Zope at 0x113884e60>
               options: {...} (6)
               target_language: <NoneType - at 0x10e04c668>
```

### What version of Plone/ Addons I am using:

Plone 5.0 (migrated from Plone 4), 5.1 (brand new site)

### Problem:

Products/CMFPlone/browser/templates/sendto_template.pt is calling `context/email_from_name`. This should be coming from the registry instead.

### Workaround:

[...]

PR #2122, opened 2017-08-21
fix how to get email_from_name information in sendto_form

cekk:
Related to issue #2105 

Now email_from_name is passed with other options.

I wanted to add a basic test for this feature, but i've seen that there isn't any test for this view.
I wanted to add a simple test that sends email after compiling the form.
I think that robots are the way, but i don't know if i should (and how to do it in a robot test) set a fake mailserver.


jensens:
get things done, lets merge it.

PR #2142, opened 2017-09-12
Don't raise unauthorized on show_inactive check

tomgross:
This PR solves an issue with catalog queries containing paths with unaccessible folders in the traversing chain. Without this patch a query to path /Plone/a/b fails with a traceback, if a user has access to folder `b` but not `a`. With this PR merged, `b` is found as it should be.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,8 @@ New features:
 
 Bug fixes:
 
-- *add item here*
+- Don't raise Unauthorized on show_inactive check in catalog search
+  [tomgross]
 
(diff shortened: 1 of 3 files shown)


tomgross:
http://jenkins.plone.org/job/pull-request-5.1/450/

tomgross:
There seems to be an issue with the "Plone Contributors Agreement verifier" check.
/cc @gforcada

gforcada:
@tomgross closing and reopening the pull request triggers it again 🙂

PR #2149, opened 2017-09-19
fix: Remove site path from path in show_inactive in catalog search

Gagaro:
If I have a content named `plone` inside a `plone` site, and I search with a path of `/plone`, the content will be used instead of the site because of the traversing.

We shouldn't include the path of the site if we start from it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,9 @@ New features:
 
 Bug fixes:
 
+- Remove site path from path in show_inactive in catalog search
+  [Gagaro]
+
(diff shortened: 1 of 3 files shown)


hvelarde:
I think this needs to include some tests.

Gagaro:
Test added

Gagaro:
I had 1 test failing before (that I fixed) and I have a lot more now (but I ran them in my IDE the second time), I'll wait for jenkins to have a proper result.

Gagaro:
As I thought, that was the IDE 👍

PR #2153, opened 2017-09-19
use registry folder and split up registry.xml

jensens:


jensens:
this needs to be done at some point but I f** this up here. Its easier to start fresh on 5.2

Issue #2187, opened 2017-10-24
Failing Plone 5.2 test_cooking_missing test due to subrequest handling


mauritsvanrees:
This is currently the [last failing test](http://jenkins.plone.org/job/plone-5.2-python-2.7/lastBuild/testReport/Products.CMFPlone.tests.testResourceRegistries/TestResourceRegistries/test_cooking_missing/) on Jenkins for the main Plone 5.2 job. This fails:

```
  File ".../Products/CMFPlone/tests/testResourceRegistries.py", line 174, in test_cooking_missing
    self.assertTrue('Could not find resource' in resp.getBody())
```

So that is [this line](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/tests/testResourceRegistries.py#L174) in the test. It expects that cooking a non existing resource leads to a response with 'Could not find resource' in the body.

It fails because of some [code in cook.py](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/resources/browser/cook.py#L110-L113). That code uses `subrequest` to get this non existing resource `'++plone++static/foobar.js'`. The result of that subrequest is a response with status 200 and a body `{"error_type": "NotFound"}`.

What apparently *used* to happen is that the response had a different status code (presumably 404) and then you end up in the [part of the code](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/resources/browser/cook.py#L130) that adds the expected 'Could not find resource' in the response.

Quick workaround would be to change the test to expect 'NotFound' in the body text. But the question remains: has something changed in `subrequest` or `Zope` or elsewhere to no longer give a 404 here? For our use case here it seems not so bad, but other code using subrequests may be tripped up when they get a 200 status for a javascript file that is not found.

jensens:
A subrequest has to return sane status codes. Mosaic will get problems too if this happens. So, we need to find the source of the changed behavior.

pbauer:
I'm pretty sure @davisagli knows where this change originates since he already dealt with many issues regarding the changed error-handling in Zope 4.

mauritsvanrees:
[...]

PR #2189, opened 2017-10-24
Set the status of an exception view according to the exception type.

mauritsvanrees:
For Plone 5.2.
Fixes https://github.com/plone/Products.CMFPlone/issues/2187.


jensens:
Great! We may want to backport this to 5.1 as well, right?

mauritsvanrees:
No, this should not be backported to 5.1: the exception view stuff is Zope 4, so not available in Plone 5.1. The `browser/exceptions.py` file is not there.

jensens:
Hmm, ok. This is confusing, I was the opinion it was not set in subrequest, but now I see the difference.

pbauer:
Looks good. The failing robot test in http://jenkins.plone.org/job/pull-request-5.2/41/testReport/junit/Products.CMFPlone.tests.test_robot/RobotTestCase/Scenario_Allow_comments_for_Link_Type/ should be unrelated.

davisagli:
Hmm. I'm confused why this is necessary, because Zope already does it when it renders an exception view: https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/WSGIPublisher.py#L108

Maybe the subrequest is wiping that out?

jensens:
Probably b/c subrequest does call own mapply, not calling all of the publisher.

PR #2190, opened 2017-10-24
Use absolute imports for Python3 compatibility

ale-rt:


CHANGES.rst
```diff
@@ -28,6 +28,9 @@ New Features:
 
 Bug Fixes:
 
+- Use absolute imports for Python3 compatibility
+  [pbauer]
```
mauritsvanrees: Leave Philip out of this. *You* did the work. :-)

Products/CMFPlone/interfaces/__init__.py
```diff
@@ -1,47 +1,47 @@
...
-from events import IReorderedEvent
-from events import ISiteManagerCreatedEvent
-from installable import INonInstallable
-from interface import IInterfaceTool
-from language import ILanguage
-from migration import IMigrationTool
-from patterns import IPatternsSettings
-from password_reset import IPasswordResetToolView
-from password_reset import IPWResetTool
+from Products.CMFPlone.interfaces.basetool import IPloneBaseTool
```
mauritsvanrees: This is fine for me. But alternatively we could do this:

```
from .basetool import IPloneBaseTool
```

That is a relative import, but explicit, so that should be fine for all Python versions.
Is there a reason to prefer one over the other? I ask because I don't know!

ale-rt: AFAIK some people do not like relative imports and also the Google doesn't.

- https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports

I do like them but I do not want to fight for a silly thing.

CHANGES.rst
ale-rt: It was a copy/paste gone mad ;)

Products/CMFPlone/interfaces/__init__.py
gforcada: 👎 on relative explicit imports, if we have to be explicit anyway, better do it fully.

This way sorting is no brainer, QA tools have it easier and things look more harmonious 🙂

PR #2197, opened 2017-10-28
Improve compatibility with Python3

ale-rt:
This fixes json.dumps({}).encode('utf8') that is failing on Python 3.
```
>>> json.dumps({}).decode('utf8')
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode'
```


Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -48,14 +49,23 @@
 """
 
 
+def encode_json(obj):
+    ''' Encode an obj into a text
+    '''
+    text = json.dumps(obj, indent=4)
+    if not isinstance(text, six.text_type):
+        text = text.decode('utf8')
+    return text
```
davisagli: I don't like this name. Encode means going from text to bytes, which is the opposite of what happens here. And it also seems bad to lose "dumps" from the name since that is the main operation happening here. My suggestion would be `dump_json_to_text`. I also think it would be best to put 2-and-3 helpers like this in a `compat` module so that we can reuse them without worrying about import cycles.

ale-rt: And `_compat.dump_json_to_text` will be!

ale-rt:
@davisagli sorry it took so long

PR #2202, opened 2017-10-31
Issue #723: added browser-based CSS hyphenation support for toolbar

### LINKED ISSUE #723: text wrapping in toolbar
See screenshot. The wrapping of the text within the toolbar is completely unacceptable. Either you adjust the width of the toolbar according to the length of the texts or you use hyphenation.
- CSS hyphenation is supported by some browsers (Firefox, Webkit/Safari) but unsupported by Chrome and even latest IE (afaik)
- use a JS based approach (hyphenator.hs) for only this part of Plone...new module, new dependency...
- put the translated terms used within the toolbar into the .po files using soft hyphens.

screenshot-dev1 veit-schiele de 8080 2015-07-18 13-40-02

zopyx:
Fixes #723

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,11 @@ New Features:
 
 Bug Fixes:
 
+- added CSS hyphenation support for toolbar for avoiding ugly
+  text wrapping
+  Fixes `issue 723 <https://github.com/plone/Products.CMFPlone/issues/723>`_.
+  [ajung]
+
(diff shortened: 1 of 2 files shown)


zopyx:
Already done.

jensens:
to fix this in 5.1 we need to cherrypick the commits on 5.1 branch

jensens:
this is only the less file, this needs compilation too, but this can be done together with other changes as well

PR #2205, opened 2017-11-04
Fix getPortalObject when portal_url is fetched as a utility

### LINKED ISSUE #2194: traversal fails with portal from getPortalObject()
In Plone 5.2 with CMF 2.4 traversal with a portal returned by `url_tool.getPortalObject()` fails since it can not acquire the request. `unrestrictedTraverse` needs that though. Found when debugging `test ./bin/test -t testIconLookupForUnknownMimeType`

```
>>> from zope.component import getUtility
>>> from Products.CMFCore.interfaces import IURLTool
>>> portal_url = getUtility(IURLTool)
>>> portal = portal.getPortalObject()
>>> portal.aq_chain
[<PloneSite at /Plone>, <Application at >]

>>> portal.unrestrictedTraverse('foo')
*** AttributeError: REQUEST

>>> from plone import api
>>> portal = api.portal.get()
>>> portal.aq_chain
[<PloneSite at /Plone>, <Application at >, <ZPublisher.BaseRequest.RequestContainer object at 0x10acc66d0>]

>>> portal.unrestrictedTraverse('foo')
*** KeyError: 'foo'
```

davisagli:
Fixes #2194

### DIFF
--- a/Products/CMFPlone/URLTool.py
+++ b/Products/CMFPlone/URLTool.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from AccessControl import ClassSecurityInfo
+from Acquisition import aq_parent, aq_inner
 from App.class_init import InitializeClass
 from plone.registry.interfaces import IRegistry
 from posixpath import normpath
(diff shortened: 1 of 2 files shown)


pbauer:
Not really sure if I understand 100% of what is going on there. As far as I read it you simply move `rewrap_in_request_container` into a separate method and reuse it on `getPortalObject`. Also you get the portal with getUtility(ISiteRoot). That looks good to me.
What I do not understand is why now way more tests are failing. Some are simply because the test-setup does not register the site as a utility yet. But others are more elusive. For example `test_baselineBrokenReferencesRemoved` is weird. During `setUpPloneSite` all ATContentTypes are present and accounted for but in the test itself they are gone.

PR #2213, opened 2017-11-14
Update latest mockup

frapell:


frapell:
I did not include a changelog entry since @thet already did for a previous update

PR #2257, opened 2018-01-08
Mockup update.

thet:


mauritsvanrees:
I guess this should be merged on master too, right?

PR #2260, opened 2018-01-15
[WIP] - Correctly handle excluded from nav items in navigation portlet

cekk:
I think that this is a regression in porting from properties to registry.
In the previous version, showAllParents was correctly set from properties, and now is always set to True.

This is wrong because in this case omitted contents are always showed in the portlet, and the flag in control panel is ignored.

I've found also a [test](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/tests/testNavigationView.py#L85) for this case, but it seems disabled or ignored when i run

    bin/test -s Products.CMFPlone

Is it correct?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -76,3 +76,7 @@ Bug Fixes:
   Completly removed ``came_from`` on ``@@register`` link.
   It does not make much sense anyway and we test nowhere if there is a came_from on that link.
   [jensens]
+
+- Fix regression in SitemapNavtreeStrategy. Now contents omitted from navigation are skipped in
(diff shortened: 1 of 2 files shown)


gforcada:
@cekk that was puzzling me, but I seem to found it (I did not try it though). If you scroll all the way down to the bottom of that file you will see that there is a ``test_suite`` function, either adapting it or removing it all together should make those tests be picked up by the test runner. I would try it first by removing it completely.

jensens:
after new navigation is in place, is this still a thing?

jensens:

PR #2277, opened 2018-01-27
Prepare for Python 2 / 3 compatibility with a little help from sixer

pbauer:


Products/CMFPlone/skins/plone_scripts/getObjSize.py
```diff
@@ -35,7 +35,7 @@
 if not size:
     return '0 %s' % smaller
 
-if same_type(size, 0) or same_type(size, 0L):
+if same_type(size, 0) or same_type(size, 0):
```
mauritsvanrees: This now says the same thing twice.
And on Python 2 it no longer recognizes longs.
`isinstance six.integertypes` or whatever it is called is probably better.

pbauer: done

Products/CMFPlone/skins/plone_scripts/getObjSize.py
```diff
@@ -8,6 +8,7 @@
 ##title=
 
 from Products.CMFPlone.utils import base_hasattr
+import six
```
pbauer: Oh, crap: `Unauthorized: import of 'six' is unauthorized`

mauritsvanrees: Crap indeed. It *might* be okay to call `allow_module` on six, but that would need more investigation on what exactly is in six, and if that is safe to allow. I guess it is better to do this by hand then...

davilima6: Good news, I confirmed `isinstance` is allowed

davilima6: Problem is you have to check for long because of large file sizes but cannot know in advance if it will be available because there's no access to sys.version_info or builtins to check if RestrictedPython is running in Python 3. @loechel, is it possible?

Otoh we could try to subset `six` (RestrictedSix?) before allowing it: https://github.com/benjaminp/six/blob/master/six.py#L235-L303

mauritsvanrees: I would say keep it simple like this:
```
try:
    integer_types = (int, long)
except NameError:
    # Python 3
    integer_types = (int, )
```

gforcada: Or replace that with a browser view/utility :-) aren't we supposed to get rid of all those skins? 😆

pbauer:
My later changes regarding ZServer should be reviewed again before merging. They only change things when running py3 but a quick look cannot hurt.

pbauer:
@davilima6 can you please explain why the change of `next(self)` is needed?

davisagli:
@pbauer because Python 3 uses `__next__` instead of `next`: http://portingguide.readthedocs.io/en/latest/iterators.html#new-iteration-protocol-next

davilima6:
@pbauer, sure. In Python 3 iterators expect a `__next__` method instead of Python 2's `next`. In order to have compatibility with both, we should rename the original implementation and then alias it (so that it can be more easily removed later).

davisagli:
Sorry @davilima6 I know I'm the wrong davi but I knew the answer. :)

davilima6:
@davisagli, np at all :)

pbauer:
Thanks for the tips. http://portingguide.readthedocs.io is a great resource.
@davilima6 content with https://github.com/plone/Products.CMFPlone/pull/2277/commits/493b9acd288f85beea0eafd97d56643ce120e9f1?

pbauer:
Can someone have a look at my approach ayt making ZServer optional: https://github.com/plone/Products.CMFPlone/compare/fc3b64f3d1eb83eb68c3d6810343efb6a10fc1b4...684ab4088c455e0e6c68a639f81367f69e4e512f
@davisagli @mauritsvanrees

davilima6:
@pbauer: I am, thanks! I pay you beer Tuesday.

mauritsvanrees:
> Can someone have a look at my approach for making ZServer optional: fc3b64f...684ab40

I haven't tried, it, but your approach makes sense to me. 👍 

One thing in `setup.py`:

```
PY2_ONLY = [
    'Products.ExternalEditor',
    'ZServer',
]
```

Are those two really *required* on Python 2? It sounds like they should be optional, perhaps in a `[webdav]` or `[python2]` extra.
But I guess it makes sense for now: this commit keeps the Python 2 dependencies the same and removes these two for Python 3. Later we can look at making them optional on Python 2.

PR #2278, opened 2018-01-28
Remove backward compatibility for Products.PlacelessTranslationService

ksuess:
Last thing used from PTS was an adapter for IUserPreferredLanguages. It is replaced in zope.publisher.

Related pull requests: https://github.com/plone/plone.i18n/pull/20 and https://github.com/plone/plone.app.testing/pull/44

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,10 @@ Breaking changes:
 - Render exceptions using an exception view instead of standard_error_message.
   [davisagli]
 
+- Remove backward compatibility for Products.PlacelessTranslationService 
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
+1 for doing this in Plone 5.2.
On 5.1 it is probably better to keep the dependency, so products with `i18n` folders will still work. I am not sure if we, as community, said that this would still be supported, or if this was supposedly no longer working from Plone 5.0 onwards. But 5.1 is already in release candidate status, so I would say we keep it there.
Anyway, I will merge this one to 5.2. Danke!

PR #2279, opened 2018-01-29
Update RSS views content type

Gagaro:
itunes podcast feeds are RSS 2.0 feeds. The search rss view is RSS 1.0. The content type for these feeds should be `application/rss+xml`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,9 @@ New Features:
 
 Bug Fixes:
 
+- Update RSS content type.
+  [Gagaro]
+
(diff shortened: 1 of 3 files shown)


Gagaro:
iTunes feeds are RSS ones (https://itunespartner.apple.com/en/podcasts/faq). The content type of this view is currently `application/atom+xml`, which is wrong. Maybe the changelog entry should be placed in the breaking change category, but it is definitely in the scope of this pull request imho.

jensens:

PR #2284, opened 2018-01-31
remove pts

jensens:


mauritsvanrees:
This breaks the build. http://jenkins.plone.org/job/plone-5.2-python-2.7/lastCompletedBuild/testReport/
Can you have a look?

jensens:
I forgot Products.CMFPlone, this should relax soon (hope so).

jensens:
Oh, moment, I'll have a look

mauritsvanrees:
Thanks.

gforcada:
@jensens https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/overrides.zcml#L7 this line needs to go :-)

jensens:
Sure, it needs to point to ``plone.i18n`` now. See https://github.com/plone/Products.CMFPlone/commit/228841dbc0daffa402662ae54beb2729b4be2699

mauritsvanrees:
I think that fixes it yes.
I do get this as message locally a few times, though no error: `Could not install product plone.i18n`.
Maybe the line in `plone.app.testing` loading `plone.i18n` zcml is not needed?

mauritsvanrees:
Ah, we should pass `install: False` to that line in `plone.app.testing`, just like we do on the next line for `plonetheme.barceloneta`.
Done and pushed directly to master.

jensens:
I am not sure when in a layer zcml loading is needed. If the overrides of Products.CMFPlone is loaded there, then it should be included anyway?

mauritsvanrees:
I guess some of those lines could be removed, yes.

gforcada:
@jensens @mauritsvanrees from what I see on this comment https://github.com/plone/plone.app.testing/blob/440ee88279ced807ca3a68fed5fa45dd0fc103ed/plone/app/testing/layers.py#L148-L149 it seems that if we declare all dependencies on CMFPlone we could get rid of all this manual set up... is there a good reason for not doing that? *if that is actually the reason of course*

mauritsvanrees:
I think the only thing `plone.app.testing` should do here, is do the job of `z3c.autoinclude`. That package makes sure the zcml from lots of packages is loaded, but this does not happen during testing.
If Plone already loads `zcml` from a package, then it does not need to be handled by `plone.app.testing`.

But for some old products it may still be needed to explicitly call `z2.installProduct` like [here](https://github.com/plone/plone.app.testing/blob/440ee88279ced807ca3a68fed5fa45dd0fc103ed/plone/app/testing/layers.py#L180-L187). I think that is because packages in the `Products` namespace always get loaded on normal startup, where other package need an explicit `five:registerPackage` in their zcml.

So their should be room for cleanup and more explicitness. :-)

PR #2285, opened 2018-01-31
Update CHANGES.rst

mrsaicharan1:
Updated the bug fixes for the unreleased version.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,8 @@ New Features:
 
 Bug Fixes:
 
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
This changelog entry does not belong in CMFPlone. Every package has its own `CHANGES.rst`. So this belongs in [`plone.app.contenttypes`](https://github.com/plone/plone.app.contenttypes/blob/master/CHANGES.rst). You should update your PR https://github.com/plone/plone.app.contenttypes/pull/446 with it.

PR #2297, opened 2018-02-08
Add required plone.app.imaging as direct dependency.

thet:
Just got an error when doing a travis build of collective.lineage:
https://travis-ci.org/collective/collective.lineage/jobs/338999501

plone.app.imaging is a direct dependency of CMFPlone, so it must be declared.
In Plone 5.1 this dependency has gone.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,10 @@ New features:
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


davisagli:
plone.app.imaging is supposed to no longer be a hard dependency of CMFPlone, because it is part of the Archetypes ecosystem (and pulls in Archetypes as a dependency).

PR #2298, opened 2018-02-09
Added tests for add-ons control panel. [5.1]

mauritsvanrees:
I took over the few existing tests from `testQuickInstaller.py` and added lots more.
And I added two very minor fixes:
- Add a link to the Site Setup.
- Let `get_product_version` work when you call it with `CMFPlacefulWorkflow` too instead of `Products.CMFPlacefulWorkflow` .

If this is approved, I will make a PR for master as well.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,11 @@ New features:
 
 Bug fixes:
 
+- Added tests for add-ons control panel.
+  Add a link to the Site Setup.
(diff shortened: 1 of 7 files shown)


gforcada:
@mauritsvanrees will you push that to master (5.2) as well?

mauritsvanrees:
@gforcada Yes :-). I have now opened https://github.com/plone/Products.CMFPlone/pull/2299

PR #2299, opened 2018-02-09
Added tests for add-ons control panel. [5.2]

mauritsvanrees:
And two very minor fixes:
- Add a link to the Site Setup.
- Let `get_product_version` work when you call it with `CMFPlacefulWorkflow` too.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,11 @@ New Features:
 
 Bug Fixes:
 
+- Added tests for add-ons control panel.
+  Add a link to the Site Setup.
+  Let ``get_product_version`` work when you call it with ``CMFPlacefulWorkflow`` too.
+  [maurits]
+
 - Fix bad domain for translating password reset mails.
   [allusa]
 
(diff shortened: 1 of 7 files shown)


mauritsvanrees:
@gforcada I tried two times on Jenkins 5.2 PR, but all robot tests are failing:

```
WebDriverException:
Message: Can't load the profile. Profile Dir: /tmp/tmpVfL_Bb/webdriver-py-profilecopy
If you specified a log_file in the FirefoxBinary constructor, check it for details.
```

Did you install `geckodriver` (or `chromedriver`) on Jenkins maybe, or did something else that could cause this as side effect?  The normal 5.2 Robot test job (so not the PR job) goes just fine.

gforcada:
@mauritsvanrees yes, sorry, I did install the gecko driver and updated firefox to version 58, but as selenium is old it does not work. On the pull request to test the newer KGS for p.a.robotframework (https://github.com/plone/buildout.coredev/pull/407) it hangs at some point :-/

I will revert the firefox version bump until we can try to fix the errors.

mauritsvanrees:
No problem. I can see it is tricky test a new KGS with new Firefox and still keep the old tests running.
I reran the job and got different errors. I'll open a different issue.

mauritsvanrees:
Green! :-)

PR #2318, opened 2018-02-23
Fix test

gforcada:
Reverts https://github.com/plone/Products.CMFPlone/commit/bc138edd52daea42d321f2f1fa55b06e1c2107e6

The test states that it is ensuring that **both** URLs do work to get to the Plone's control panel, but since that commit, is actually testing twice the same URL :sweat_smile:


mauritsvanrees:
I don't know why the 'Plone Contributors Agreement verifier' check is waiting. You have obviously signed the agreement ages ago.
LGTM. I ran the tests locally for good measure, and checked in the browser to make sure that `plone_control_panel` is still rendering okay, even though it is deprecated. I will merge.

PR #2320, opened 2018-02-28
Include tinymce 4.7.6

frapell:
https://github.com/plone/mockup/pull/832


frapell:
Yeah, TinyMCE went through major refactoring on 4.6 I think...

Issue #2322, opened 2018-03-01
BUG REPORT: Some nested `section id="edit-bar"` tag in folder_contents page


terapyon:
### What I did:

- Click any `fc-breadcrumbs` on folder_contents page
- And Re-Click the `fc-breadcrumbs`

### What I expect to happen:

I think `section id="edit-bar"` tag will not be shown  in dynamic html.

### What actually happened:

`section id="edit-bar"` tag will be made too.

thet:
The issue is still open.
But I have a fix for it...

PR #2324, opened 2018-03-01
Fix nested edit bar

terapyon:
BUG REPORT: Some nested `section id="edit-bar"` tag in folder_contents page #2322 for Plone 5.1


gforcada:
@terapyon thanks for the bug fix, but your commit unfortunately was based on a quite old version of CMFPlone. Would you mind rebasing your commit? Is it meant for 5.1 or 5.2? :thinking: you might have to change the branch you based your commit on.

terapyon:
@gforcada Thank you for your comment. I want to fix on Plone 5.1.x.
Please let me know how to make a branch. Because I don't have a enough skill for github, I'm sorry.

gforcada:
Oh, now I see, the commits were fine, just that when you created the pull request, github by default selected master branch as the target branch to merge yours, while you wanted to merge it against branch 5.1.x

Fortunately, the edit button allows one to change the target branch :slightly_smiling_face:

jensens:
The tests that are broken here have no connection to the change. My PR #2342 has the exact same failure. Any idea what's going on here?

Products/CMFPlone/static/patterns/toolbar/src/toolbar.js
```diff
@@ -390,7 +390,7 @@ define([
           url: $('body').attr('data-portal-url') + path + '/@@render-toolbar'
         }).done(function(data) {
           var $el = $(utils.parseBodyTag(data));
-          that.$el.replaceWith($el);
+          that.$el.parent.replaceWith($el);
```
thet: That broke toolbar reloading when browsing folder_contents.
That also shows the importance of having tests, which we apparently do not have here.

thet: It should read ```that.$el.parent().replaceWith($el);```

thet:
@terapyon Maybe #2324 is something specific to a special integration project of yours? With my fixes in place the site is even more broken than before. see: https://github.com/plone/Products.CMFPlone/pull/2398

thet:
@terapyon no, it is not specific to your problem. I can reproduce the problem. However, this fix needs a different approach.

PR #2326, opened 2018-03-02
Fix rendering the Exception-View when main_template can't be rendered

### LINKED ISSUE #2325: Some Errors cannot render their exceptions during the creation of portal
This is a follow-up for #2203.

A exception of the following type cannot rendered in Plone 5.2:

```
ValueError

Traceback (innermost last):

    Module ZServer.ZPublisher.Publish, line 144, in publish
    Module ZPublisher.mapply, line 85, in mapply
    Module Products.PDBDebugMode.runcall, line 68, in pdb_runcall
    Module ZServer.ZPublisher.Publish, line 44, in call_object
    Module Products.CMFPlone.browser.admin, line 271, in __call__
    Module Products.CMFPlone.factory, line 146, in addPloneSite
    Module Products.GenericSetup.tool, line 388, in runAllImportStepsFromProfile
    __traceback_info__: profile-Products.CMFPlone:plone
    Module Products.GenericSetup.tool, line 1433, in _runImportStepsFromContext
    Module Products.GenericSetup.tool, line 1245, in _doRunImportStep
    __traceback_info__: rolemap
    Module Products.GenericSetup.rolemap, line 102, in importRolemap
    Module <string>, line 8, in manage_permission
    Module AccessControl.requestmethod, line 79, in _curried
    Module OFS.role, line 86, in manage_permission
    Module AccessControl.rolemanager, line 224, in manage_permission

ValueError: The permission <em>Use Database Methods</em> is invalid. 
```

The issue is that the portal and the main_template exist during rendering of ExceptionView but the main_template cannot be rendered because the context of the exception is not the Plone site but the Application. Thus the exception-template cannot be rendered properly. To fix that i suggest we try to render main_template and switch to the basic_template when that fails.

pbauer:
Fixes #2325

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,9 @@ New Features:
 
 Bug Fixes:
 
+- Fix Exception-View when main_template can't be rendered. Fixes #2325.
+  [pbauer]
+
 - Removed extra methods and tests for CMFQuickInstallerTool.
   Moved those to the Products.CMFQuickInstallerTool package.
   [maurits]
--- a/Products/CMFPlone/browser/exceptions.py
+++ b/Products/CMFPlone/browser/exceptions.py
@@ -38,9 +38,10 @@ class ExceptionView(BrowserView):
             # We cannot get the site, so we cannot render our nice template
             template = self.basic_template
         else:
-            # Use a simplified template if main_template is not available
+            # Use a simplified template if main_template is not renderable
             try:
-                self.context.unrestrictedTraverse('main_template')
+                main = self.context.unrestrictedTraverse('main_template')
+                main()
             except:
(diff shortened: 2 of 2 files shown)


mauritsvanrees:
The drawback is that this takes longer: for we render an extra template. And that is not just for this corner case, but for every exception.
And I guess the main_template could give an error even when rendering the pretty error template would have worked.

pbauer:
Ok, an alternative would be to wrapt the `return template(...)` in a try/except.

davisagli:
Oh, that's what Maurits already said. I should finish reading my email before commenting. :)

pbauer:
Will do

Issue #2245, opened 2017-12-11
Is not possible to disable TinyMCE menu items


keul:
Customer ask for removing the "Align" menu item at all (and other menus too).

I tried to empty it by providing an emtpy registry collection, but the menu entry is still displayed.

An easy fix can be to not put menu entries in case there's not items, like there:

https://github.com/plone/Products.CMFPlone/blob/6dd64f562576ecf47e067d0ed4a8564ab7e1c653/Products/CMFPlone/patterns/tinymce.py#L102

ale-rt:
This should be fixed by #2335, am I right?

PR #2334, opened 2018-03-08
Fixed  TinyMCE disability issue

mrsaicharan1:
Disable issue in TinyMCE menu items
Removed the **align** entry

With reference to #2245


mrsaicharan1:
@ale-rt Oh ok! I get it now, we render only those style items which are present in the enabled ones! Thank you for the update on the code.

mrsaicharan1:
I've made a similar change in the code @ale-rt . Could you check it out once? Thanks!

Issue #2359, opened 2018-03-27
Upgrade steps  not sufficient for last changes of tinymce integration ..... ?


fgrcon:
cc: @frapell , @thet  

# Steps to reproduce:

 * clean 5.0.x  build
 * create plone site

 * clean 5.1.x build   (including https://github.com/plone/Products.CMFPlone/pull/2320 and later)
 * move dat.fs  from 5.0  to 5.1.x
 * run Upgrade

# Results:
![edit-bug](https://user-images.githubusercontent.com/1640384/37965028-f01f624a-31c3-11e8-82aa-1d045dd046a0.jpg)

## error messages when building bundle:
``
CSS Bundle bauen
1 CSS Dateien zu bauen
Fehler beim Kompilieren der LESS-Datei http://192.168.1.99:8083/Plone_5_0_X/++plone++static/plone-logged-in.less: Error: 'http://192.168.1.99:8083/Plone_5_0_X/++plone++static/components//tinymce-builded/js/tinymce/skins/lightgray/skin.less' wasn't found (404)
Fehler beim Kompilieren der LESS-Datei http://192.168.1.99:8083/Plone_5_0_X/++plone++static/plone-logged-in.less: Error: 'http://192.168.1.99:8083/Plone_5_0_X/++plone++static/components//tinymce-builded/js/tinymce/skins/lightgray/skin.less' wasn't found (404)
Fehler beim Kompilieren der LESS-Datei http://192.168.1.99:8083/Plone_5_0_X/++plone++static/plone-logged-in.less: Error: 'http://192.168.1.99:8083/Plone_5_0_X/++plone++static/components//tinymce-builded/js/tinymce/skins/lightgray/skin.less' wasn't found (404)
Fehler beim Kompilieren der LESS-Datei http://192.168.1.99:8083/Plone_5_0_X/++plone++static/plone-logged-in.less: Error: 'http://192.168.1.99:8083/Plone_5_0_X/++plone++static/components//tinymce-builded/js/tinymce/skins/lightgray/skin.less' wasn't found (404)
Fehler beim Kompilieren der LESS-Datei http://192.168.1.99:8083/Plone_5_0_X/++plone++static/plone-logged-in.less: Error: 'http://192.168.1.99:8083/Plone_5_0_X/++plone++static/components//tinymce-builded/js/tinymce/skins/lightgray/skin.less' wasn't found (404)
[...]

PR #2361, opened 2018-03-28
Include new paths for TinyMCE css

frapell:
gh-2359


jensens:
Excluded both, `pytz` and `date-utils` from this PR. New PR for those are in preparation.

PR #2380, opened 2018-04-11
Move the gtbn_funcglobals patch security test to functional tests

Rotonen:
Seems the patches having already been run in for the unit test layer actually depends upon a test layer leak. Moving to functional tests to make this test unflaky.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,9 @@ New features:
 
 Bug fixes:
 
+- Unflakied a unit test.
+  [Rotonen]
+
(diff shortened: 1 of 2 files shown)


Rotonen:
I think I accidentally somehow tripped up the CL verifier bot with a wrong batch base after a confusing cherry-pick merge force push batch job. Dunno which party could not keep up there, me or Github.

Rotonen:
As I think I cannot fix that, I'll just re-PR. Sorry for the noise.

PR #2401, opened 2018-05-03
Optional auto-login after passwort (re-)set.

jensens:


Products/CMFPlone/browser/password_reset.py
```diff
@@ -75,6 +84,40 @@ class PasswordResetView(BrowserView):
...
+                )
+            IStatusMessage(self.request).addStatusMessage(
+                _(
+                    'password_reset_successful',
+                    default='Password reset successful, '
+                            'you are logged in now!',
+                ),
+                'info',
+            )
+            url = INavigationRoot(self.context).absolute_url()
```
davisagli: I think we should use the login_next script so that the location is determined in the same way as for a normal login?

jensens: This is designed to work with plone.login, so no such scripts...

jensens:
The whole here is wip. The redirection target could be configureable, also if the login should happen at all. Ideas are welcome!

Products/CMFPlone/browser/password_reset.py
agitator: https://github.com/plone/Products.CMFPlone/issues/2439

Issue #2316, opened 2018-02-21
Creating a new 5.1-pending site with policy package -> aborting transaction due to no CSRF protection


zopyx:
Creating a new Plone 5.1-pending site with a simple policy package and these dependencies (the dependencies are irrelevant since I can reproduce the error with plone.app.dexterity as single dependency too)

```
<?xml version='1.0' encoding='UTF-8'?>
<metadata>
  <version>1000</version>
  <dependencies>
      <dependency>profile-plone.app.dexterity:default</dependency>
      <dependency>profile-plone.restapi:default</dependency>
      <dependency>profile-Products.PloneFormGen:default</dependency>
      <dependency>profile-plone.app.multilingual:default</dependency>
      <!--
      <dependency>profile-dynamore.seminarsdx:default</dependency>
      -->
  </dependencies>
</metadata>
```

causes this error

```

2018-02-21 08:13:28 INFO ZServer HTTP server started at Wed Feb 21 08:13:28 2018
        Hostname: 0.0.0.0
        Port: 5080
2018-02-21 08:13:31 INFO Products.PloneFormGen gpg_subprocess initialized, using /usr/bin/gpg
2018-02-21 08:13:32 INFO ZEO.ClientStorage zeostorage ClientStorage (pid=27729) created RW/normal for storage: '1'
2018-02-21 08:13:32 INFO ZEO.cache created temporary cache file '<fdopen>'
2018-02-21 08:13:32 INFO ZODB.blob (27729) Blob directory `/home/plone/sandboxes/dynamore.buildout.plone5/var/blobstorage` is unused and has no layout marker set. Selected `bushy` layout. 
2018-02-21 08:13:32 INFO ZEO.asyncio.base Connected Protocol(('127.0.0.1', 5200), '1', False)
2018-02-21 08:13:32 INFO ZEO.ClientStorage zeostorage Connected to storage: ('localhost', 5200)
2018-02-21 08:13:41 INFO PloneFormGen Patching plone.app.portlets ColumnPortletManagerRenderer to not catch Retry exceptions
2018-02-21 08:13:41 INFO Zope Ready to handle requests
2018-02-21 08:13:49 ERROR Zope.SiteErrorLog 1519197229.260.35271801443 http://dynamore.zopyx.de:5080/manage_delObjects
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
[...]

PR #2410, opened 2018-05-14
Add Plone Site: do a transaction commit before profiles

mauritsvanrees:
When adding a Plone Site, do a transaction commit before applying extra profiles. This avoids errors in some cases when selecting add-ons in the advanced Plone Site add form, where installing them right after site creation in the add-ons control panel goes fine.

It should certainly fix some CSRF autoprotection errors raised by the first view after creating a Plone Site.

Fixes https://github.com/plone/Products.CMFPlone/issues/2316.
And I think there was a similar issue raised recently, but I cannot find it.

Note for tests: since this does a transaction commit, any test code that calls `addPloneSite` should use a functional test layer, to avoid leakage into other tests. Case in point is an initial test failure in CMFPlone itself:

```
Error in test testPlonesiteWithoutUnicodeTitle
(Products.CMFPlone.tests.test_factory.TestFactoryPloneSite)
  File ".../unittest/case.py", line 329, in run
    testMethod()
  File ".../Products/CMFPlone/tests/test_factory.py", line 29,
 in testPlonesiteWithoutUnicodeTitle
    self.app, 'ploneFoo', title=TITLE, setup_content=False)
  File ".../Products/CMFPlone/factory.py", line 132, in addPloneSite
    context._setObject(site_id, PloneSite(site_id))
  File ".../OFS/ObjectManager.py", line 324, in _setObject
    v = self._checkId(id)
  File ".../OFS/ObjectManager.py", line 127, in checkValidId
    'The id "%s" is invalid - it is already in use.' % id)
BadRequest: The id "ploneFoo" is invalid - it is already in use.
```

I fixed that in the PR by adding a condition: when no extra profiles are selected, don't do a commit.
Only two test files in CMFPlone call `addPloneSite` directly, plus of course `plone.app.testing` in test layer setup, so this should not normally cause problems.

This PR is for master (5.2). I can port it to other versions once approved. I would say: only 5.1.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,12 @@ New Features:
 
 Bug Fixes:
 
+- When adding a Plone Site, do a transaction commit before applying extra profiles.
+  This avoids errors in some cases when selecting add-ons in the advanced Plone Site add form,
+  where installing them right after site creation in the add-ons control panel goes fine.
+  Fixes `issue 2316 <https://github.com/plone/Products.CMFPlone/issues/2316>`_.
+  [maurits]
+
 - Unflakied a unit test.
   [Rotonen]
 
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -9,6 +9,9 @@ from zope.event import notify
 from zope.interface import implementer
 from zope.site.hooks import setSite
 
+import transaction
+
+
 _TOOL_ID = 'portal_setup'
 _DEFAULT_PROFILE = 'Products.CMFPlone:plone'
 _CONTENT_PROFILE = 'plone.app.contenttypes:plone-content'
@@ -162,6 +165,14 @@ def addPloneSite(context, site_id, title='Plone site', description='',
     # Do this before applying extension profiles, so the settings from a
     # properties.xml file are applied and not overwritten by this
     site.manage_changeProperties(**props)
+
+    # In some cases we can get errors when add-ons are installed immediately.
+    # Installing the add-ons separately within the Add-ons control panel
+    # goes fine.
+    # One example is https://github.com/plone/Products.CMFPlone/issues/2316
+    # So it seems helpful to do a transaction commit here.
+    if extension_ids:
+        transaction.commit()
 
     for extension_id in extension_ids:
         setup_tool.runAllImportStepsFromProfile(


jensens:
LGTM

jensens:
+1 for backporting this to Plone 5.1

zopyx:
Is it appropriate to commit here? addPloneSite() is supposed to be transactional. An error at some point of the setup procedure should not leave any traces...so what happens if an error occurs after the commit?

mauritsvanrees:
Then you have a Plone Site without add-ons. That is probably acceptable here.

The only thing that happens after applying the extra profiles, is a redirect to the Plone Site url:
https://github.com/plone/Products.CMFPlone/blob/5.1.2/Products/CMFPlone/browser/admin.py#L265-L276

And... actually it should return an empty string after setting the redirect. Currently it also calls `self.index()`, and the redirect only really happens after the rendering is done. Rendering the index/template is unneeded, and might be a source of CSRF errors. Is that something you could try instead in a project where you see this problem?

zopyx:
I don't think that it is acceptable...everything can happen during during the extensions profile after the commit. This will of course produce a naked Plone site but it would not produce what you expect. There is an unwritten contract that you should not commit yourself in Zope and Plone because this is handled by the ZPublisher. In that sense: `addPloneSite()` is no longer transactional.  Also consider using `addPloneSite()` as part of a provisioning API (e.g. in plone.api). A remote API client can  expect that a site is created according to its specs or not...but nothing in between...so I assume that committing here is a workaround at best...I would not consider this as a suitable solution.

mauritsvanrees:
Ah, I found the other issue that I was looking for: https://github.com/plone/Products.CMFPlone/issues/2228
Not exactly the same, but also a problem that only happens during initial site creation and not when using the add-ons control panel. Or at least that is the case for @ida. Maybe not for @zopyx who added that issue.
That issue is where I originally suggested a transaction commit, which apparently helped for Ida.

mauritsvanrees:
I had another look and I see that the transaction commit [doens't really help](https://github.com/plone/Products.CMFPlone/issues/2228#issuecomment-388801748). It only avoids printing several tracebacks, so that you are more likely to notice the original traceback that was always visible if you cared to scroll up far enough.

I will revert this PR.

mauritsvanrees:
Directly reverted on master in 4e677244018784160a87d9e5c22baa7ffcf23b8b.

PR #2423, opened 2018-05-17
Adapt tests to `Products.GenericSetup >= 2.0`.

icemac:
This is needed to get Plone 5.2 coredev buildout green again. After the Python 3 changes of `Products.GenericSetup` have been merged to master.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ Changelog
 
 Breaking changes:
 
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
Yihaa! Go ahead with the merges I would say. And probably for all those packages make sure that Plone 5.1 and earlier use a different branch than master.

Issue #2422, opened 2018-05-17
Move GopipIndex from plone.app.folder to Products.CMFPlone


pbauer:
Needed to remove Archetypes. Subticket for #2390

davisagli:
This is low priority. plone.app.folder's dependency on Archetypes is already optional.

petschki:
fixed in #2424

PR #2424, opened 2018-05-18
move test and init of gopipindex to Products.CMFPlone

jmevissen:
https://github.com/plone/Products.CMFPlone/issues/2422


mauritsvanrees:
I haven't checked, but I suspect that we need to provide migration. Existing folders of class `plone.app.folder.folder.ATFolder` will probably end up as `Broken` because the class is gone.
Same for instances of the `GopipIndex`.
Maybe this is tracked somewhere else, or already taken care of somehow.

petschki:
@mauritsvanrees you're right ... I think this isn't covered anywhere else but I'll make an upgrade step in plone.app.upgrade for this tomorrow.

PR #2436, opened 2018-05-30
Mockup update

frapell:
Once https://github.com/plone/mockup/pull/850 gets merged in, this brings the changes to Plone


frapell:
There is no changelog entry because there is one already for "Mockup update"

petschki:
I've merged current master and recompiled resources. Jenkins is now green.
as @jensens mentioned here https://github.com/plone/mockup/issues/856 I'd also like to backport this to `5.1` branch.
If everybody agrees, I'll create a pull request with backporting this changes and update the existing pull request for p.a.upgrade https://github.com/plone/plone.app.upgrade/pull/163 to add the resources in a `to_513` profile

sunew:
@frapell as far as I can see, these changes are already included in 5.1.x and in master?

master
9ba4a84aa2 (Franco Pellegrini   2018-07-10 23:47:13 -0300  867)   <records prefix="plone.resources/mockup-patterns-datatables"

5.1.x
0a479b9dec (Franco Pellegrini   2018-07-10 23:17:42 -0300  867)   <records prefix="plone.resources/mockup-patterns-datatables"

So we can close this one?

sunew:
only difference is that this PR is using the minified resource. I assume thats an error?

   <records prefix="plone.resources/datatables.net-autofill-bs"
             interface='Products.CMFPlone.interfaces.IResourceRegistry'>
-      <value key="js">++plone++static/components/datatables.net-autofill-bs/js/autoFill.bootstrap.min.js</value>
+      <value key="js">++plone++static/components/datatables.net-autofill-bs/js/autoFill.bootstrap.js</value>
   </records>

frapell:
@sunew It has been so long that I don't remember what happened with this to be honest... As per your question, yes, I guess we should be pulling the non minified version right?

sunew:
@frapell I'm not sure what is the best approach :) But we will bundle the resources when building the plone bundles, so I guess non-minified is the way to go.

I close this one.

PR #2479, opened 2018-07-31
Add X-Content-Type-Options and X-XSS-Protection security headers

hvelarde:
refs. https://infosec.mozilla.org/guidelines/web_security

### DIFF
--- a/Products/CMFPlone/skins/plone_templates/main_template.pt
+++ b/Products/CMFPlone/skins/plone_templates/main_template.pt
@@ -20,7 +20,9 @@
                   site_properties context/portal_properties/site_properties;
                   ajax_load request/ajax_load | nothing;
                   ajax_include_head request/ajax_include_head | nothing;
-                  dummy python:request.RESPONSE.setHeader('X-UA-Compatible', 'IE=edge');"
+                  dummy python:request.RESPONSE.setHeader('X-UA-Compatible', 'IE=edge');
+                  dummy python:request.RESPONSE.setHeader('X-Content-Type-Options', 'nosniff');
+                  dummy python:request.RESPONSE.setHeader('X-XSS-Protection', '1; mode=block');"
       tal:attributes="lang lang;
                       xml:lang lang">
 
--- a/Products/CMFPlone/skins/plone_templates/main_template.pt
+++ b/Products/CMFPlone/skins/plone_templates/main_template.pt
@@ -20,6 +20,7 @@
                   site_properties context/portal_properties/site_properties;
                   ajax_load request/ajax_load | nothing;
                   ajax_include_head request/ajax_include_head | nothing;
+                  dummy python:request.RESPONSE.setHeader('X-UA-Compatible', 'IE=edge');
                   dummy python:request.RESPONSE.setHeader('X-Content-Type-Options', 'nosniff');
                   dummy python:request.RESPONSE.setHeader('X-XSS-Protection', '1; mode=block');"
       tal:attributes="lang lang;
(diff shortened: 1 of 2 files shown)


hvelarde:
@esteele @plone/framework-team this is ready!

ale-rt:
Thanks @hvelarde. I would merge this one, but maybe the "mode=block" is too strict. I would also ask the @plone/security-team to review this one.

jensens:
- X-Content-Type-Options [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options) - yes, we have to set it.
- X-XSS-Protection [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection) - as far as I know this is relevant for old browsers only. But, well, yes we can support those.
- X-UA-Compatible header: this is only effective in IE8, 9 and 10. It was disabled in 11. I am not sure how much value it add these days. OTHO it does not hurt, because its ignored in all modern browser.

So, w/o checking Plone 5 - are those set there similarly already?

mauritsvanrees:
Word choice: you are not '_adding support_' for these headers. You are **actually adding** the headers. 'Adding support' suggests to me that this is optional and that the user can choose to set these headers: maybe the webmaster can set this in a control panel, or an editor can set this when editing a page. That is not the case. It is hardcoded.

You can already set these headers in a web server, like in the doc you point to. So I wonder if that then causes problems. Are headers then set twice?  Does the one in the web server win (I hope so)?  The one in Plone?

I wouldn't touch the `X-UA-Compatible` header in Plone 4.3.  Too high chance that some theme relies on this header, and that every page on the site breaks when viewed with IE if you remove this.

hvelarde:
@ thanks! I'll fix this according to your comments.

hvelarde:
@mauritsvanrees done! answering to your question, any site administrator would be able to change any of the response headers as I did in the example above; what I want is to accomplish here is basic protection with those headers for people that don't know what they are doing.

those headers are in Mozilla's [Web Security Cheat Sheet](https://infosec.mozilla.org/guidelines/web_security) a document I just discovered and seems a good starting point to enhance our solution.

mauritsvanrees:
There is a [report for 4.3](https://github.com/plone/Products.CMFPlone/issues/2964) where this change causes breakage because the template is not correct.
I cannot reproduce it, but maybe someone else has an idea.

mauritsvanrees:
The breakage is indeed caused by this PR. See [my comment](https://github.com/plone/Products.CMFPlone/issues/2964#issuecomment-548132465). Same for the similar change in plonetheme.classic, linked there.

So again a minor new feature breaks Plone 4. (`base` tag anyone?)
That's it. No more new features for Plone 4, however small they are. I will veto it. I mean it.
cc @plone/release-team 

I am torn between merging PR #2965 that fixes it in CMFPlone, and doing a similar PR in plonetheme.classic.
Or reverting the feature. I am inclined to do this. No new features, so if a recently added feature is broken, we may as well rip it out.

You know what is the most idiotic part? Without wanting to lay blame anywhere, because this is a community effort so we are all responsible:

1. Someone makes PRs for two themes, that break those themes, and does not realise it because the default theme is not affected. 
2. Someone merges those PRs.
3. No one else noticed, including me, even though there was about a year between merging the CMFPlone PR and the next release.
4. There are no tests in core that catch this.

I would have hoped that at least *one* of those four had not happened. Doesn't matter which one.

Now excuse me while I go drink some relaxing chamomile tea and head to bed. :-)

PR #2484, opened 2018-07-31
Get rid of obsolete X-UA-Compatible header

hvelarde:
refs. https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/cc288325(v=vs.85)


hvelarde:
@plone/framework-team build seems to be broken for a completely unrelated reason.

PR #2489, opened 2018-08-03
Scenarios added and fixed timing problems in querystring robot tests.

llisa123:
- new Scenarios added
- replaces Sleep by marking
- fixed existing Scenario (Location query)

Problems left (not critical):
- plone.app.robotframework.content creation mechanism does not set the timezone for date fields of plone.app.event, therefore some of the date related test are disabled.


llisa123:
see also https://github.com/plone/Products.CMFPlone/issues/1656

jensens:
Still fails with Firefox with 4 tests, no problems in Chromium. In Firefox a ``Sleep  2`` may help, but that's no solution. We need to fix our setup to have either stable Firefox or run tests on Chromium. 

Running on Chromium (the OpenSource Chrome) is really simple by installing from official Ubuntu 18.04 packages with ``apt install chromium-chromedriver chromium-browser``

@plone/testing-team Why don't we switch to Chromium as default test browser on Jenkins? IMO this would save us a lot of headache.

jensens:
See also https://github.com/plone/jenkins.plone.org/issues/227

PR #2496, opened 2018-08-08

### LINKED ISSUE #2052: Viewing robots.txt on Zope root gives AttributeError
When you visit http://localhost:8080/robots.txt you get a short error from Zope:

```
Error Type: AttributeError
Error Value: 'NoneType' object has no attribute 'absolute_url'
```

The `robots.txt` view should either not be defined on the Zope root, or the code fixed so that it at least gives no error.

This is Plone 5.0.7.

hvelarde:
closes #2052

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,9 @@ Breaking changes:
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


hvelarde:
@plone/testing-team I tried to run the jobs but seems Jenkins is not working.

PR #2497, opened 2018-08-08

hvelarde:
refs. #2052

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,9 @@ New features:
 
 Bug fixes:
 
+- Fix registration of ``robots.txt`` browser view to avoid ``AttributeError`` on Zope's root (fixes `#2052 <https://github.com/plone/Products.CMFPlone/issues/2052>`_).
+  [hvelarde]
+
 - Fix plone.app.redirector support for JSON/unspecified requests.
   [rpatterson]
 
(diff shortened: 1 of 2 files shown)


ale-rt:
I liked having this one registered for every context :)
I would prefer to fix the code logic but I am not against merging this one and the related ones (#2497, #2498)

hvelarde:
@ale-rt why did you like it? how can this be useful in other contexts? can you argument a little bit more?

ale-rt:
Theoretically you could publish a subfolder of your Plone site with something like: `/VirtualHostBase/$scheme/$host:$server_port/Plone/client/VirtualHostRoot/$1`. With this patch the robot.txt view will not be available anymore.
But as said this is something special, outside the normal Plone user experience and can be easily fixed by adding another rewrite or customizing the robots.txt view.
For that reason my prefered choice was to fix the attribute error, but I am totally ok with this solution!

hvelarde:
@ale-rt sincerely I don't see too much value on that as it's a corner case; anyway I think the browser view could be fixed with the following patch:

```diff
-        portal_url = portal_state.portal_url()
+        try:
+            portal_url = portal_state.portal_url()
+        except AttributeError:  # Zope root?
+            portal_url = '/'
```

I would prefer to just register the view where it belongs: the navigation roots.

ale-rt:
Ok, no problem @hvelarde, I have started the tests (I had to submit the form a couple of time).
If they pass feel free to merge this one yourself!

PR #2498, opened 2018-08-08

hvelarde:
refs. #2052


ale-rt:
I am merging this even if "Plone Jenkins CI - pull-request-5.0" is failing, I think they should not even be reuquerired CC @plone/testing-team

PR #2504, opened 2018-08-30
Inhibit install of broken add-ons, show warning.

ida:


ida:
Withdrawing this PR, because it introduces the method `getBrokenDependencyProfileIds`, which becomes obsolete once the next release of Products.GenericSetup happens, which contains an identical method `getBrokenDependencies` with test-coverage. Better re-use that one.

PR #2511, opened 2018-09-09
Fix changelog after 5.1.x items sneaked in, in commit 5384b810d, and …

sunew:
…even more in 7cefb8f

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,194 +10,9 @@ Changelog
 
 Breaking changes:
 
-- Remove all dependencies on plone.app.controlpanel. 
+- Remove all dependencies on plone.app.controlpanel.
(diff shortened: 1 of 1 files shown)


sunew:
Note that all 5.1.x items are in HISTORY.txt

hvelarde:
this is why I dislike the usage of `CHANGES.rst merge=union` in our repositories: people don't check and we end up with a mess in the changelog.

mauritsvanrees:
This is why I want to use towncrier. See https://community.plone.org/t/using-towncrier-for-changes-rst-in-plone-packages/6486/7
I started preparing stuff a couple of months ago, but didn't have time to really finish it, or keep track of if someone else picked up that ball.

PR #2514, opened 2018-09-14
Update TinyMCE to 4.7.13 in master

erral:


sunew:
@erral the changelog has been garbled again - a lot of old stuff has sneaked in.

sunew:
@erral (I just merged a cleanup not long ago - see https://github.com/plone/Products.CMFPlone/commit/d07d596bb78df07271ddfdb18636a8181afbef88 )

sunew:
@erral we also need an update of the bundle dates in pa.upgrade, but I'm merging in https://github.com/plone/mockup/pull/863 so I guess it will be covered by that update. (I'll do a PR for pa.upgrade as well)

erral:
I see, I rebased on top of master when saw your changes, but it looks like CHANGES's ones where not applied correctly. I have just fixed them.

Issue #2526, opened 2018-09-27
DX-Documents have no size


pbauer:
Dexterity-Based Documents with a richtext-field as primary field do not have a size. 
This is a regression compared to Archetypes. 

The failing test in CMFPlone (py3 branch) is:

```python
def test_getObjSize(self):
    from Products.CMFPlone.CatalogTool import getObjSize
    get_size = getObjSize.callable
    # FIXME: getObjSize die not count text in DX
    self.doc.text = RichTextValue('a' * 1000)
    self.doc.reindexObject()
    self.assertEqual(get_size(self.doc), '1 KB')
```

plone.app.contenttypes only has getObjSize indexers for files and images.

reinhardt:
The test passes fine for me. The ```getObjSize``` indexer calls [DAVResourceMixin.get_size()](https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/filerepresentation.py#L54) which returns 1000 which is in turn correctly converted to "1 KB".

Am I missing something?

reinhardt:
I'll have a look at the comment in https://github.com/plone/plone.dexterity/pull/89#issuecomment-425803602.

davisagli:
Good catch and thanks for the fix, @reinhardt

PR #2538, opened 2018-10-01
Fix DX document size for Python 3

reinhardt:
Refs #2526

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,9 @@ New features:
 
 Bug fixes:
 
+- Fixed getObjSize indexer for Python 3. #2526
+  [reinhardt]
+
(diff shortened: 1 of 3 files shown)


ale-rt:
Sweet! Can this fix also be applied here: https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/skins/plone_scripts/getObjSize.py#L48?

reinhardt:
Pretty sure it can, but that's a plone_scripts script that's supposed to go away anyway...

reinhardt:
Actually, the python3 branch already has a fix that should be good enough until the script is removed.

https://github.com/plone/Products.CMFPlone/blob/python3/Products/CMFPlone/skins/plone_scripts/getObjSize.py#L49

PR #2544, opened 2018-10-01
Factored out format_size method for replacing getObjSize.py

reinhardt:
The only difference I can see between the skin script getObjSize.py and the indexer method getObjSize is that the former takes a ```size``` parameter as an alternative to getting the size from an object. I've factored out a format_size() method that can easily be used instead of the skin script in places where it is used that way (i.e. with the ```size``` parameter).

See #1801

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,9 @@ Breaking changes:
 
 New features:
 
+- Factored out format_size method for replacing getObjSize.py. #1801
+  [reinhardt]
+
 - Update TinyMCE to 4.7.13
   [erral]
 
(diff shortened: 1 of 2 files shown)


hvelarde:
there's another function that does the same here: https://github.com/plone/plone.app.contenttypes/commit/9173ca5126b4c140bac5ed4b3faf9e321fe6be19

reinhardt:
Ah! Thanks, I'll try to consolidate.

reinhardt:
I'll move the method to Products.CMFPlone.utils and import from there both in CatalogTool and plone.app.contenttypes.browser.file. Let me know about any objections.

reinhardt:
```human_readable_size``` also uses B (Bytes) for output, which I don't think is necessary for human readable sizes. I'll take the other implementation which only has KB, MB and GB. It also handles more special cases.

Issue #2315, opened 2018-02-20
Toolbar in 5.1RC1: clipping of text 


zopyx:
See photos (german UI, Chrome 64 Linux). 

Texts would be "Bearbeiten" and "Hinzufügen"...so at least one character missing.

![img_1301](https://user-images.githubusercontent.com/594239/36419923-9cdc3a32-1633-11e8-9648-a8a045ddebac.jpg)
![img_1302](https://user-images.githubusercontent.com/594239/36419925-9d0a08fe-1633-11e8-95d0-a31dc7e6a57d.jpg)

iamar7:
@zopyx I want to work on this issue. can you tell me where i can find the required code to proceed.

iamar7:
You can correct this by following this-:
   Toolbar -> Admin -> Site setup -> Go to control panel
   In the control panel -> Go to Resource Registry
   Enable Development mode
   In the "plone" bundle below click on develop CSS
   Go to Less variables tab
   Find the variable plone-left-toolbar-expanded and set it to 300 px or according to your use
   Hit the save button and reload the page 
   go back to Registry tab and hit the Build button for the plone module
   Now check your issue should have been solved.

![corrected](https://user-images.githubusercontent.com/20831630/37863908-4736e7dc-2f5e-11e8-9bd4-300b09b8ae8e.png)

tmassman:
Same as #723 and fixed with hyphenation support added.

PR #2545, opened 2018-10-01
Fix toolbar word break

tmassman:
Fix wrong CSS and recompile plone resources.


thet:
Refs:
- https://github.com/plone/Products.CMFPlone/pull/2202
- https://github.com/plone/Products.CMFPlone/issues/2315
- https://github.com/plone/Products.CMFPlone/issues/2315

@zopyx @jensens this is done due to poor browser support for the ``hyphen`` property - see: https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#Browser_compatibility
It also didn't work properly in the browsers we tested - right @tmassman ?

If there is any disagreement - let's discuss it.

tmassman:
Yes, `word-wrap` is the only reliable property we can use here. Thx for the merge!

zopyx:
Hyphenation is supported on latest Firefox and Chrome versions on Mac and Windows (Linux support only fire FF), supported by Safari, supported by iOS. And hyphens: auto does not hurt if it is not supported.

thet:
@zopyx done with:
https://github.com/plone/Products.CMFPlone/commit/3700a62852d75632545ba6c0d1d222df6a86f5f1
updated in:
https://github.com/plone/Products.CMFPlone/pull/2546

bundle compilation on both 5.1 and 5.2 will be done soon.

PR #2546, opened 2018-10-01
Fix toolbar word break

tmassman:
Fix wrong CSS selector which prevents correct word break of too long toolbar titles.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,7 +61,10 @@ New features:
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


thet:
The toolbar is being copied to mockup in 5.2.

tmassman:
Since it will be moved to `mockup` we can close this here. Maybe we can apply the changes from #2545 in mockup to fix the word break issue there for 5.2.x.

PR #2558, opened 2018-10-03
re #1801: Removed enableHTTPCompression.py skin script.

tlotze:


reinhardt:
The test failures I saw are known to be caused by isolation issues. Merged anyway.

PR #2564, opened 2018-10-04
Ported generateUniqueId skin script to utils.generate_unique_id() method

reinhardt:
Skin script is not removed yet; will be a separate PR when all usages have been changed.
See #1801


ale-rt:
should we also remove the script in this PR?

reinhardt:
If we first add the replacement method, then change the usages, then remove the script, we can have the tests green all the time. :-) Any arguments for removing the script now?

reinhardt:
We can actually also keep the tests green by testing all the related pull requests together. I'll go for that.

mauritsvanrees:
Both approaches make sense.
When there are too many usages that need to be fixed, and they can't be done today, merging the new utility method may work better.

A few tests that call the new function would be nice though. A simple unittest, without any Plone layer setup.

reinhardt:
Thanks for the input!

> A few tests that call the new function would be nice though.

That on the one hand makes total sense. On the other, the main intention of the method is to generate an ID that is unlikely to be the same as any other ID that's currently in the system. That's pretty tough to test I think, especially if the test is supposed to be reproducible. But I'll think about it.

reinhardt:
Is the format important, though?

reinhardt:
I'd rather test that a couple of consecutive calls don't yield the same ID more than once. That probably doesn't catch any real life issues but it's at least true to the idea of the method...

davisagli:
That sounds good too. Maybe the format isn't important, but it would make sure we notice if something changes.

reinhardt:
Ah, good point. If e.g. the random part is omitted then we'll catch that by checking the format. I'll add something like that as well, thanks!

reinhardt:
How about this? 8de9d19

reinhardt:
I'll merge - let me know in case you have more suggestions for the tests.

mauritsvanrees:
> Will Archetypes need to support 5.2?

Yes, Archetypes will still be there in Plone 5.2 and needs to be supported.
But that is only on Python 2. It will not be ported to Python 3.
But that is not the issue here.

> I have these other PRs but I'm not sure whether they make sense:

Yes, they make sense.
One problem though: for these two packages, the master branch is also used on Plone 5.1. So if we/you merge the PRs, Jenkins 5.1 will be broken.

The nicest thing to do, would be to backport (copy) the new `utils.generate_unique_id` function to CMFPlone 5.1. Probably the tests too. But keep the skin script there.
Then any package that used the script until now, can move to the new way already in Plone 5.1, making it easier to have an add-on that works on both 5.1 and 5.2.

reinhardt:
Great, thanks. Then I'll do a little backporting.

jensens:
Sorry, I am late at the party. But, isn't it simpler to just move the script to Archetypes skins folder and let it die there together with Archetypes? In fact it has no use outside Archetypes. Thus we just need to delete it im CMFPlone and add it there, all done. Or do I miss something?

reinhardt:
> But, isn't it simpler to just move the script to Archetypes skins folder and let it die there together with Archetypes?

I think that makes sense. I've done it like that for other scripts. I may have missed it here because I didn't realise that LinguaPlone also counts because it depends on Archetypes.

jensens:
close in favour of #2600

PR #2587, opened 2018-10-09
Make test more robust

tschorr:
s. https://github.com/plone/plone.protect/pull/85#issuecomment-427833736 and following for why this is.


tschorr:
Jenkins is already running, s. https://jenkins.plone.org/job/pull-request-5.2/1550

mauritsvanrees:
Yes, I see. Strangely, the check from mr.roboto says "Pending — Please run the job, click here". Oh, well, it is running now.

tschorr:
I think it's because I've updated the changelog after starting Jenkins.... I keep forgetting Jenkins runs take so long.

PR #2603, opened 2018-10-28
removed Products.ResourceRegistries

ksuess:
see pull request plone.app.upgrade https://github.com/plone/plone.app.upgrade/pull/183/files


gforcada:
:100: thanks!! I removed the reference on buildout.coredev on tests.cfg, so that it is no longer on the installed eggs.

@plone/release-team @plone/framework-team *(open question)* Should we also remove it from sources.cfg and versions.cfg? :thinking: I would be :+1: on sources.cfg but on +0 on versions.cfg :confused:

mauritsvanrees:
👍 for removing from `sources.cfg` (otherwise mr.roboto would needlessly want to do PRs for it) and for keeping it in `versions.cfg`.

gforcada:
I see that @jensens went ahead and removed it from sources.cfg :+1:

PR #2611, opened 2018-11-01
Add bin/instance verifydb command to validate pickles

davisagli:
This adds a `bin/instance verifydb` command which can be used to check a database after migrating to Python 3 to make sure all pickles can still be unpickled.

I'm submitting this to Plone rather than Zope because Plone sites contain some pickles that can't be loaded unless Plone's codebase is present (for example, the plone.dexterity.schema.generated module is magical, and there are some module aliases set up when plone.app.upgrade is imported).

### DIFF
--- a/Products/CMFPlone/_scripts/verifydb.py
+++ b/Products/CMFPlone/_scripts/verifydb.py
@@ -38,17 +38,17 @@ def verify_zodb(obj, debug=False):
 
     logger.info('Scanning ZODB...')
 
-    next = None
+    next_ = None
     count = 0
     errors = 0
     while True:
         count += 1
-        oid, tid, data, next = storage.record_iternext(next)
+        oid, tid, data, next_ = storage.record_iternext(next_)
         logger.debug('Verifying {}'.format(oid))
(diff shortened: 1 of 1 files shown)


Products/CMFPlone/_scripts/verifydb.py
```diff
@@ -0,0 +1,88 @@
...
+def verify_zodb(obj, debug=False):
+    storage = obj._p_jar._db._storage
+    if not IStorageCurrentRecordIteration.providedBy(storage):
+        raise TypeError(
+            'ZODB storage {} does not implement record_iternext'.format(
+                storage))
+
+    logger.info('Scanning ZODB...')
+
+    next = None
```
jensens: `next` is a builtin, please use a different variable name.

davisagli: Ah yes, thanks. I added https://github.com/zopefoundation/ZODB/issues/233 to fix the example I based this on.

davisagli:
@jensens updated

Issue #2596, opened 2018-10-20
backport safe_nativestring to Plone 5.0 and 5.1


pbauer:
`safe_nativestring` is useful when writing py2 and py3 compatible code. It could be used in may places, e.g. https://github.com/plone/Products.PortalTransforms/pull/38. To allo wthe use in addons that support Plone 5.2 and older versions we should add it to 5.0 and 5.1 as well.

ale-rt:
`safe_nativestring` is really great!
A little bit off topic, what about also renaming `safe_unicode` and `safe_encode` to `safe_text` and `safe_bytes`?

Regarding the fact to backport that method to Plone <5.2 I am not sure if this will cause confusion.
You are adding a feature that will be available only for some minor versions of Plone.

That can cause some headaches.

jensens:
Closing as outdated (8 years old). Backport request to Plone 5.0/5.1, both long EOL. Moot in Python 3-only Plone 6.
If you think this issue is still relevant please reopen it. Thank you for the original report!

PR #2614, opened 2018-11-04
Rename safe_unicode to safe_text and safe_encode to safe_bytes

pbauer:
Keep old names as aliases. See #2596


ale-rt:
Test failure unrelated, should be fixed by https://github.com/plone/plone.app.testing/pull/55

Products/CMFPlone/utils.py
```diff
@@ -505,21 +505,27 @@ def safe_unicode(value, encoding='utf-8'):
     return value
 
 
-def safe_encode(value, encoding='utf-8'):
-    """Convert unicode to the specified encoding.
+safe_unicode = safe_text
```
hvelarde: why not include a deprecation warning on this?

davisagli: Because safe_unicode is used widely and that would then require a lot of work to get rid of the warnings. I don't see a problem with continuing to support the old name.

Issue #2626, opened 2018-11-11
Legacy resources like comments.js are not loaded


ksuess:
## BUG/PROBLEM REPORT (OR OTHER COMMON ISSUE)
legacy resources like comments.js are not loaded

### What I did:

Site root on domain.com/somepath/myroot/
_myroot_ is the name of the Plone instance.

### What I expect to happen:
++plone++static/plone-legacy-compiled.js includes comments.js

### What actually happened:
++plone++static/plone-legacy-compiled.js reads
```
...
/* reset requirejs definitions so that people who
   put requirejs in legacy compilation do not get errors */
var _old_define = define;
var _old_require = require;
define = undefined;
require = undefined;
try{

/* Could not find resource: https://domain.com/somepath/myroot/++resource++plone.app.discussion.javascripts/comments.js */
...
```

### What version of Plone/ Addons I am using:
Plone 5.1.4
commenting activated

jensens:
Closing as outdated (7 years old). About legacy resource compilation (plone-legacy-compiled.js). The entire legacy bundle system was removed in Plone 6.
If you think this issue is still relevant please reopen it. Thank you for the original report!

PR #2627, opened 2018-11-11
Load legacy resources like comments.js

ksuess:
Compile ++plone++static/plone-legacy-compiled.js with correct resource url of comments.js and others.
see #2626

### DIFF
--- a/Products/CMFPlone/resources/browser/cook.py
+++ b/Products/CMFPlone/resources/browser/cook.py
@@ -88,7 +88,7 @@ def cookWhenChangingSettings(context, bundle=None):
 
         if css_path:
             for css_resource in resource.css:
-                css_url = '/' + css_resource
+                css_url = css_resource
                 response = subrequest(css_url)
                 if response.status == 200:
                     logger.info('Cooking css %s', css_resource)
@@ -107,7 +107,7 @@ def cookWhenChangingSettings(context, bundle=None):
                     logger.warn('Could not find resource: %s' , css_resource)
         if not resource.js or not js_path:
             continue
-        js_url = '/' + resource.js
+        js_url = resource.js
         response = subrequest(js_url)
         if response.status == 200:
             js = response.getBody()
--- a/Products/CMFPlone/resources/browser/cook.py
+++ b/Products/CMFPlone/resources/browser/cook.py
@@ -88,7 +88,7 @@ def cookWhenChangingSettings(context, bundle=None):
 
         if css_path:
             for css_resource in resource.css:
-                css_url = siteUrl + '/' + css_resource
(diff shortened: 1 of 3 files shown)


ksuess:
@thet, you know this resource registry very well. Can you have a look?

jensens:
hm, whats the state here?

ksuess:

Products/CMFPlone/resources/browser/cook.py
```diff
@@ -107,7 +107,7 @@ def cookWhenChangingSettings(context, bundle=None):
                     logger.warn('Could not find resource: %s' , css_resource)
         if not resource.js or not js_path:
             continue
-        js_url = siteUrl + '/' + resource.js
+        js_url = resource.js
```
thet: what is the final url which is generated for js_url and css_url?
is it relative to the current context, e.g. when calling https://domain.com/subsite/context1/context2, is it https://domain.com/subsite/context1/context2/plone-logged-in.js or is it https://domain.com/subsite/plone-logged-in.js ?
If it's the first, that would be problematic as we cannot load the same resource for each context in the site separately.

ksuess: There is no final url for this resource (comment.js, event.js,..). 
js_url is used just for cooking. A somehow final url is the one of the cooked js plone-legacy-compiled.js

ksuess:
The issue is about **cooking** ++plone++static/plone-legacy-compiled.js

before:

    js_url = siteUrl + '/' + resource.js
    response = subrequest(js_url)

after:

    js_url = '/' + resource.js
    response = subrequest(js_url)


before means: subrequest is building path

    /VirtualHostBase/http/subdomain.example.com:80/VirtualHostRoot/_vh_somepath//somepath/Plone/++resource++plone.app.discussion.javascripts/comments.js
which is false and the resource cannot be found. So cook.py gets no response to include in cooked javascript ++plone++static/plone-legacy-compiled.js
cook.py prints the following in ++plone++static/plone-legacy-compiled.js:

    /* Could not find resource: https://subdomain.example.com/somepath/Plone/++resource++plone.app.discussion.javascripts/comments.js */

after means: subrequest is building path

    /VirtualHostBase/http/subdomain.example.com:80/VirtualHostRoot/_vh_somepath/Plone/++resource++plone.app.discussion.javascripts/comments.js
    which can be found and the response is included in cooked javascript ++plone++static/plone-legacy-compiled.js
cook.py prints the following in ++plone++static/plone-legacy-compiled.js:

    /* resource: ++resource++plone.app.discussion.javascripts/comments.js */
    if(require===undefined)require=function(reqs,torun){'use strict';return torun(window.jQuery);}; ...

Issue #2620, opened 2018-11-08
Duplicate Farsi translation for pickadate


mauritsvanrees:
I ran `check-manifest` in `Products.CMFPlone`` (5.0/5.1/5.2) and got this:

```
check-manifest 
lists of files in version control and sdist do not match!
missing from sdist:
  Products/CMFPlone/static/components/pickadate/lib/translations/fa_ir.js
suggested MANIFEST.in rules:
  recursive-include Products *.js
```

This is on a Mac, which is case insensitive or case unaware or whatever, and it does not see that there is both a [`fa_ir.js`](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/static/components/pickadate/lib/translations/fa_ir.js) file and a  [`fa_IR.js`](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/static/components/pickadate/lib/translations/fa_IR.js) file. I happen to get only the one with the two capital letters, and the other file is simply not there on my system.

Funnily enough, I *can* call `git log` on both:

```
$ git log Products/CMFPlone/static/components/pickadate/lib/translations/fa_ir.js
commit aa4d9c3f7179972c8f86a4085344d8c963bd4278
Author: Johannes Raggam <thetetet@gmail.com>
Date:   Tue Jun 21 14:06:15 2016 +0200

    Updated components directory, recompiled bundles.
✔ ~/community/plone-coredev/5.2/src/Products.CMFPlone [master|⚑ 1] 
20:56 $ git log Products/CMFPlone/static/components/pickadate/lib/translations/fa_IR.js
commit 0d7e8851e4b8a0bf05cf2d7a375ea2ed0a7fc88a
Author: Kumar Akshay <k.akshay9721@gmail.com>
Date:   Thu Mar 29 09:20:50 2018 +0530

    Revert "Touch fix v1.0"

commit 3e13920f11fe428a4ecb7cbff9be77257c3636b2
Author: Nehsus <sushenksharma@gmail.com>
Date:   Thu Mar 29 04:09:55 2018 +0530

    fixed untracked files

commit 3708355bca672a8e19e8d12952b92b4a401f44a1
Author: vangheem <vangheem@gmail.com>
Date:   Fri Jul 3 09:50:59 2015 -0500

    upgrade pickadate
```

@vangheem and @thet both committed one version.
All other files in that directory have two capital letters, so that is apparently the correct spelling.

Is there anything against removing `fa_ir.js`?

[...]

PR #2638, opened 2018-11-28
Delete fa_ir.js

mauritsvanrees:
We have `fa_ir.js` and `fa_IR.js`. That gives problems on Mac with case insensitive file system. See issue #2620. I had to create this PR through the github UI. :-)
Only the one with two capital letters, is needed. That is what is happening for other languages too.


mauritsvanrees:
I was inclined to directly do this on master, but probably better not to. But I only fired one Jenkins job, for 2.7. That seems enough.

PR #2646, opened 2018-12-04
Fix Folder layout fallback to default view if not existing

gbastien:
Call browserDefault.getLayout() with parameter check_exists=True to fallback to default view in case layout selected for Folder does not exist.

This is related to https://github.com/plone/Products.CMFDynamicViewFTI/issues/16

A test for this issue is added to Products.CMFDynamicViewFTI, a PR is waiting there (https://github.com/plone/Products.CMFDynamicViewFTI/pull/17).

Thank you,

Gauthier


mauritsvanrees:
PR job on Python 2 is succesful, on Python 3.6/7 it only fails with a known unstable plone.restapi test. Let's merge.

PR #2647, opened 2018-12-04
Restore the possibility to sort catalog query results with multiple indexes [5.1.x]

ale-rt:

PR #2651, opened 2018-12-06
[5.0.x] Restore the possibility to sort catalog query results with multiple indexes

### LINKED ISSUE #2464: Catalog sort_on on multiple columns is broken
ZCatalog can handle multiple sort_indexes. 

Products.ZCatalog.Catalog.py [1014] 
accepts a list of search indexes:
```
    def _getSortIndex(self, args):
        """Returns a list of search index objects or None."""
        sort_index_names = self._get_sort_attr("on", args)
        if sort_index_names is not None:
            # self.indexes is always a dict, so get() w/ 1 arg works
            sort_indexes = []
            if not isinstance(sort_index_names, (list, tuple)):
                sort_index_names = [sort_index_names]
            for name in sort_index_names:
                sort_index = self.indexes.get(name)
```
But commit 0909383eb2bbc8d31794a7dbf4392eec6236be22 breaks this 

CatalogTool.py [461]
```
        sort_on = kw.get('sort_on')
        if sort_on and sort_on not in self.indexes():
            # I get crazy sort_ons like '194' or 'null'.
           kw.pop('sort_on')
```
This assumes that sort_on is a index_name and fails if it is a list of index names.

Also the QueryBuilder has the same flaw:
plone.app.querystring.queryparser [53]
```
     # Add sorting (sort_on and sort_order) to the query
    if sort_on:
        catalog = getToolByName(context, 'portal_catalog')
        # I get crazy sort_ons like '194' or 'null'.
        if sort_on in catalog.indexes():
            query['sort_on'] = sort_on
            if sort_order:
                query['sort_order'] = sort_order
```
The later can be circumvented by using the custom-query parameter, but not the former.

### What I did:

results = catalog(**{ ...,  'sort_on' : ['title','description']], ... })

### What I expect to happen:

The result should be sorted after both indexes.

### What actually happened:

The result is not sorted at all.

5.1

Shall I make a pull request?

ale-rt:
Fixes #2464

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -1,12 +1,6 @@
-import logging
-import re
-import time
-import urllib
-
 from AccessControl import ClassSecurityInfo
 from AccessControl.PermissionRole import rolesForPermissionOn
-from AccessControl.Permissions import (
-    manage_zcatalog_entries as ManageZCatalogEntries)
+from AccessControl.Permissions import manage_zcatalog_entries as ManageZCatalogEntries  # noqa
 from AccessControl.Permissions import search_zcatalog as SearchZCatalog
 from Acquisition import aq_base
 from Acquisition import aq_inner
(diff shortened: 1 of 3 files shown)


ale-rt:
This is a backport to the 5.0.x branch of #2647 and #2648

PR #2663, opened 2018-12-12
Do not save type settings in "content-controlpanel" when switching between types.

cekk:
I've found two problems with previous implementation:
- every time you switch between types, settings are saved and we perform some unwanted db writes
- if we set "redirect_links" for Link type and save, and the switch to another type, this setting will be overwrite with a default (False) because the value isn't in the form.

With this change, we save settings only when the "Save" button is pressed.

This fix is for 5.1 branch, but i could fix it also on the master


cekk:
jenkins fails with some unrelated tests

PR #2668, opened 2018-12-18
fixes #2667 - cleanup deprecation warnings

### LINKED ISSUE #2667: Many deprecation warnings in 5.2
If running tests there are many deprecation warnings in Products.CMFPlone.

jensens:


Products/CMFPlone/tests/testSyndication.py
```diff
@@ -1,25 +1,25 @@
...
+class BaseSyndicationTest(PloneTestCase.PloneTestCase):
     def afterSetUp(self):
-        self.syndication = getToolByName(self.portal, 'portal_syndication')
-        self.folder.invokeFactory('Document', 'doc1')
-        self.folder.invokeFactory('Document', 'doc2')
-        self.folder.invokeFactory('File', 'file')
+        self.syndication = getToolByName(self.portal, "portal_syndication")
+        self.folder.invokeFactory("Document", "doc1")
+        self.folder.invokeFactory("Document", "doc2")
+        self.folder.invokeFactory("File", "file")
```
mauritsvanrees: Huh? Don't we prefer single quotes?

pyproject.toml
```diff
@@ -18,3 +18,7 @@ showcontent = true
 directory = "bugfix"
 name = "Bug fixes:"
 showcontent = true
+
+[tool.black]
```
mauritsvanrees: I haven't seen this one yet.
Note to self: read https://black.readthedocs.io/en/stable/index.html
Is this what caused the change from single to double quotes though?

mauritsvanrees: Apparently: yes. See rationale at https://black.readthedocs.io/en/stable/the_black_code_style.html#strings.
cc @gforcada

ale-rt: That's one of the reason I prefer https://github.com/google/yapf over black.

gforcada: I, and we are using that at work, would run black with the -S switch and keep the flake8 pluguins we are using for keeping the single quotes consistency

Products/CMFPlone/tests/testSyndication.py
jensens: huh, seems I used black with wrong params here. yeah, I would love to go to black defaults, but not now. First 5.2, then code style discussions ;)

pyproject.toml
gforcada: which two lines? :thinking:  :sweat_smile:

jensens: lines 23 and 24 in this file - `skip-string-normalization = true` means the same as `-S` on command line.

gforcada: :sweat_smile: sorry as you pointed to the black docs I thought they were there.

As for these two lines, I would remove the line length one, it is specially annoying for long imports, and as black takes care of it anyway, let's not bang our heads against it (IMHO)

Issue #448, opened 2015-04-21
Remove Products.CMFFormController dependency.


tisto:
http://jenkins.plone.org/view/Dependencies/job/plone-package-dependencies-cmf/36/artifact/deps/Products.CMFFormController.txt

```
Products/CMFPlone/factory.py:                u'Products.CMFFormController:CMFFormController',
Products/CMFPlone/setuphandlers.py:            'CMFFormController', 'Products.CMFFormController',
plone/app/upgrade/v30/tests.py:from Products.CMFFormController.interfaces import IFormControllerTool
plone/app/upgrade/v30/alphas.py:from Products.CMFFormController.interfaces import IFormControllerTool
plone/app/testing/layers.py:        ('Products.CMFFormController',           {'loadZCML': True}, ),
```

vangheem:
Seems like as long as we're checking ATContentTypes around, we can not remove CMFFormController in Plone 5. This will have to wait for Plone 5.1 when we do not ship with ATContentTypes.

tomgross:
The following core packages depend on CMFFormController:

 - [ ] CMFEditions
 - [ ] ATContentTypes
 - [ ] CMFPlone (login!)
 - [ ] PasswordResetTool
 - [ ] Archetypes

All non AT dependencies could be removed earlier. What's the state of plone.login BTW?

jensens:
This was addressed in the Plone 5.2 python3 migration and is removed already there (branch python3)

PR #2680, opened 2018-12-29
feat: remove Products.CMFFormController

gforcada:
Only a few traces were still left, most importantly on setup.py.

Fixes https://github.com/plone/Products.CMFPlone/issues/448


mauritsvanrees:
Well, I said 'looks good to me', but I didn't yet notice that there were many test failures. Seems we are not completely independent of CMFFormController yet.

gforcada:
After a closer look at the errors, it looks like we first need to clean up all the scripts on `skins` folder (in CMFPlone) as well as various other APIs.

On python2, as archetypes is there, makes things a *tad* more complicated :confounded:

gforcada:

PR #2688, opened 2019-01-07
Use correct permission for mail controlpanel form

fredvd:
Site Administrators cannot access this control panel allthough infrastructure with a separate permission is in place.


jensens:
@fredvd I think we can merge it after you add a changelog (news) entry.

ale-rt:

Issue #2694, opened 2019-01-10
Improper breadcrumbs port to Python 3: TypeError: decoding str is not supported


zopyx:
```
  /home/ajung/sandboxes/onkopedia_buildout-plone5/4f63a4b68fb9ec4821b03c1b801d2079.py(1310)render_content()
  /home/ajung/sandboxes/onkopedia_buildout-plone5/7c10bf8d506a40f6780bd59a8c66272d.py(2940)__fill_main()
  /home/ajung/sandboxes/onkopedia_buildout-plone5/7c10bf8d506a40f6780bd59a8c66272d.py(987)render_search_results()
> /home/ajung/sandboxes/onkopedia_buildout-plone5/eggs/Products.CMFPlone-5.2a2-py3.6.egg/Products/CMFPlone/browser/search.py(218)breadcrumbs()
-> empty = {'absolute_url': '', 'Title': six.text_type('…', 'utf-8')}
(Pdb) six.text_type('…', 'utf-8')
*** TypeError: decoding str is not supported
``` 
I don't see any  reason for the `six.text_type()` call here.

PR #2695, opened 2019-01-10
fixed Python 3 related str decoding issue in breadcrumbs

zopyx:
Fix for #2694

### DIFF
--- a/Products/CMFPlone/browser/search.py
+++ b/Products/CMFPlone/browser/search.py
@@ -214,7 +214,7 @@ class Search(BrowserView):
             return None
         if len(breadcrumbs) > 3:
             # if we have too long breadcrumbs, emit the middle elements
-            empty = {'absolute_url': '', 'Title': '…'}
+            empty = {'absolute_url': '', 'Title': u'…'}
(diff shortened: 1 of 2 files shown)


Products/CMFPlone/browser/search.py
```diff
@@ -214,7 +214,7 @@ def breadcrumbs(self, item):
             return None
         if len(breadcrumbs) > 3:
             # if we have too long breadcrumbs, emit the middle elements
-            empty = {'absolute_url': '', 'Title': six.text_type('…', 'utf-8')}
+            empty = {'absolute_url': '', 'Title': '…'}
```
ale-rt: You are right six.text_type does not make sense here.
But in order to have the title type coherent with the previous implementation I would replace:
```
empty = {'absolute_url': '', 'Title': '…'}
```
with:
```
empty = {'absolute_url': '', 'Title': u'…'}
```

news/2694.fixed
```diff
@@ -0,0 +1 @@
+fixed Python 3 related str decoding issue in breadcrumbs
```
gforcada: The filename suffix should be bugfix. I will fix it directly on master

Issue #2707, opened 2019-01-31
Unstable robot Scenario: A page is opened to edit in TinyMCE


mauritsvanrees:
The robot test "Scenario: A page is opened to edit in TinyMCE" is unstable. See [Jenkins](https://jenkins.plone.org/job/plone-5.2-python-3.6-robot-chrome/356/robot/Robot/Test%20Tinymce/Scenario%3A%20A%20page%20is%20opened%20to%20edit%20in%20TinyMCE/). It seems a bit better recently, but I have often seen pull requests fail because of this one.

Locally, today it always fails, both with Chrome and with Firefox.

I will submit a PR.

uwaiszaki:
@mauritsvanrees  Tests are also failing in 5.1.x. 
[https://jenkins.plone.org/job/pull-request-5.1/lastBuild/robot/report/robot_log.html#s1-s7-t1](url)
Should i start a new issue ?

mauritsvanrees:
My PR can probably be cherrypicked to 5.1

uwaiszaki:
I am working on plone 5.1 buildout. How can i cherrypick pr in 5.1 from master ? 
should i submit a new pr for this?

mauritsvanrees:
Create a new branch of CMFPlone master and ‘git cherry-pick’ the commit with the fix. Then push it and create a PR.

uwaiszaki:
@mauritsvanrees . I have created a pr [https://github.com/plone/Products.CMFPlone/pull/2716](url) .
Please review it.

mauritsvanrees:
Thanks! Merged.

PR #2708, opened 2019-01-31
Fixed unstable robot test Scenario: A page is opened to edit in TinyMCE.

mauritsvanrees:
Fixes https://github.com/plone/Products.CMFPlone/issues/2707.

Related note: I also see a few lines in this file that are meant to keep the tests working on an old Firefox 34. I tried cleaning those up, but that didn't work reliably with Firefox locally. So never mind that for now.


mauritsvanrees:
One approval, and all three Jenkins jobs pass, so I will merge.
Thanks.

PR #2759, opened 2019-02-19
Views for title and description.

iham:
adding views for title and description into 5.1.x too, to make this feature available asap


jensens:
Backporting sounds reasonable to me. Even if someone customized the main template it wont break anything, it just lacks the feature.

Issue #2765, opened 2019-02-20
Stop calling initializeTFC over and over


gforcada:
## BUG/PROBLEM REPORT

At production we use Products.LongRequestLogger and since we upgraded to 5.1.5 (yeah, sorry late at the party :sweat_smile: although the problematic code is still there on CMFPlone master ) we noticed that there is a huge amount of traces to this function: ``initializeTFC``:

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/PloneBaseTool.py#L23-L35

The problem is that as we don't have ATCT available, the function is called over and over on every single request.

The global ``TempFolderClass`` is never assigned anything, so it always tries to import ATCT, failing at it, and giving up (the ``pass``).

I guess that just changing the pass to a ``TempFolderClass = False`` would be enough. The function would still be called anyway, but the *expensive* ``pkg_resources.get_distribution`` would be called only once.

gforcada:
@ale-rt thanks for the merges!

davisagli:
Nice catch!

PR #2766, opened 2019-02-20
Only check once if ATCT is available

gforcada:
Otherwise the call to check if the distribution is available is done on every single call to ``getOAI``, which can be potentially hundreds of times per request. 

Fixes https://github.com/plone/Products.CMFPlone/issues/2765


gforcada:
@jenkins-plone-org please run jobs :+1:

PR #2769, opened 2019-02-20
Resolve circular dependency between Products.CMFPlone and plone.i18n.

### LINKED ISSUE #2049: Circular dependency between CMFPlone and plone.i18n
Since https://github.com/plone/plone.i18n/pull/18 was merged, ``plone.i18n`` [now depends](https://github.com/plone/plone.i18n/blob/0e49ae551e91de3a7190c253ffe7986879cc19e5/setup.py#L33) on ``Products.CMFPlone``, just like ``Products.CMFPlone`` [depends on plone.i18n](https://github.com/plone/Products.CMFPlone/blob/049c4a6cbc41f45a675cd65f5f25919a32446cf0/setup.py#L108). 🎉 surprise, surprise, a new circular dependency is born (or rather, an implicit one is made explicit).

On plone.i18n there is really only one place where CMFPlone is imported: https://github.com/plone/plone.i18n/blob/69c712da9ea01dd1922d27c3268dd55d7989caae/plone/i18n/utility.py#L11 the purpose there is to get language related registry settings.

Should this dependency be broken from plone.i18n side and not rely on the interface but rather use a string for the key?

Should this dependency be broken from CMFPlone side and move the interface to plone.i18n and update all imports?

sallner:
In https://github.com/plone/plone.i18n/pull/30 the interface was integrated in `plone.i18n`. This PR uses it in CMFPlone.

Fixes #2049


sallner:

sallner:
I am unsure with this line https://github.com/plone/Products.CMFPlone/blob/a2ce0c7c3bd71be05a112e27680041481ab82f41/Products/CMFPlone/profiles/dependencies/registry.xml#L5 Should this be integrated in `plone.i18n` or should the dependency point to `plone.i18n` as there are just dependencies to `Products.CMFPlone` declared in this file, I don't know how to proceed?

gforcada:
IMHO I would keep it here and update the reference.

This way, plone.i18n does not need a registry.xml.

We might need an upgrade step to copy over the settings I guess, but someone with more upgrade-foo should comment @mauritsvanrees or @pbauer, could you? 😃

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -155,180 +154,6 @@ class IEditingSchema(Interface):
         required=False)
```
jensens: I would add here also a `zope.deferredimport.deprecated` in case someone imported from here.

sallner: Yes, it seems that this is necessary, I had it first there and then thought, that the API is always imported from `Products.CMFPlone.interfaces`

sallner:

gforcada:
Sorry I closed it by accident 😅

sallner:

and there is goes again :cyclone:

jensens:
@sallner thanks so far, now there is "only" the upgrade step missing migrating existing registry records to the new location. Would you give this a try? I would start by asking this as a question at https://community.plone.org

gforcada:
@jensens I hold back on merging this one until we had the upgrade steps, merging the plone.i18n was fine as even if it gets released, no one is pointing to that interface there, so no harm is done.

Here though, we need the plone.i18n released at the same time, but specially the upgrade steps, if not, anyone trying to upgrade from 5.2alphas or betas will get a non-working language control panel.

And most of it all I'm saying it because in a week, after the plone tagung, I know that a new release will be asked to the release.

So getting those upgrade steps will definitely be a blocker :wink: 

Now, you have something to work on the sprints :stuck_out_tongue_winking_eye:

jensens:
I will look at this before release if nobody else does. Should not be that difficult to do, need to dig a bit into registry internals.

gforcada:
I don't think you need to go to the registry internals, just plain old *copy && paste && delete*:
pseudo code ahead:

```
foo1 = api.get_registry_value(my_field)
api.set_registry_value(my_new_field=foo1)
foo2 = api.get_registry_value(my_field)
api.set_registry_value(my_new_field=foo2)
foo3 = api.get_registry_value(my_field)
api.set_registry_value(my_new_field=foo3)
```

the interface is still importable from CMFPlone, so it should not break getting the value...

Issue #2763, opened 2019-02-20
Run Plone 5.2 on wsgi by default


pbauer:
Plone 5.2 should run on wsgi using waitress by default unless the admin configures it to run on ZServer. The coredev is already set up like that but a clean minimal buildout is not unless you set `wsgi=on` for `plone.recipe.zope2instance`. 

I suggest to set `wsgi=on` as the default and the fallback to ZServer would only happen if it is explicitly set to `off` (and running on Python 2). In Python 3 the setting could be ignored altogether.

We should also display if it is running on wsgi or ZServer in /@@overview-controlpanel

pbauer:
Displaying the wsgi-status and server is started in https://github.com/plone/Products.CMFPlone/pull/2770.

The change in the recipe is still a TODO.

pbauer:
See https://github.com/plone/plone.recipe.zope2instance/pull/73 for the change in the recipe.

PR #2770, opened 2019-02-22
Display wsgi-status and name/version of used server

pbauer:


pbauer:
@tschorr do you know how to find the name of the wsgi-server we use while the instance is running?

pbauer:

tschorr:
> @tschorr do you know how to find the name of the wsgi-server we use while the instance is running?

No, and I can't find anything about this in https://www.python.org/dev/peps/pep-3333/#id23. So I have some doubts that it is possible to do this in an explicit, straightforward way for all available middleware.

pbauer:
I changed the code to do some guessing using the request. At least it works fine for ZServer and waitress.

jensens:
Look good enough to me. Later improvement is always possible.

ale-rt:
I would suggest something like this to have transaltions:
```
    <p tal:define="
         server_info view/server_info;
         has_wsgi server_info/wsgi;
       ">
      <ul>
        <li>
          <tal:i18n i18n:translate="">WSGI:</tal:i18n>
          <span i18n:translate="" tal:condition="has_wsgi">On</span>
          <span i18n:translate="" tal:condition="not:has_wsgi">Off</span>
        </li>
        <li>
          <tal:i18n i19n:translate="">Server:</tal:i18n>
          <span>${server_info/server_name}</span>
          <span>${server_info/version}</span>
        </li>
      </ul>
    </p>
```
Use of i18n:name is a plus.

tschorr:
@jensens @pbauer why not, it's definitely worth a try. And it should also work for the ZServer case, so maybe the code could be shortened a bit.

pbauer:
> > We should also display if it is running on wsgi or ZServer in /@@overview-controlpanel
> 
> My assumption when reading this in #2763 was that we want to display we are running on WSGI full stop.

Yes, it should be the default but my thinking was that you can to opt-out by setting `wsgi=off` when running Python 2. That would give you ZServer and support for FTP and WebDAV.

tschorr:
> > > We should also display if it is running on wsgi or ZServer in /@@overview-controlpanel
> > 
> > 
> > My assumption when reading this in #2763 was that we want to display we are running on WSGI full stop.
> 
> Yes, it should be the default but my thinking was that you can to opt-out by setting `wsgi=off` when running Python 2. That would give you ZServer and support for FTP and WebDAV.

That's fine for me. I was just assuming we would only display either ZServer or WSGI and not bother about the actual WSGI middleware/server software used.

tschorr:
just tried uWSGI - It shows server: unknown for both uwsgi and http socket.

tschorr:
... gunicorn sets the SERVER_SOFTWARE variable so it will already contain version info:

gunicorn/__init__.py:8:SERVER_SOFTWARE = "gunicorn/%s" % __version__

jensens:
WSGI 2 proposal makes SERVER_SOFTWARE required: https://wsgi.readthedocs.io/en/latest/proposals-2.0.html#making-some-keys-required
I have no idea about the state of the proposal.

tschorr:
> I have no idea about the state of the proposal.

@jensens the note at the top of https://wsgi.readthedocs.io/en/latest/proposals-2.0.html#proposals-related-to-wsgi-2-0 doesn't sound very encouraging to me :-)

tschorr:
... but still I think using SERVER_SOFTWARE to get some information about the middleware/server software is a good idea and possibly the best option we have.

jensens:
> doesn't sound very encouraging to me

That is why I wrote: no idea.

pbauer:
Ok, I added i18n and changed it a little so that server_version is only shown when server_name is a package. In some cases (like gunicorn) server_name also holds the version.

pbauer:

Products/CMFPlone/controlpanel/browser/overview.py
```diff
@@ -45,6 +48,31 @@ def core_versions(self):
...
         return 'PIL' in self.core_versions()
 
+    def server_info(self):
+        wsgi = isinstance(self.request, WSGIRequest)
+        server_name = 'unknown'
+        server_version = ''
+
+        # check for ZServer
+        servers = getattr(getConfiguration(), 'servers', None)
+        if servers and 'ZServer' in servers[0].__module__:
```
tschorr: ZServer also sets SERVER_SOFTWARE , so it could be handled in the same way as the WSGI servers. This would simplify the code.

pbauer: Oh, I missed that. I update the code.

pbauer:

Issue #2771, opened 2019-02-23
Plone 5.2: improper handling of `LOGIN_TEMPLATE_IDS` in `Products.CMFPlone.browser.login.login.LoginForm.get_came_from`


d-maurer:
## BUG/PROBLEM REPORT (OR OTHER COMMON ISSUE)
BUG

### What I did:
`http://localhost:8080/Plone-IDP/acl_users/credentials_cookie_auth/require_login?came_from=http%3A//lsdm%3A8180/Plone-IDP/idp/idpsso_logged_in%3Fskey%3Drs%253A846c1bbb-5037-4f06-8342-7e587529caa9`

You cannot reproduce this exactly. However, there should be no need: the important thing is that the path of `came_from` contains `logged_in` (but not a `LOGIN_TEMPLATE_IDS`).

### What I expect to happen:
That after a successful login, there would be a redirect to `came_from`.

### What actually happened:
`came_from` was ignored.

The reason: `Products.CMFPlone.browser.login.login.LoginForm.get_came_from` looks for `LOGIN_TEMPLATE_IDS` in the path of `came_from` (here `Plone-IDP/idp/idpsso_logged_in`)
and wrongly believes, it contains `logged_in` (while this is actually `idpsso_logged_in`).
I think, that `get_came_from` should only look at the last path component of `came_from` (here `idpsso_logged_in`) and check for identity; at the least it should not compare on string level but instead split the path into its segments and check whether the login template id is one of the segments.

### What version of Plone/ Addons I am using:
Plone 5.2aX

`dm.zope.saml2` (unreleased; I am working to get it Plone 5.2/Zope 4/Python 3 compatible).

pbauer:
@d-maurer please tag tickets according to the right milestone (5.2 in this case) and make sure that the issue is thill the case with the newest version (5.2b1 or - even much better - coredev).

d-maurer:
Philip Bauer wrote at 2019-2-23 02:13 -0800:
>@d-maurer please tag tickets according to the right milestone (5.2 in this case)

I cannot: "Labels", ... "Milestone", etc. are all passive elements for me
(likely because I have no write permissions).

>and make sure that the issue is thill the case with the newest version (5.2b1 or - even much better - coredev).

In fact, I am using "coredev" -- however, I do not know
which version it corresponds to. Can you give me a tip how to
determine this?

I have just made a `git pull` in my local "Products.CMFPlone" git clone
[...]

PR #2783, opened 2019-03-02
fix redirection to came_from after login

petschki:
see #2771 

I solved this by checking the whole path to make path traversal after login also a working possibility


petschki:

PR #2791, opened 2019-03-08
Fix wrong default values for msgid label_email

### LINKED ISSUE #2790: Wrong default values for msgid label_email
label_email msgid is used with a wrong default value:

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/login/login.py#L106

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/login/login_help.py#L62

Plone is already using "E-mail" as default value elsewhere:

https://github.com/plone/plone.app.portlets/blob/master/plone/app/portlets/portlets/login.pt#L48

https://github.com/plone/plone.app.users/blob/master/plone/app/users/schema.py#L67

erral:
Fixes #2790

### DIFF
--- a/Products/CMFPlone/browser/login/login.py
+++ b/Products/CMFPlone/browser/login/login.py
@@ -104,7 +104,7 @@ class LoginForm(form.EditForm):
 
         if self.use_email_as_login():
             self.widgets[fieldname_name].label = _(u'label_email',
-                                                   default=u'Email')
+                                                   default=u'E-mail')
         self.widgets['came_from'].mode = HIDDEN_MODE
         self.widgets['came_from'].value = self.get_came_from()
 
--- a/Products/CMFPlone/browser/login/login_help.py
+++ b/Products/CMFPlone/browser/login/login_help.py
@@ -60,7 +60,7 @@ class RequestResetPassword(form.Form):
         if self.use_email_as_login():
             self.widgets['reset_password'].label = _(
                 u'label_email',
-                default=u'Email'
+                default=u'E-mail'
             )
 
     @button.buttonAndHandler(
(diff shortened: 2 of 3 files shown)


erral:

Products/CMFPlone/browser/login/login.py
```diff
@@ -104,7 +104,7 @@ def updateWidgets(self):
 
         if self.use_email_as_login():
             self.widgets[fieldname_name].label = _(u'label_email',
-                                                   default=u'Email')
+                                                   default=u'E-mail')
```
jensens: Proof me wrong, but usually in English it is written `Email`` even if e-mail is allowed too:
https://en.oxforddictionaries.com/definition/email

erral: I was also searching for the correct way and it seems that both are OK... The point of the PR was that we are already using "E-mail" somewhere else for long time (https://github.com/plone/plone.app.portlets/blob/master/plone/app/portlets/portlets/login.pt#L48 or https://github.com/plone/plone.app.users/blob/master/plone/app/users/schema.py#L67 for example), and the point was to unify them all changing this one, because it's the latest thing that was added here.

jensens: Lets ask some native speakers? Ping @tkimnguyen @MatthewWilkes

MatthewWilkes: As @jensens says, both are acceptable but "email" is more common these days. That said, I agree with @erral's logic, it's better to use the slightly less common (but still correct) form everywhere than to have a mixture.

thet:
Regarding Wikipedia, E-mail is the least common form:
https://en.wikipedia.org/wiki/Email#Terminology

PR #2811, opened 2019-03-19
Fix some more unstable robot tests by adding waits

mauritsvanrees:
A few of these have actually failed at some point in Jenkins, others probably not. But it seems that if you go to a url or click somewhere, that it is best to first wait on something before you do a test assertion: the browser may need a few (micro) seconds to get to the point where the assertion will pass.  So I added some waits.


mauritsvanrees:

PR #2822, opened 2019-03-27
Don't use // as bundle comment separators Closes #2820

### LINKED ISSUE #2820: Plone 5.1.4 to 5.1.5 update: resource registry meta bundle generator comments first css construct of individual bundles
In one of our projects, after upgrading from Plone 5.1.4 to Plone 5.1.5  A very small part of the css became broken in plone.app.mosaic layouts . Images inside a tile no longer had a "height: auto" on them. This is normally included in mosaic-styles.css , and the mosaic styles were included in default.css.  

We quickly patched the missing statement into our theme file and did a patch release, but the underlying problem was vague.  The problem would only appear on production, running the site locally did not show the problem, so my attention was pulled to the metabundle generation. This was modified between 5.1.4 and 5.1.5 in https://github.com/plone/Products.CMFPlone/commit/397918cd39ba0be4e2e150df5f5f2220e6ecc828 by @vangheem 

The problematic code is in this part:

https://github.com/plone/Products.CMFPlone/blob/2195c4a43ba100fb2b7973dccb4299dad2de42fe/Products/CMFPlone/resources/browser/combine.py#L123-L132

The individual bundles are separated by comment lines with // Start bundle and // End Bundle, but // actually comments out the first following css construct , more info at  https://www.xanthir.com/b4U10

And the mosaic-styles.css individual bundle start with:

```
// Start Bundle: mosaic-styles
/* Images will never be bigger then a tile */
.mosaic-tile img {
  max-width: 100%;
  height: auto;
}
```

If even skips over the /* */ comment on the next line and comments the first {} it sees
So that is how only our height: auto; got disabled in production . 

This is  at the moment only broken in Plone 5.1 , In Plone 5.2 the whole bundler got rewritten again, partly because of Python3 support and doesn't seem to insert these comments. I have swapped the // comment for /* */ comments and this also solves the problem. I'll create a pull request shortly.

fredvd:
Don't use // as bundle separators in the resource registry's meta bundle generator. It comments the first css construct in every bundle that follows. 

Doesn't need a fix for 5.2/master as the bundle generation was completely rewritten for python3/unicode support.

### DIFF
--- a/Products/CMFPlone/resources/browser/combine.py
+++ b/Products/CMFPlone/resources/browser/combine.py
@@ -124,9 +124,9 @@ class MetaBundleWriter(object):
         fi = StringIO()
         for bname, script in resources.items():
             fi.write('''
-// Start Bundle: {0}
+/* Start Bundle: {0} */
 {1}
-// End Bundle: {2}
+/* End Bundle: {2} */
 '''.format(bname, script, bname))
         self.folder.writeFile(self.name + postfix, fi)
         resources.clear()
--- a/news/2820.bugfix
+++ b/news/2820.bugfix
@@ -1,0 +1,1 @@
+Don't use // as bundle separators in the resource registry's meta bundle generator. It comments the first css construct in every bundle that follows. [fredvd]


fredvd:

PR #2829, opened 2019-04-10
fix show_excluded_items regression

### LINKED ISSUE #2828: exclude_from_nav and show_excluded_items are not handled correctly in portal_tabs_view
### What I did:

Create a folder on top level and exclude id from navigation

### What I expect to happen:

the folder should disappear from global navigation **when I'm not inside it's context**

### What actually happened:

the folder is always in the navigation

plone 5.2rc2

### REASON:

I'm fixing a bug found here: https://github.com/plone/plone.app.layout/issues/190
which leads to the same result in `portal_tabs_view` here https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/navigation.py#L151-L153

if `show_excluded_items` is activated (which actually is the default setting) every item is shown regardless of `exclude_from_nav` setting. But what the setting claims is to show the excluded item only if you're inside of it.

PR is in preparation...

petschki:
fixes #2828

### DIFF
--- a/Products/CMFPlone/browser/navigation.py
+++ b/Products/CMFPlone/browser/navigation.py
@@ -72,7 +72,7 @@ class CatalogSiteMap(BrowserView):
 class CatalogNavigationTabs(BrowserView):
 
     def _getNavQuery(self):
-                # check whether we only want actions
+        # check whether we only want actions
         registry = getUtility(IRegistry)
         navigation_settings = registry.forInterface(
             INavigationSchema,
@@ -100,6 +100,9 @@ class CatalogNavigationTabs(BrowserView):
             query['review_state'] = navigation_settings.workflow_states_to_show
 
         query['is_default_page'] = False
+
+        if not navigation_settings.show_excluded_items:
+            query['exclude_from_nav'] = False
 
         if not navigation_settings.nonfolderish_tabs:
             query['is_folderish'] = True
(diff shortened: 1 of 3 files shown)


petschki:

petschki:
This needs https://github.com/plone/plone.app.layout/pull/192 too ... I'll start the Jenkins jobs manually with both PR urls

Products/CMFPlone/browser/navigation.py
```diff
@@ -146,11 +149,13 @@ def _get_url(item):
...
             return get_view_url(item)
 
+        context_path = '/'.join(context.getPhysicalPath())
+
         # now add the content to results
         for item in rawresult:
-            if not navigation_settings.show_excluded_items:
-                if item.exclude_from_nav:
-                    continue
+            if not context_path.startswith(item.getPath()) and item.exclude_from_nav:  # noqa: E501
```
mauritsvanrees: Same remark as in the plone.app.layout PR: first check `item.exclude_from_nav` for slightly better performance.

petschki: man @mauritsvanrees you're faster than me 😉 ... done

gforcada: *(sorry for the 4 months delay)*

If this is about performance, getting Asko's default getattr branch in plone.dexterity would be much more efficient, though some adjustments have to be done, like here :)

Not all objects might provide an `exclude_from_nav` attribute, so a `getattr()` would be safer IMHO

PR #2839, opened 2019-04-15
Pythonify templates

pbauer:
Using `@@` should speed up acquiring the template since it bypasses attributes, properties and skin-templates.

I'd prefer using `python:context.restrictedTraverse('@@main_template').macros['master']` but that is  not possible because ttw-customized templates (using `portal_view_customization`) do not allow to call the property `macros`. I don't know why...


pbauer:

PR #2840, opened 2019-04-15
gracefully handle tracebacks when installing addons during site creation

### LINKED ISSUE #2228: AttributeError: 'PersistentAdapterRegistry' object has no attribute 'lookupAll'
The following error is raised in Plone 5.1 when you try to install a new Plone site with an add-on where the Dexterity content-class could not be imported properly e.g. for some NameError in my case.

In particular this error message shadows the primary problem. It seems to me that some errors in the bootstrapping phase of Plone or inside of GS are just swallowed.

```
2017-11-30 11:29:03 ERROR txn.140599418205952 Failed to abort resource manager: <Products.CMFCore.indexing.QueueTM object at 0x7fdfddecbd50>
  File "/home/ajung/.buildout/eggs/transaction-2.1.2-py2.7.egg/transaction/_transaction.py", line 514, in abort
    rm.abort(self)
  File "/home/ajung/.buildout/eggs/Products.CMFCore-2.2.12-py2.7.egg/Products/CMFCore/indexing.py", line 321, in tpc_abort
    self.queue.abort()
  File "/home/ajung/.buildout/eggs/Products.CMFCore-2.2.12-py2.7.egg/Products/CMFCore/indexing.py", line 237, in abort
    for name, util in sm.getUtilitiesFor(IIndexQueueProcessor):
  File "/home/ajung/.buildout/eggs/zope.interface-4.4.2-py2.7-linux-x86_64.egg/zope/interface/registry.py", line 290, in getUtilitiesFor
    for name, utility in self.utilities.lookupAll((), interface):
AttributeError: 'PersistentAdapterRegistry' object has no attribute 'lookupAll'
```

petschki:
fixes #2228

This shows a warning instead of the mentioned traceback (https://github.com/plone/Products.CMFPlone/issues/2228#issuecomment-483191236) after creating the site ... 

Bildschirmfoto 2019-04-15 um 12 24 47

### DIFF
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from Products.CMFPlone import PloneMessageFactory as _
 from Products.CMFPlone.Portal import PloneSite
 from Products.CMFPlone.events import SiteManagerCreatedEvent
 from Products.CMFPlone.interfaces import INonInstallable
@@ -168,11 +169,15 @@ def addPloneSite(context, site_id, title='Plone site', description='',
             setup_tool.runAllImportStepsFromProfile(
                 'profile-%s' % extension_id)
         except Exception as msg:
-            IStatusMessage(request).add(
-                'Could not install {}: {}\n'
+            IStatusMessage(request).add(_(
+                'Could not install ${profile_id}: ${error_msg}! '
                 'Please try to install it manually using the "Addons" '
                 'controlpanel and report any issues to the '
-                'addon maintainers'.format(extension_id, msg.args),
+                'addon maintainers.',
+                mapping={
+                    'profile_id': extension_id,
+                    'error_msg': msg.args,
+                }),
                 type='warning')
 
     if snapshot is True:
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -3,6 +3,7 @@ from Products.CMFPlone.Portal import PloneSite
 from Products.CMFPlone.events import SiteManagerCreatedEvent
 from Products.CMFPlone.interfaces import INonInstallable
 from Products.GenericSetup.tool import SetupTool
+from Products.statusmessages.interfaces import IStatusMessage
 from plone.registry.interfaces import IRegistry
 from zope.component import queryUtility
 from zope.event import notify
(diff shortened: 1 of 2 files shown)


petschki:
and likely a backport to `5.1` too

petschki:

petschki:
> Alternatively, we could get the original site or site manager (I am never sure of the difference) before creating the Plone Site, and restore that one in a try/finally. Then perhaps the current unhelpful lookupAll AttributeError would be solved as well. That is just a hunch that I now have when looking at the code.

@mauritsvanrees sorry I don't know what you mean with that. As I just skip the faulty profile I don't think we need a finally clause here. The site gets created correctly with the info what to do with that addon. That's it.

Maybe @polyester has some wording recommendations for that issue. I'm happy to fix that.

mauritsvanrees:
You could argue that a faulty profile *should* raise an error, and that creating the Plone Site *should* fail. That is what was happening so far. Your fix displays a warning and happily continues with the next add-on.

What I mean is an alternative to your fix, or an addition to it. Pseudo code:

```
orig_site = getSite()
site = create Plone site without add-ons
setSite(site)
try:
    install add-ons
except:
    add warning status message
finally:
    setSite(orig_site)
return orig_site  # or None, or context, or whatever the calling function needs.
# after this, in case of an error we should display the add site form again,
# with the warning message. No Plone Site should have been created.
```

I guess this means I am half arguing *against* merging, instead of my approval. ;-) But at least your fix gets things going.

petschki:
Ok ... personally I'm more into moving forward, but now I understand your objections. If there's a consensus that we should fail rather than warn about the error, I'll change that. So we need at least one more voice to make a decision.

/cc @jensens @pbauer @agitator maybe?

jensens:
I would turn the message into an error. And indeed, thinking about it, a ``logger.exception(same_message_as_in_ui)`` in the except is a must have! and go on with this fix to get things done. 

In a second step we can try to improve the situation to let the whole site install fail, I suppose this is what a administrator would expect if one addon fails.

petschki:
Current output:

Bildschirmfoto 2019-04-16 um 11 23 33

Log:

Bildschirmfoto 2019-04-16 um 11 24 04

petschki:

jensens:

PR #2857, opened 2019-05-09
Allow empty default_page setting

### LINKED ISSUE #2813: Can not save website controlpanel when default page is empty.
In some project I have an empty `default page` setting in my registry. 
```
  <records
      interface="Products.CMFPlone.interfaces.controlpanel.ISiteSchema"
      prefix="plone">
    <value key="default_page"></value>
...
```

After save of Website settings I got a validation error "required input is missing" (German: Erforderliche Eingabe fehlt.). But there was no field marked, its floats free between robots and roles field. With inspector I see there a hidden input:
```
<input id="form-widgets-default_page" name="form.widgets.default_page" value="" class="hidden-widget" type="hidden" data-_extension-text-contrast="bg">
```
The field is required but hidden with an empty value, so this is what it is causing the problem.

petschki:
fixes #2813

### DIFF
--- a/Products/CMFPlone/interfaces/controlpanel.py
+++ b/Products/CMFPlone/interfaces/controlpanel.py
@@ -1155,7 +1155,7 @@ class ISiteSchema(Interface):
         description=_(
             u'Select which IDs (short names) can act as fallback '
             u'default pages for a container.'),
-        required=True,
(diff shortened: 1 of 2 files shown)


petschki:

PR #2899, opened 2019-06-24
add plone.resource as a dependency before plone.app.discussion

pbauer:
This fixes `combine_bundles` with static resources in the legacy-bundle during site-creation
With https://github.com/plone/plone.app.discussion/pull/158 p.a.discussion gains a resource in the legacy-bundle thus triggering the combination of bundles. But at this state the tool `portal_resource` and the utility `IResourceDirectory` are not yet initialized. 

See https://github.com/plone/plone.app.discussion/issues/157


pbauer:

jensens:
If pa.discussion needs it put it in its metadata.xml, not in here. Otherwise future side effects are guaranteed.

pbauer:
@jensens good point, that works as well. I thought I tried that and it failed as well but now it works. Must have had the order or dependencies wrong.

Issue #2890, opened 2019-06-20
PLIP: Add support for Python 3.8


pbauer:
## Abstract

Python 3.8 is in beta right now and the final final release is scheduled for October 21st 2019. (see https://www.python.org/dev/peps/pep-0569/#schedule)

We should make sure that Plone 5.2 supports it asap. Zope 4.1 already has preliminary support for Python 3.8. 

Python 3.8 has many new features and a bunch of spee-optimizations from which we might also benefit. See https://docs.python.org/3.8/whatsnew/3.8.html for details

## Motivation

Keeping up-to-date with current Python versions.

## Assumptions

Updating our codebase should be relatively easy. Some deprecated imports need to be replaces (most notably `from cgi import escape`) but I do not expect bigger problems. 

## Proposal & Implementation

* Fix all issues. 
* Add official support during a release in the Plone 5.2.x series

## Deliverables

* [x] Pull-requests for all packages that do not work in Python 3.8
* [x] Create jenkins-jobs for Python 3.8
* [x] Fix all issues to get the build green
* [ ] Update documentation to mention 3.8

## Risks

* Jenkins needs to run 4 jobs instead of three which uses more time and resources

## Participants

* Philip Bauer

pbauer:
I implemented support for Python 3.8 with the following pull-requests:

* [ ] https://github.com/plone/Products.CMFPlone/pull/2896
* [ ] https://github.com/plone/diazo/pull/77
* [ ] https://github.com/plone/plone.app.contentmenu/pull/25
* [ ] https://github.com/plone/plone.app.dexterity/pull/298
* [ ] https://github.com/plone/plone.app.layout/pull/200
* [ ] https://github.com/plone/plone.app.users/pull/82
* [ ] https://github.com/plone/plone.app.viewletmanager/pull/21
* [ ] https://github.com/plone/Products.CMFDiffTool/pull/36
* [ ] https://github.com/plone/Products.CMFFormController/pull/20
* [ ] https://github.com/zopefoundation/Products.DCWorkflow/pull/16

With these Plone seems to run fine on Python 3.8. 
But there are still some failing tests and a lot of `SyntaxWarnings` that need to be adressed. 

### 1. SyntaxWarnings

When running the application or the test many warnings such as these appear:

```
1b119b71d694960f28514f723a94273c.py:305: SyntaxWarning: "is" with a literal. Did you mean "=="?
[...]

PR #2904, opened 2019-06-29
Declare support for Python 3.8

pbauer:
Since https://github.com/plone/Products.CMFPlone/issues/2890 is done.


pbauer:
Since Zope does that (see https://github.com/zopefoundation/Zope/blob/master/setup.py) I think we should as well.

pbauer:

PR #2915, opened 2019-07-08
Fix #2914: Fix the TinyMCE pattern loading HTML as CSS...

### LINKED ISSUE #2914: The TinyMCE pattern loads HTML as CSS and slows edit forms
### What I did:

Open an `@@edit` form with the browser's network inspector open.

### What I expect to happen:

It should load quickly and with the correct resources.

### What actually happened:

For every field using the TinyMCE pattern, there was a network request for the portal root rendered at HTML but loaded as CSS.  This takes a lot of time and greatly slows the page load time.

Products.CMFPlone==5.1.5
plone.app.theming==2.0.5

rpatterson:
For some reason, plone.app.theming===2.0.5 [replaces empty resource lists with
empty
strings](https://github.com/plone/plone.app.theming/blob/2.0.x/src/plone/app/theming/utils.py#L318).
While this seems less than ideal to me, that using None/null is more explicit,
it still seems like a bug that in the pattern here the empty string ends up
adding the nav root URL to the list of CSS files to load.  This results in a
warning in the browser console and greatly slows down loading the edit form.

### DIFF
--- a/Products/CMFPlone/patterns/tinymce.py
+++ b/Products/CMFPlone/patterns/tinymce.py
@@ -48,7 +48,7 @@ class TinyMCESettingsGenerator(object):
                 files.append('/'.join([self.nav_root_url, url.strip()]))
         theme = self.get_theme()
         tinymce_content_css = getattr(theme, 'tinymce_content_css', None)
-        if tinymce_content_css is not None:
+        if tinymce_content_css:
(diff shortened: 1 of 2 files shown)


jensens:

jensens:
Does it need a forward port to master?

jensens:
@rpatterson ping?

PR #2943, opened 2019-08-23
Fix typo in README

khink:
As this is the first bit of text one sees on https://pypi.org/project/Products.CMFPlone/, i think it deserves extra attention.

https://www.grammarly.com/blog/its-vs-its/


mauritsvanrees:
Thanks for the fix. I have merged it manually on the command line.

PR #2948, opened 2019-09-15
update frontpage.pt

polyester:
use https and correct links


polyester:

Issue #2964, opened 2019-10-24
Typo in 4.3.19 causes plone site to crash


avolkov:
## BUG/PROBLEM REPORT (OR OTHER COMMON ISSUE)

On upgrade the site crashes displaying tal error. It seems to be a syntax error issue, I will attach

### What I did:
Upgrade plone 4.1 to plone 4.3

### What I expect to happen:
Expect existing site running

### What actually happened:
Crash with the following message

```
PTRuntimeError('Page Template main_template has errors:
 [\'Compilation failed\', \'zope.tal.taldefs.TALError: 
Python expression error:\\nEOL while scanning string literal (PythonExpr, line 1) 
in expression u"python:request.RESPONSE.setHeader(\\\'X-XSS-Protection\\\', \\\'1", at line 6, column 1\']',)
(Also, the following error occurred while attempting to render the standard error message, 
please see the event log for full details:
Page Template main_template has errors: 
['Compilation failed', 'zope.tal.taldefs.TALError: Python expression error:
EOL while scanning string literal (PythonExpr, line 1) in expression 
u"python:request.RESPONSE.setHeader(\'X-XSS-Protection\', \'1", at line 6, column 1'])
```

### What version of Plone/ Addons I am using:

    Plone 4.3.19 (4321)
    CMF 2.2.10
    Zope 2.13.29
    Python 2.7.14 (default, May 17 2018, 16:02:42) [GCC 7.3.0]
    PIL 3.3.3 (Pillow)

The bug in this package in particular is in Producs.CMFPlone-4.3.19

The proble is in **skins/plone_templates/main_template.pt** in branch 4.x on line 25 it seems that one of the semicolons is not properly escape. I'll add a pull request that fixes this.

avolkov:
Here is the full traceback of the issue

`s2019-10-24 14:59:16 ERROR Zope.SiteErrorLog 1571943556.10.501197343689 http://localhost:8080/poa/@@editing-controlpanel
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module zope.formlib.form, line 873, in __call__
  Module zope.formlib.form, line 863, in render
  Module Products.Five.browser.pagetemplatefile, line 125, in __call__
  Module Products.Five.browser.pagetemplatefile, line 59, in __call__
[...]

PR #2972, opened 2019-10-31
Removed broken X-XSS-Protection header.

mauritsvanrees:
Fixes https://github.com/plone/Products.CMFPlone/issues/2964 for the unstyled theme (Plone Default).

Added robot test to check that the homepage renderers without error in Plone Default style
Added robot test to check that the homepage renderers without error in Plone Classic Theme style.
(Note: I could not trigger an error in other, simpler tests.)

Test and merge this in combination with https://github.com/plone/plonetheme.classic/pull/8
Merge that one first.


mauritsvanrees:
My PR for plonetheme.classic was merged, so I merge this one too.

PR #2994, opened 2019-12-03
Correctly fire events when user auto login after the password has been reset

### LINKED ISSUE #2993: Login user after password reset notify events with Anonymous User
## BUG/PROBLEM REPORT

The option **Login user after password reset** allows a user that just performed a password reset to automatically log in after the .process is complete.

One of the two events (**UserLoggedInEvent**, **UserInitialLoginInEvent**) is triggered by this process, but with the Anonymous User instead of the user that just performed the password request.

### What I did:

- Newly created Plone site (no addons)
- **Login user after password reset** selected on */@@security-controlpanel*
- Create a new user
- Request a reset user password
- Follow the generated link

### What I expect to happen:

**UserLoggedInEvent** or **UserInitialLoginInEvent** should be triggered with the newly logged in user.

### What actually happened:

**UserLoggedInEvent** and **UserInitialLoginInEvent** are triggrered with **<SpecialUser 'Anonymous User'>**.

* Plone 5.2
* No addons

ericof:
Fix #2993

### DIFF
--- a/Products/CMFPlone/browser/login/password_reset.py
+++ b/Products/CMFPlone/browser/login/password_reset.py
@@ -10,6 +10,7 @@ from Products.CMFPlone.interfaces import IPasswordResetToolView
 from Products.CMFPlone.interfaces.controlpanel import IMailSchema
 from Products.CMFPlone.PasswordResetTool import ExpiredRequestError
 from Products.CMFPlone.PasswordResetTool import InvalidRequestError
+from Products.CMFPlone.RegistrationTool import get_member_by_login_name
 from Products.CMFPlone.utils import safe_unicode
 from Products.CMFPlone.utils import safeToInt
 from Products.Five import BrowserView
@@ -89,8 +90,8 @@ class PasswordResetView(BrowserView):
     subpath = None
 
     def _auto_login(self, userid, password):
-        aclu = getToolByName(self.context, 'acl_users')
-        membership_tool = getToolByName(self.context, 'portal_membership')
+        context = self.context
+        aclu = getToolByName(context, 'acl_users')
         for name, plugin in aclu.plugins.listPlugins(ICredentialsUpdatePlugin):
             plugin.updateCredentials(
                 self.request,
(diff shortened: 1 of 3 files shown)


ericof:

ericof:

Products/CMFPlone/browser/login/password_reset.py
```diff
@@ -90,14 +90,15 @@ class PasswordResetView(BrowserView):
...
                 self.request,
                 self.request.response,
                 userid,
                 password
             )
-        user = getSecurityManager().getUser()
+        user = membership_tool.getMemberById(userid).getUser()
         login_time = user.getProperty('login_time', None)
         if login_time is None:
             notify(UserInitialLoginInEvent(user))
```
jensens: If the notify still needed or does `getMemberById` takes care of tit as well?

ericof: @jensens I think **notify** is still needed here because **getMemberById** -- or the **updateCredentials** above -- is not authenticating the user.

The original code would always, in a default Plone installation, return an Anonymous User,  and that is why we use the membership tool to get a user object for that userid.

ericof:

Products/CMFPlone/browser/login/password_reset.py
ale-rt: Not related to this PR, but it feels strange to me that a call to `getSecurityManager().getUser()` in a method called `_auto_login` before firing the `UserLoggedInEvent` does not return the proper user :thinking:

ericof: @ale-rt I agree. My understanding of the whole situation is the following:

- Request to the password reset view is done without a valid auth cookie
- Zope/PAS authentication considers this user as anonymous
- After the password is reset we come to this method, still without authentication
- We call update credentials for all registered plugins (in the default Plone installation, this will send back a new cookie to the user)
- Current request still has no auth cookie (only the response will set it)
- getSecurityManager still believes this is an Anonymous user
- User receives a redirect and the new cookie
- Next request will see the user authenticated (as the new cookie is valid)

ale-rt: Yes, that is really plausible. I am wondering if this method should actually login the user so that the security manager is aware of that.
But maybe this has some other implications.
For me the PR is fine as a workaround (then mergeable) but the real issue is that the user at this point is not logged in.
Any other opinion?

ericof:
@ale-rt, @jensens Could you take a look at the latest change? Instead of using getMemberById, I used get_member_by_login_name -- the same way PasswordResetTool uses it.

ericof:

PR #3001, opened 2019-12-10
Update test: plone.app.discussion moderation

ksuess:
plone.app.discussion introduces an additional workflow. This change takes the additional workflow states into account.
see https://github.com/plone/plone.app.discussion/issues/164


ksuess:
wrong target branch

Issue #3007, opened 2019-12-31
Plone 5.2 advanced search breaks when searching with a date range


frapell:
This happens on a fresh Plone 5.2, on Python 3.

Steps to reproduce:
1. Go to http://localhost:8080/Plone/@@search
2. Under "Filter results" -> "New items since" choose "Last week"

A traceback is thrown in the terminal:
```
2019-12-30 23:21:49,512 ERROR   [Zope.SiteErrorLog:251][waitress] 1577758909.50604490.5644442555016539 http://localhost:8080/Plone/@@search
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 156, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 338, in publish_module
  Module ZPublisher.WSGIPublisher, line 256, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 62, in call_object
  Module zope.browserpage.simpleviewclass, line 41, in __call__
  Module Products.Five.browser.pagetemplatefile, line 126, in __call__
  Module Products.Five.browser.pagetemplatefile, line 61, in __call__
  Module zope.pagetemplate.pagetemplate, line 135, in pt_render
  Module Products.PageTemplates.engine, line 88, in __call__
  Module z3c.pt.pagetemplate, line 173, in render
  Module chameleon.zpt.template, line 306, in render
  Module chameleon.template, line 209, in render
  Module chameleon.utils, line 75, in raise_with_traceback
  Module chameleon.template, line 187, in render
  Module 951fe55309af08870bb8a0f61fbd6b21, line 3090, in render
  Module ca3ff2d3cceb4aebe5c885a0ce7e8708, line 582, in render_master
  Module ca3ff2d3cceb4aebe5c885a0ce7e8708, line 931, in render_content
  Module 951fe55309af08870bb8a0f61fbd6b21, line 2891, in __fill_main
  Module Products.CMFPlone.browser.search, line 81, in results
  Module Products.CMFPlone.CatalogTool, line 457, in searchResults
  Module Products.ZCatalog.ZCatalog, line 611, in searchResults
  Module Products.ZCatalog.Catalog, line 1091, in searchResults
  Module Products.ZCatalog.Catalog, line 634, in search
  Module Products.ZCatalog.Catalog, line 564, in _search_index
  Module Products.PluginIndexes.unindex, line 447, in query_index
[...]

PR #3012, opened 2020-01-03
Fix advanced search issue #3007 

frapell:


frapell:
So, it turns out that ZCatalog will specifically check if the index is a dict or not in https://github.com/zopefoundation/Products.ZCatalog/blob/e751eff05374e7f8b39d59b9b9e16e33fab47e11/src/Products/ZCatalog/query.py#L68

If it's not, it will just copy the index right away into the keys, causing the `TypeError: unhashable type: 'record'`
And this is because the `created` parameter is not a dict, but a `ZPublisher.HTTPRequest.record`

This fix will cast the `record` type as `dict`, which solves the issue.

This is a fix for issue https://github.com/plone/Products.CMFPlone/issues/3007

frapell:

ericof:

ale-rt:

jensens:

wesleybl:
@frapell in fact, this error occurs with any search that has the `query` key.

For example the url below also fails:

http://localhost:8080/Plone/@@search?end.query:record:list:date=2022-2-2+00%3A00%3A00&end.range:record=min

So the fix must be made to all parameters that have the `query` key, not just the `created` parameter.

wesleybl:
In fact, since dictionaries in url are turned into record, maybe the best solution is to test:

```python
if isinstance(param, dict) or isinstance(param, record):
```
in:

https://github.com/zopefoundation/Products.ZCatalog/blob/e751eff05374e7f8b39d59b9b9e16e33fab47e11/src/Products/ZCatalog/query.py#L68

See: https://github.com/plone/Products.CMFPlone/pull/3012#issuecomment-570630842

@frapell @ericof @ale-rt opinions?

frapell:
@wesleybl It's been 2 years since this fix, so I don't fully remember, and I don't know what would be the proper fix, however on a first glance it does look like it should be for the `@@search` view to properly convert `records` into `dicts`

Issue #3021, opened 2020-01-22
Integrate PloneHotfix20200121


mauritsvanrees:
Main issue for tracking that the [January 2020 Hotfix](https://plone.org/security/hotfix/20200121) is applied everywhere.

We could copy this into separate issues per Plone version if we want. And/or create an issue per package that needs a fix. Let's see. I am happy for others to jump in.

## Privilege escalation when plone.restapi is installed

- https://plone.org/security/hotfix/20200121/privilege-escalation-when-plone-restapi-is-installed
- Plone 5.2, 6.0
- [x] Needs fix in `plone.restapi`. Fixed in 6.2.1.

## SQL injection in DTML or in connection objects

- https://plone.org/security/hotfix/20200121/sql-injection-in-dtml-or-in-connection-objects
- All Plone versions. All Zope versions.
- [x] Needs fix in `DocumentTemplate`, `Products.PythonScripts`,  `Products.ZSQLMethods` (for ` Shared.DC.ZRDB.Connection`) [Update: done, coredev PRs created. And `PythonScripts needed no fix.]
- [x] The hotfix may break when using Postgres (for example double quotes ending up twice in the database), so we might need a version 1.1 of the hotfix. The main fix is in DocumentTemplate, and that has a good version of the code.

## XSS in the title field on plone 5.0 and higher

- https://plone.org/security/hotfix/20200121/xss-in-the-title-field-on-plone-5-0-and-higher
- Plone 5 and 6
- Needs fix in `plone.app.content` and `plone.app.layout`
- [x] `plone.app.content` 3.5.x: https://github.com/plone/plone.app.content/pull/195
- [x] `plone.app.content` master: https://github.com/plone/plone.app.content/pull/193
- [x] `plone.app.layout` only master: https://github.com/plone/plone.app.layout/pull/232

## Privilege escalation for overwriting content

- https://plone.org/security/hotfix/20200121/privilege-escalation-for-overwriting-content
- All Plone versions
- Needs fix in `plone.app.contenttypes`
- [x] 1.1.x: https://github.com/plone/plone.app.contenttypes/pull/546
- [x] 1.4.x: https://github.com/plone/plone.app.contenttypes/pull/545
- [x] master: https://github.com/plone/plone.app.contenttypes/pull/543

## An open redirection on the login form and possibly other places

- https://plone.org/security/hotfix/20200121/an-open-redirection-on-the-login-form-and-possibly-other-places
- All Plone versions
[...]

PR #3022, opened 2020-01-22
4.3.x - Password strength checks were not always checked

gotcha:
Include Hotfix20200121

see #3021


gotcha:
What is the usual practice when merging hotfixes, do we add tests ?

mauritsvanrees:
Yes. I have tests for this in the private hotfix repository. But if you or someone else creates a PR first, I can always update the PR with some tests. Or I add tests in a later PR.

PR #3023, opened 2020-01-22
4.3.x - Password strength checks were not always checked 

gotcha:
Include Hotfix20200121

see #3021

Better branch name; replaces #3022


gotcha:
wrong base branch

PR #3024, opened 2020-01-22
4.3.x Hotfix20200121 isURLInPortal

gotcha:
The isURLInPortal check that is done to avoid linking to an external
site could be tricked into accepting malicious links.

See #3021


gotcha:
@mauritsvanrees Should I squash commits before merging ?

mauritsvanrees:
Yes, squashing seems like a good idea. The 'squash and merge' option in the GitHub UI could be used for that.

mauritsvanrees:

PR #3032, opened 2020-02-04
plone.app.discussion extends the review workflow (Plone 5.2)

ksuess:
plone.app.discussion extends the review workflow for moderation of comments. This change takes the additional workflow states into account.
see https://github.com/plone/plone.app.discussion/pull/166


ale-rt:
@ksuess This is not needed anymore because #3036 #3038 makes the tests independent of the workflows of the other packages

PR #3033, opened 2020-02-04
plone.app.discussion extends the review workflow (Plone 5.1)

ksuess:
plone.app.discussion extends the review workflow for moderation of comments. This change takes the additional workflow states into account.
see plone/plone.app.discussion#166


ale-rt:
This is not needed anymore: The work on plone.app.discussion should not touch at all Plone 5.1.
Anyway #3037 does the same in a way that is not affected by changes in plone.app.discussions workflows.

Issue #2812, opened 2019-03-20
Drop Python 2 Support for Plone 6


tisto:
## PLIP (Plone Improvement Proposal)

When Plone 6 will be out, Python 2 will be officially unsupported. It does not make any sense if we ship a major Plone version with Python 2. With the current schedule, we will ship Plone 6 one year after Python 2 is unsupported.

## Responsible Persons

Timo Stollenwerk

## Abstract

Drop support for Python 2 on the next major Plone version (Plone 6).

## Motivation

Maintaining two Python versions has major downsides. We have lots of conditional imports and try/exept, if/else statements all over the place that are hard to maintain and debug. We should focus on areas where it makes more sense.

## Assumptions

Python 2 will not be supported after Jan. 2020. Plone 6 will be out long after Python 2 will be out of security fixes.

## Proposal & Implementation

Officially drop Python 2 support for Plone 6 and cut Python 3 only branches for Plone 6.

## Deliverables

- Do not set up Jenkins jobs for Python 2 for Plone 6
- Cut branches on all individual packages that drop Python 3 support

## Risks

Branching too early might lead to the overhead of backporting lots of fixes. We shouldn't do this too early.

## Participants

Timo Stollenwerk

tisto:
@plone/framework-team please discuss and review. 

@pbauer wanna second? :)

ale-rt:
IMHO this is so obvious that should not even be a PLIP :)

sneridagh:
Cleaning all the branching and the six workarounds will be hard, and should be done. Having it as a PLIP is important for not overlook it, I think.

jensens:
Communication wise it is good to have a PLIP, even if its the next obvious thing.

ebrehault:
The FWT approved this PLIP

fredvd:
As the possibld risks list: actively removing Python 2 compatibility means creating a new major release of every subpackage Plone depends on first and blocking any sharing of stable not often changing package versions between Plone 5.2 and Plone 6.

This will negatively impact the support window for Plone 5.2. and increase the load in the release team to support the previous Plone version (backports).

Is this really such a necessity with huge benefits that it has to be forced/policied on Plone 6 and can’t wait until Plone 7 ?

Juronski:
Would like to point to efforts for making Plone run on Tauthon (2.8.1). Tauthon is a Python 2.7 fork.  
[...]

PR #3041, opened 2020-02-11
make setup.py reflect to be Plone 6 and Py3 only

jensens:
part of #2812


mauritsvanrees:
The `Framework :: Plone :: 6.0` classifier does not exist in https://pypi.org/classifiers/ though. That might also mean an upload of this package to PyPI would fail (not sure).
I will request a new classifier, which we [have done before](https://github.com/pypa/warehouse/issues?q=is%3Aissue+label%3A%22classifier+request%22+plone).

Issue #3042, opened 2020-02-11
mail_password_form throws error when called without user name


achimwilde:
### What I did:

- Have a fresh current default Plone 5.1.x instance with no add-ons.
- Click "Login" and then "we can send you a new one" to get to /mail_password_form
- Click "Start password reset", without entering a valid user name first (empty field or wrong user name)

### What I expect to happen:

Get a message "The username you entered could not be found." and below the previous form again.

### What actually happened:

Got a message "The username you entered could not be found." and an error message ("We’re sorry, but there seems to be an error…")

Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module Products.CMFCore.FSPythonScript, line 127, in __call__
  Module Shared.DC.Scripts.Bindings, line 322, in __call__
  Module Shared.DC.Scripts.Bindings, line 359, in _bindAndExec
  Module Products.PythonScripts.PythonScript, line 344, in _exec
  Module script, line 16, in mail_password
   - <FSPythonScript at /Plone/mail_password>
   - Line 16
AttributeError: 'RequestContainer' object has no attribute 'mail_password_form'

Already reported here:
https://community.plone.org/t/attributeerror-requestcontainer-object-has-no-attribute-mail-password-form/8587

Note:
CMFPlone 5.2.x is not affected because the script responsible for the issue got replaced completely.

achimwilde:
Analysis: The line "response = context.mail_password_form()" doesn't work because the context object (PloneSite) doesn't have this attribute.

Proposed solution: Replace the method call with a redirect, similar to the CMFPlone 5.2.x way:
https://github.com/plone/Products.CMFPlone/blob/5.2.x/Products/CMFPlone/browser/login/mail_password.py

PR #3046, opened 2020-02-12
Fix mail_password_form throwing error when submitted without entering…

achimwilde:
fixing issue #3042


achimwilde:

Issue #3076, opened 2020-04-08
Add markdown settings from portal transforms to controlpanel


thomasmassmann:
Markdown settings in portal_transforms are currently not persisted (see https://github.com/plone/Products.PortalTransforms/issues/29).

The changes provided in the PR (https://github.com/plone/Products.PortalTransforms/pull/30, the first commit in there) are partially fixing the problem. Changes are now persisted, but to take effect when using the markdown transform module, a restart is currently required.

It would be better to have the settings in the control panel, specifically in the Markup Controlpanel. The PR (https://github.com/plone/Products.PortalTransforms/pull/30) is currenlty adding settings to the TinyMCE registry records which is not working (due to a missing field definition).

I therefor propose to add the markdown extension settings to the Markup Controlpanel. The transform module can then take those new settings from the registry. A notification like its done for the safe_html settings in portal_transforms should be added as well.

Issue #2991, opened 2019-11-26
text/x-web-textile not working


krissik:
### What I did:

I added a Page (on http://demo.plone.org) and selected text/x-web-textile as markup. When I add textile text to it the markup is shown while viewing this page. I expected formatted text.
Am I missing a configuration option?

### What version of Plone/ Addons I am using:
Plone 5.2.0 (5207)
CMF 2.4.0
Zope 4.1.1
Python 3.7.2 (default, Mar 11 2019, 14:41:38) [GCC 5.4.0 20160609]

jensens:
confirmed for 5.2.10 and 6.0.2

PR #3077, opened 2020-04-08
Add markdown extension settings to markup control panel.

thomasmassmann:
Adds enabled markdown extensions for markup portal transform module to markup controlpanel.

The required changes in Products.PortalTransforms will be applied in https://github.com/plone/Products.PortalTransforms/pull/30.

Needs PR:
https://github.com/plone/plone.app.upgrade/pull/229
https://github.com/plone/Products.PortalTransforms/pull/30

### DIFF
--- a/Products/CMFPlone/controlpanel/tests/test_controlpanel_browser_markup.py
+++ b/Products/CMFPlone/controlpanel/tests/test_controlpanel_browser_markup.py
@@ -72,3 +72,32 @@ class MarkupControlPanelFunctionalTest(unittest.TestCase):
         settings = registry.forInterface(IMarkupSchema, prefix='plone')
         self.assertEqual(settings.allowed_types,
                          ('text/html', 'text/x-web-textile'))
+
+    def test_markdown_extensions(self):
+        self.browser.open(
+            "%s/@@markup-controlpanel" % self.portal_url)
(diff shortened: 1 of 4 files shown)


thomasmassmann:

jensens:
This also need an upgrade step in plone.app.upgrade.

thomasmassmann:
> eventually I'd change the default in `allowed_types` from `default=('text/html', 'text/x-web-textile')` to `default=('text/html', 'text/x-web-textile', 'text/x-web-markdown')`. Is textile even used or working at all (#2991)?

I would do this once there is a markdown editor available to be used. I'm currently evaluating https://ui.toast.com/tui-editor/, but the plugins have to match (what you see in the preview and what Python-Markdown is rendering).

thet:
I think the point of markdown is not to need to use an editor.

However, having an editor available would be nice. But that's a lot more effort.

jensens:
I would add plain markdown to the list as @thet suggested. And merge this PR.
All further changes can go in a second PR.

OT here: Btw. I never used Textile. Removing this in favour of markdown sounds reasonable, given textile is broken anyway. OTOH this is a breaking change, so maybe we should do the removal for Plone 6?

thomasmassmann:
Closes #3076.

PR #3078, opened 2020-04-10
Fix typo

fredvd:
causes double translation text. Same message is on line 350


ale-rt:

PR #3091, opened 2020-04-27
zope.app.locales is no longer needed, all zope.app.* was removed

jensens:
cleanup


jensens:

vincentfretin:
FYI in `zope/schema/interfaces.py` there is some messages that are translated in zope.app.locales, like min and max for number fields ("Start of the range", "End of the range (including the value itself)"), and also "Required" and "Tells whether a field requires its value to exist." that show up in dexterity schema editor if I remember. I know those ones were used in Plone 4.3.x because I overrode the translations in a project.

jensens:
@vincentfretin what do yout think depending plone.app.locales on zope.app.locales and removing the dependency from Products.CMFPlone?

vincentfretin:
Yes you can do that.

PR #3097, opened 2020-05-19
Catch more errors in generating scripts listing when bundle resources…

fredvd:
… cannot be found.

### DIFF
--- a/Products/CMFPlone/resources/browser/combine.py
+++ b/Products/CMFPlone/resources/browser/combine.py
@@ -49,7 +49,7 @@ def get_resource(context, path):
 
     try:
         resource = context.unrestrictedTraverse(path)
-    except NotFound:
+    except (NotFound, AttributeError):
         logger.warning(u'Could not find resource {0}. You may have to create it first.'.format(path))  # noqa
         return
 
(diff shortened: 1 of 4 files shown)


fredvd:
Test is not complete, it tests now if the combine bundle is still working, but the missing KeyError in Bundle class its realpath method is not triggered with the current test.

The traversing to a broken path/resource triggers either a KeyError, AttributeError or a NotFound, I think it depends on an already missing path or ++resource++ (I'm using notfound in the test) or a missing file 'in' the ++resource++does.exist/missing-file.js . 

Are there any other existing ++resource++ locations I can us in a test?

fredvd:

mauritsvanrees:

fredvd:
> LGTM, although I am not sure why in one case you except AttributeError and in another KeyError. Maybe do both. But maybe you have a reason. And this already seems a helpful improvement.

I only used these to be on the conservative side with swallowing Exceptions. I ran into this problem locally and pdb'ed through the code to see where which exception was raised and only added the Exceptions which were triggered in this case. 

My first try was the compile code, and when I fixed that the bundle process threw this.   This PR at solves part of the frustration of making one typo in resource registry config and breaking the whole Plone site, but I'm not an expert here.

mauritsvanrees:
Well, seems an improvement to me. I will merge.

Issue #2535, opened 2018-09-28
TinyMCE default styles incompatible with HTML filtering


zopyx:
Usability bug in Plone 5.1.2:

- create a new document, add some text and select "strikethrough" as character format -> text show as strikethrough inside TinyMCE
- "Save" -> text display as normal text

jensens:
I close the issue, because it addresses a Plone version that is no longer supported. 
If you think this is wrong please reopen the issue and assign a matching milestone.

And: probably solved already.

Issue #2482, opened 2018-07-31
Default TinyMCE /HTML filtering settings incompatible with HTML list classes


zopyx:
Plone 5.1 does not define any of the CSS classes used for lists in TinyMCE as allowed CSS classes in HTML filtering. They are stripped already client side.

See

https://community.plone.org/t/ordered-list-in-other-formats-beside-numbers-are-not-showing-up-properly/6953/6?u=zopyx

jensens:
Closing as outdated (8 years old). About TinyMCE 4 HTML filtering stripping list classes. TinyMCE integration and filtering stack replaced in Plone 6.
If you think this issue is still relevant please reopen it. Thank you for the original report!

PR #3102, opened 2020-05-26
Update default HTMLFilter values to fix TinyMCE styling features [5.1]

### LINKED ISSUE #2329: Tinymce: unable to set an alignment style
I want to continue a discussion started on the [forum](https://community.plone.org/t/tinymce-unable-to-set-an-alignment-style/5480) and on irc today.

The problem is that when you want to align some text, Tiny add an inline style to selected tag, but on save, Plone filters remove that attributes (styles) because they are blacklisted.

Going deeper in this feature, i’ve found that from 5.0 to 5.1 the html filters control panel has changed, and exactly [this commit](https://github.com/plone/Products.CMFPlone/commit/d975d1fdcbd050052401b4e11104fdfe2679b67b#diff-fa6b670554e72078393552c295dc150dL406) introduced the “problem”:
in Plone 5.0 there was a “_style_whitelist_” field with a set of allowed styles (also text-align), but in the new control panel we have less fields and we can only enable the whole “styles” attribute.

@MrTango did that change ;)

Obviously i don’t know all the background about this decision, but can this be some kind of regression?

I see two possible solutions:

- add the config in tiny control panel + some css classes in the theme
- revert attributes filter and controlpanel entry and allow text-align styles.

For a customer we did the first one without problems, as gp54321 (sorry, i don't know your github account) suggested.
We only needed to update tinymce config, and add a set of css classes like this:

     "alignleft": {
         "classes": "tinyAlignLeft",
         "block": "div"
      },
      "aligncenter": {
          "classes": "tinyAlignCenter",
          "block": "div"
      },
      "alignright": {
          "classes": "tinyAlignRight",
          "block": "div"
      }

And add these classes to our theme to have some text-align styles (i haven't found anything similar in barceloneta).

petschki:
fixes #2329 #2482  #2535

### DIFF
--- a/Products/CMFPlone/interfaces/controlpanel.py
+++ b/Products/CMFPlone/interfaces/controlpanel.py
@@ -407,6 +407,7 @@ class IFilterSchema(Interface):
             u'head',
             u'header',
             u'hgroup',
+            u'hr',
             u'html',
             u'i',
             u'iframe',
@@ -458,7 +459,7 @@ class IFilterSchema(Interface):
     custom_attributes = schema.List(
         title=_(u'Custom attributes'),
         description=_(u'These attributes are additionally allowed.'),
-        default=[],
+        default=['style'],
         value_type=schema.TextLine(),
         missing_value=[],
         required=False)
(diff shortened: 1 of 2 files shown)


petschki:

PR #3089, opened 2020-04-23
Add custom.css resource after diazo bundle

MrTango:
Insert virtual custom.css bundle after Diazo bundle.

PR #3117, opened 2020-06-11
Add custom.css resource after diazo bundle [5.2]

mauritsvanrees:
Backport of PR #3089 from 6.0/master into 5.2, as [approved by the Framework Team](https://github.com/plone/Products.CMFPlone/pull/3089#issuecomment-634032215).

### DIFF
--- a/Products/CMFPlone/resources/browser/styles.py
+++ b/Products/CMFPlone/resources/browser/styles.py
@@ -1,9 +1,12 @@
 # -*- coding: utf-8 -*-
 from plone.app.layout.viewlets.common import ViewletBase
+from plone.app.theming.interfaces import IThemeSettings
+from plone.registry.interfaces import IRegistry
 from Products.CMFPlone.resources.browser.cook import cookWhenChangingSettings
 from Products.CMFPlone.resources.browser.resource import ResourceBase
 from Products.CMFPlone.utils import get_top_request
 from six.moves.urllib import parse
+from zope.component import getUtility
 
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

Products/CMFPlone/resources/browser/styles.py
```diff
@@ -143,6 +152,18 @@ def styles(self):
...
+        custom_css = {
+            'rel': 'stylesheet',
+            'conditionalcomment': '',
+            'src': "{0}/custom.css?timestamp={1}".format(
+                self.site_url,
+                self.custom_css_timestamp,
+            ),
+            'bundle': 'custom-css'
+        }
+        result.append(custom_css)
```
ale-rt, line 166: I still think that this should be added only if not empty (see https://github.com/plone/Products.CMFPlone/pull/3089#discussion_r423717288) but we can do this refinement later.

ale-rt:
Does this also need a plone.app.upgrade upgrade step?

mauritsvanrees:
No. The upgrade step is in https://github.com/plone/plone.app.theming/pull/184 and it is applied automatically when upgrading Plone, because `plone.app.theming` is in the list of core packages for which we look for upgrades.

Products/CMFPlone/resources/browser/styles.py
mauritsvanrees: I agree.

PR #3142, opened 2020-07-04
Unordered folders make next/prev navigation at plone.restapi don't work

rodfersou:
related to: https://github.com/plone/plone.app.contenttypes/pull/553


rodfersou:
looks like Jenkins is not working now.. will try again other day 😄

rodfersou:
related to https://github.com/plone/plone.app.contenttypes/pull/553

Issue #3141, opened 2020-07-02
traversing macro doesn't works, testing upgrade to Plone 5.2.2 (pending)


mamico:
### What I did:

Upgraded my buildout to Plone 5.2.2 (pending release)

### What I expect to happen:

It is a minor change, everything should be fine!

### What actually happened:

Some custom views are broken, Traceback:
```
...
Module zope.pagetemplate.engine, line 85, in __call__
Module zope.traversing.adapters, line 158, in traversePathElement
   - __traceback_info__: (<Document at /Plone/page>, 'custom_macros')
Module zope.traversing.adapters, line 46, in traverse
   - __traceback_info__: (<Document at /Plone/page>, 'custom_macros', ['foo'])
zope.location.interfaces.LocationError: zope.location.interfaces.LocationError: (<Document at /Plone/page>, 'custom_macros')

```
This is an excerpt from my code:

document.pt:
```
...
<div metal:use-macro="context/custom_macros/foo" />
...

```

configure.zcml:

```
  <browser:page
    for="*"
    name="custom_macros"
    template="macros.pt"
    permission="zope2.View"
    />
```

macros.pt:
```
<html xmlns="http://www.w3.org/1999/xhtml" 
      xml:lang="en" 
      xmlns:tal="http://xml.zope.org/namespaces/tal" 
      xmlns:metal="http://xml.zope.org/namespaces/metal">

  <metal:foo define-macro="foo">
    <div>macro</div>
  </metal:foo>
...
</html>
```

Differences between compiled templates in Chamelon cache:

Plone 5.2.1 / Zope 4.2.1:
```
from Products.PageTemplates.expression import TrustedBoboAwareZopeTraverse as _TrustedBoboAwareZopeTraverse
_static_140708896619584 = _TrustedBoboAwareZopeTraverse()
...
                        # <Value 'context/custom_macros/foo' (17:34)> -> __macro
                        __token = 758
                        __macro = _static_140708896619584(getitem('context'), econtext, True, ('custom_macros', 'foo', ))

```

Plone 5.2.2 / Zope 4.4.4:
```
from Products.PageTemplates.engine import _compile_zt_expr as __compile_zt_expr
_static_140007861886448 = __compile_zt_expr
...
                       # <Value 'context/custom_macros/foo' (17:34)> -> __macro
                        __token = 758
                        __macro = _static_140007861886448('path', 'context/custom_macros/foo', econtext=econtext)(_static_140007861886208(econtext))

``` 

[...]

PR #3149, opened 2020-07-20
Return a Zope aware engine for page templates based on zope.pagetemplate

mauritsvanrees:
Fixes possible problems with such templates, for example z3c.form ones, with Zope 4.4 and higher.
Most templates are based on `Products.PageTemplates`, which is working fine.

See https://github.com/plone/Products.CMFPlone/issues/3141

cc @d-maurer.

### DIFF
--- a/Products/CMFPlone/configure.zcml
+++ b/Products/CMFPlone/configure.zcml
@@ -131,4 +131,9 @@
       provides="plone.scale.interfaces.IScaledImageQuality"
       />
 
+  <adapter
+      for="zope.pagetemplate.engine.ZopeBaseEngine"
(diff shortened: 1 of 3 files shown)


mauritsvanrees:

ale-rt:

mauritsvanrees:
> LGTM.
> I was thinking about adding a `zcml:condition="Products.PageTemplates.interfaces.IZopeAwareEngine"` but at the end I think the condition will always be met in all the practical cases.

I was thinking the same and tried it yesterday, but I think the condition always actually *failed*. It seems the zcml condition can only contain packages, or maybe modules, but not individual classes or interfaces.

PR #3153, opened 2020-07-23
Redirection view: factor our navigation root editing to a method. [5.2]

mauritsvanrees:
This makes it possible to override this method to not do anything.

With the original Products.RedirectionTool in Plone 4.3, nothing was done with the navigation root.
So when you are in `/nav-root/page` and add an alternative url `/fnord`, it was stored as `/fnord`.
In current Plone 5.2, it is stored as `/nav-root/fnord`.
Makes sense to me, but some want the old behavior back.
Now they can, by overriding the new `edit_for_navigation_root` method to simply return the redirection unchanged.


mauritsvanrees:

mauritsvanrees:
Thanks for the review. I will merge as is, and then update the news snippet after merge.

Issue #3154, opened 2020-07-24
Unauthorized on member.canSetPassword condition in configlet


mauritsvanrees:
[edited after I accidentally pressed Enter far too soon]

I tried an upgrade from Plone 5.2.1 to 5.2.2 pending. This is in a project with six Plone Sites in one Zope. In only *one* of the Plone Sites do I get an error when going to the overview control panel:

```
2020-07-24 02:14:09,059 ERROR   [Zope.SiteErrorLog:252][waitress-3] 1595549649.060.0403072128614 http://localhost:8080/nl/@@overview-controlpanel
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 162, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 359, in publish_module
  Module ZPublisher.WSGIPublisher, line 262, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 63, in call_object
  Module Products.CMFPlone.controlpanel.browser.overview, line 34, in __call__
  Module Products.Five.browser.pagetemplatefile, line 126, in __call__
  Module Products.Five.browser.pagetemplatefile, line 61, in __call__
  Module zope.pagetemplate.pagetemplate, line 135, in pt_render
  Module Products.PageTemplates.engine, line 322, in __call__
  Module z3c.pt.pagetemplate, line 176, in render
  Module chameleon.zpt.template, line 299, in render
  Module chameleon.template, line 214, in render
  Module chameleon.template, line 192, in render
  Module a5e5b9a6b05600f91d918f35fa6b8fdc, line 1021, in render
  Module 859b00e113ddbff3769ebb8e467855e2, line 294, in render_master
  Module a1fa38b5c0524a1ea4363369646a37be, line 1182, in render_master
  Module 859b00e113ddbff3769ebb8e467855e2, line 274, in __fill_content
  Module a1fa38b5c0524a1ea4363369646a37be, line 500, in render_content
  Module 859b00e113ddbff3769ebb8e467855e2, line 265, in __fill_main
  Module a5e5b9a6b05600f91d918f35fa6b8fdc, line 469, in __fill_prefs_configlet_main
  Module zope.tales.pythonexpr, line 73, in __call__
   - __traceback_info__: (view.sublists(category.get('id')))
  Module <string>, line 1, in <module>
  Module Products.CMFPlone.controlpanel.browser.overview, line 137, in sublists
[...]

PR #3155, opened 2020-07-24
Control panel configlets: first check visibility, then check condition.

mauritsvanrees:
Visibility is cheaper to check.
Also fixes https://github.com/plone/Products.CMFPlone/issues/3154.


mauritsvanrees:

Issue #3150, opened 2020-07-21
Split isURLInPortal off into Products.isurlinportal


mauritsvanrees:
This is an idea from me, briefly discussed in the Plone Security Team. We can discuss it openly though.

I have copied the `isURLInPortal` method from the `CMFPlone` `URLTool` to a new package:
https://github.com/plone/Products.isurlinportal
Once Travis is finished and green (it has not started yet), I want to release version 1.0.0 of this.

Code of 1.0.0 is the same as current Plone 4.3.19, 5.1.6, 5.2.1, with the January 2020 hotfix applied, plus compatibility fixes.

I think the [readme](https://github.com/plone/Products.isurlinportal) explains everything I want to say.
In short: most Plone security hotfixes contain updated code for the `isURLInPortal` method. This part of the hotfix is usually not severe. When this method is in a separate package, we can simply release a new version of this package. When severe enough, we can still do an announcement.

Just add the package to the eggs in your buildout config, use the latest version pin, rerun the buildout, restart the instance and you are done. I guess it should even work in Plone 4.2 and in Python 2.6, but this is not tested and not supported.

Nothing needs to happen in Plone itself for this to happen. But in future Plone releases we should include the package automatically. To make this work:

- In `Products.CMFPlone` depend on the new package.
- In `Products.CMFPlone` remove the current method and make sure the new one is always loaded. This is actually only needed to make the tests use the new package, because the Products code is not loaded automatically in the tests. I have a branch for this locally. (I thought about making sure the package is loaded in `plone.app.testing`, but this does not help, because the current `testURLTool.py` is a plain unittest without layers.)
- In `buildout.coredev` add `Products.isurlinportal` in various configs. I have this ready too.

I want to do this in Plone 5.2.2 already, even though it has pending for a while.
I would do the same for 4.3 and 5.1 before making releases of those.

mauritsvanrees:
Done in 4.3, 5.1, 5.2, 6.0.

PR #3156, opened 2020-07-27
Depend on new package Products.isurlinportal. [4.3]

mauritsvanrees:
This contains the `isURLInPortal` method that was split off from our `URLTool`.
See https://github.com/plone/Products.CMFPlone/issues/3150.

We depend on a version less than 2.0.0.
This is in the expectancy that such a breaking version will be released at some point,
probably when dropping support for Python 2 or for Zope 2.

Backported from 5.2.


mauritsvanrees:

PR #3167, opened 2020-09-03
Fixed deprecation warning for zope.site.hooks.

mauritsvanrees:
Note: we still have `zope.site` in our `setup.py` `install_requires`.
We should probably keep it there in 5.2, to keep backwards compatibility for other packages that may not have it in their own install_requires.

Also, I am not sure if we are indirectly depending on other code in `zope.site`, or if the entire package is deprecated.


mauritsvanrees:

PR #3171, opened 2020-09-09
Fixed invalid escape sequences in regular expressions.

mauritsvanrees:
Note: I am less sure about the change in the `UnicodeSplitter`. Note that in Python 2, you can use `ur'abc'`, but on Python 3 this is not allowed. But we don't seem to the the unicode character here. At least Plone starts up fine in both Python 2.7 and 3.7 after this change. And `bin/test -s Products.CMFPlone -m testUnicodeSplitter` passes for both. So should be okay.

Also note that in Python 3 you get a DeprecationWarning for invalid escape sequences, even when they are not *really* invalid, but they just *might* be invalid and Python cannot tell for sure. In a future Python, this may become a SyntaxError.

Also note that to see these deprecation warnings, you may need to remove the `*.pyc` files (or `__pycache__` directories actually).


mauritsvanrees:

gforcada (approved):

PR #3175, opened 2020-09-16
PloneBatch: define __bool__ as copy of __nonzero__.

mauritsvanrees:
Python 2 calls `__nonzero__`, Python 3 calls `__bool__`.
Since `__bool__` was not defined, Python 3 called `len(batch)` instead, leading to these warnings when a template has code like `tal:condition="batch"`:

```
Products/PageTemplates/Expressions.py:253: DeprecationWarning: Using len() is deprecated.
Use the `length` attribute for the size of the current page, which is what we return now.
Use the `sequence_length` attribute for the size of the entire sequence. 
  return bool(value)
```

(I now think we should undo this deprecation, but that is something for a different issue.)

### DIFF
--- a/Products/CMFPlone/PloneBatch.py
+++ b/Products/CMFPlone/PloneBatch.py
@@ -27,10 +27,13 @@ class Batch(QuantumBatch):
                           'Use the `sequence_length` attribute for the size of the '
                           'entire sequence. '))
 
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

ale-rt (approved):

PR #3185, opened 2020-09-26
Pyupgrade

jensens:
Applied ``find . -name "*.py" |grep -v skins|xargs pyupgrade --py36-plus --py3-only``.
Lets see what Jenkins says.


pbauer:
`--py3-only` is obsolete here. `--py36-plus` is enough and will give us f-strings :)

mauritsvanrees:
I would also say use `--py36-plus`. Let's have all changes in one commit.

jensens:
Doesn't `--py36-plus` implies `--py3-only` already?

mauritsvanrees:
`--py36-plus` implies `--py3-only`, but not the other way around.

jensens:
to merge or not to merge, that's here the question....

pbauer:
I'm generally +1 but I want it with f-strings. When I run the command `find . -name "*.py" |grep -v skins|xargs pyupgrade --py36-plus` I get at least some some f-strings, e.g. in https://github.com/plone/Products.CMFPlone/pull/3185/files#diff-1582a68636fd8187555ca419b38fce6cL306 your result is `'{}/{}'.format(category, id)` whereas mine is `f'{category}/{id}'`, which I prefer.
Can you please do again the same without `--py3-only`?

jensens:

Issue #3162, opened 2020-08-05
Add Bootstrap Icons


santonelli:
# Abstract

This issue belongs to #2967 (Change Templates to Bootstrap 4). 

Since we have to touch a lot of icons anyway, we decided to use [Bootstrap Icons](https://icons.getbootstrap.com/) in the upcoming theme. Since the current icon handling is complex and icons are difficult to customize we come up with this proposal.

## TL;DR

Icons are referenced in Plone only by its semantic name. A resolver then builds an URL or tag for consumption in templates or icon expressions. It looks up the name in the Plone registry, where an resource URL is stored for each name. 

The bootstrap icon set is used by default. Other icon sets are provided as add-ons. Those or customization in projects are done by remapping the names to other resources. 

Inline SVG is the preferred icon format, for legacy PNG is possible too. Inlining SVG at all places is not scope of this proposal.

# Motivation

A full set of useful, clean and modern icons should be available in Plone by default. 

Icon names should reflect the semantics for what they are used, not what they show.

The current way to register icons is not simple. Definitions of icons are in code, LESS/CSS and configuration (icon expression). In future there should be one place to register icons (registry) and one way to use them.

# Assumptions

- Plone 6
- Bootstrap 4 based Theme

# Proposal & Implementation

## Semantic Icons

Icons should be provided by a name to be resolved to the actual icon. An icon should be accessed by name only, independent of the actual icon-set used. Thus it will be way easier to customize an icon theme or provide a new one.

Icons names are semantic, so an icon may be named: i.e. `plone-copy`, `plone-paste`, `plone-contenttype-image`,  and a resolver then delivers the configured icon. Icon mapping happens in the registry. The same icon will be registered under a more generic name, like `generic-image`.

## Icon Registry

Wherever icons are used they are referenced by semantic name. The name is then resolved either to the resource URL or the HTML-tag.

We use the `plone.registry` for resolving. This works similar as already used in other places. Using a prefix for each icons `plone-ICONNAME`.  The value is the resource URL.

Thus any icon can be customized by changing its Plone registry record: only the resource URL of the name need to be updated.

## Default Full Icon Set

By default we provide the [Bootstrap Icons](https://icons.getbootstrap.com/). The full set is registered using their bootstrap name. Those names usages is disallowed in Plone Core. They are meant to be used in customizations only.

[...]

PR #3192, opened 2020-10-09
Add Iconresolver

santonelli:


Products/CMFPlone/browser/iconresolver.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, context, name):
+        self.context = context
+        self.name = name
+
+    def __call__(self):
+        registry = getUtility(IRegistry)
+        prefix = 'plone.staticresources.icon.'
+        icon = prefix + self.name
+        if icon in registry:
+            return registry[icon]
```
jensens, line 27: If not possible to lookup a LocationError would be nice
```suggestion
            return registry[icon]
        raise LocationError(self.name)
```

Products/CMFPlone/browser/iconresolver.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, context, name):
+        self.context = context
+        self.name = name
+
+    def __call__(self):
+        registry = getUtility(IRegistry)
+        prefix = 'plone.staticresources.icon.'
+        icon = prefix + self.name
+        if icon in registry:
```
jensens, line 26:
```suggestion


PREFIX = 'plone.staticresources.icon.'

@implementer(IPublishTraverse)
class IconBaseResolver:

    def __init__(self, context, name):
        self.context = context
        self.name = name

    def __call__(self):
        registry = getUtility(IRegistry)
        icon = PREFIX + self.name
        if icon in registry:
```

ale-rt: I would set the `PREFIX` as a class variable:
```
class IconBaseResolver:
     prefix = 'plone.staticresources.icon.'
```
IMO it is easier to override.

jensens: ok, good point

jensens:
Some basic tests are missing, right?

Products/CMFPlone/browser/iconresolver.py
```diff
@@ -0,0 +1,85 @@
...
+                alt_tag = self.name,
+            )            
+        return tag
+
+
+@implementer(ITraversable)
+@adapter(ISiteRoot, Interface)
+class IconBaseTraverser:
+
+    resolver = None
```
ale-rt, line 66: Do we really need another component here? How do you plan to use this one? Before merging I would like to see some documentation draft to understand the big picture.

Products/CMFPlone/browser/iconresolver.py
```diff
@@ -0,0 +1,85 @@
...
+        if icon in registry:
+            return registry[icon]
+        raise LocationError(self.name)
+
+
+class IconURLResolver(IconBaseResolver):
+
+    def __call__(self):
+        value = super().__call__()
+        url = getSite().absolute_url() + '/' + value
```
ale-rt, line 36: I think here you want the navigation root URL from plone portal state: https://github.com/plone/plone.app.layout/blob/5f597e41c413c1584a25f01347ead48da52d77c4/plone/app/layout/globals/portal.py#L67
Example: https://github.com/plone/Products.CMFPlone/blob/c3c7ec9ea52fde4d1e868f7df703a80cc8325c0a/Products/CMFPlone/browser/ploneview.py#L156-L160

Products/CMFPlone/browser/iconresolver.py
santonelli: The big picture is described in the issue #3162. This PR is one of two possible solutions. Please consider PR #3194 as prefered.

Products/CMFPlone/browser/iconresolver.py
santonelli: I'm not sure. URL is used to construct the src of e.g. an image-tag. I expcected absolute_url of site is right, when using ++resolveiconurl++some-icon.

Products/CMFPlone/browser/iconresolver.py
ale-rt: Thanks!

santonelli:
Followed by #3194

PR #3194, opened 2020-10-12
Add Icon Resolver as View

santonelli:
Icon resolver implemented as view. I'd prefer a view (other option is a traverser). Usage in a template is more like we handle e.g. images. The view can be called directly (including traversal) to get an icon url. A tag method provides the full tag.


santonelli:
@jensens request for feedback. Example is provided in test_rendering.pt.

Todos:
- cleanup
- add url method (for consistence)
- add tests

Products/CMFPlone/browser/icon.py
```diff
@@ -0,0 +1,60 @@
...
+@implementer(IPublishTraverse)
+class IconView:
+
+    prefix = 'plone.staticresources.icon.'
+
+    def publishTraverse(self, request, name):
+        request['TraversalRequestNameStack'] = []
+        return self
+
+    def __init__(self, context, request):
```
ale-rt, line 24: Instead of redefining the `__init__` of the view I would set the icon attribute in the `publishTraverse` method.

jensens:
IIRC a view is slightly slower than a traverser (but proof me wrong, this is old knowledge and might be outdated). Anyway, if we define the view once (i.e. in main template) and reuse it this would not matter (or would speed up things).

Products/CMFPlone/browser/icon.py
jensens: ... and inherit from BrowserView

Products/CMFPlone/browser/icon.py
```diff
@@ -0,0 +1,60 @@
...
+            self.icon = request.path[0]
+
+    def __call__(self, icon=None):
+        if icon:
+            self.icon = icon
+        name = self.icon
+        url = getSite().absolute_url() + '/' + self.get_icon(name)
+        return url
+
+    def get_icon(self, name):
```
jensens, line 36: i dont like the Java'esc `get_icon`. I would name it `registered_icon` or `icon_from_registry` (latter is what it is).

Products/CMFPlone/browser/icon.py
```diff
@@ -0,0 +1,60 @@
...
+from zope.interface import Interface
+from zope.location.interfaces import LocationError
+from zope.traversing.interfaces import ITraversable
+from zope.publisher.interfaces import IPublishTraverse
+
+
+@implementer(IPublishTraverse)
+class IconView:
+
+    prefix = 'plone.staticresources.icon.'
```
jensens, line 18: we should name it just `plone.icon.`

Products/CMFPlone/browser/icon.py
```diff
@@ -0,0 +1,60 @@
...
+    def publishTraverse(self, request, name):
+        request['TraversalRequestNameStack'] = []
+        return self
+
+    def __init__(self, context, request):
+        super(IconView, self).__init__(context, request)
+        if len(self.request.path) == 1:
+            self.icon = request.path[0]
+
+    def __call__(self, icon=None):
```
jensens, line 29: API wise I would put the code in here in a `def url(....` and use it in `__call__` (if at all). But when used like an API it read better:
```
<div tal:define="icon_view nocall:context/@@icon">
   <img src="${python:icon_view.url('alarm')" />
   <p><tal:icon tal:replace="structure python:icon_view.tag('archive', tag_alt='foobar')" /></p>
```

Products/CMFPlone/browser/icon.py
santonelli: Setting the icon attribute in publishTraverse doesn't work for me. I just followed https://docs.plone.org/develop/plone/serving/traversing.html#custom-traversal.

Products/CMFPlone/browser/icon.py
santonelli: This needs to fit to what we register in plone.staticresources. I agree to keep the name as short as possible.

santonelli:
Should we entirely get rid of the traversal stuff? The view (directly called) could return the actual icon (from registry). In Code and templates you can use .tag() and .url() based on what you need.

Products/CMFPlone/browser/icon.py
jensens: I used this several times. How does it not work?

Products/CMFPlone/browser/icon.py
jensens: Then we should change it over there.

Products/CMFPlone/browser/templates/test_rendering.pt
```diff
@@ -457,7 +482,7 @@
           <span class="optionsToggle">
             <input type="checkbox" checked="checked" class="noborder" id="pt_toggle" value="#" name="pt_toggle"
               onclick="javascript:toggleSelect(this, 'portal_type:list', true)"
-              src="http://localhost:8080/Plone/select_all_icon.png">
+              src="http://localhost:8080/Plone/select_all_icons.png">
```
jensens, line 485: ${context/portal_url}
```suggestion
              src="${context/portal_url}/select_all_icons.png">
```

Products/CMFPlone/browser/configure.zcml
```diff
@@ -263,4 +263,11 @@
       permission="zope2.View"
       />
 
+  <browser:page
+      for="*"
+      name="icons"
```
jensens, line 268: I would not name it that generic, but what it is:
```suggestion
      name="iconresolver"
```
(need changes below as well)

Products/CMFPlone/browser/templates/test_rendering.pt
santonelli: The whole section (everything below examples of how to use new icons view) is commented out. This is work in progress an should be done/cleaned up when most of the templates/sections in Plone are done.

Products/CMFPlone/browser/configure.zcml
santonelli: During yesterday's discussion we said @@icons instead of @@icon would be acceptable. @@icons is similar to @@scales. It's short, it's simple. Let's keep it.

jensens: In facts its defined once in main template, so I would prefer the adapter name to match the functionality. The resulting name in tal:define can be icons, that's what I understood yesterday.

agitator: a good compromise... iconresolver says what it does, icons as usage makes it easy to use

PR #3195, opened 2020-10-12
Robot tests: Do not use jQuery.size() but use ``.length`` instead.

thet:
``.size()`` is deprecated since 1.8.

This PR targets the ``5.2.x`` branch.
The one for the ``master`` branch is at: https://github.com/plone/Products.CMFPlone/pull/3196

PR #3196, opened 2020-10-12
Robot tests: Do not use jQuery.size() but use ``.length`` instead.

thet:
``.size()`` is deprecated since 1.8.

This PR targets the ``master`` branch.
The one for the ``5.2.x`` branch is at: https://github.com/plone/Products.CMFPlone/pull/3195


santonelli (approved):
Looks good to me.

thet:
Applied via rebase on 5.2.x just now.

PR #3198, opened 2020-10-14
Remove unused redirectOnResponse on delete action

### LINKED ISSUE #3197: Toolbar delete action - ignored actionOption redirectOnResponse for modal pattern
The delete action menu item in the Plone toolbar uses pat-modal, with an `actionOptions` data object to provide configuration for the pattern.  This is the rendered HTML:

```
<a href="https://blahblahblah/delete_confirmation?_authenticator=blah"
  title=""
  id="plone-contentmenu-actions-delete"
  class="actionicon-object_buttons-delete pat-plone-modal"
  data-pat-plone-modal="{&quot;actionOptions&quot;: {&quot;disableAjaxFormSubmit&quot;:true, &quot;redirectOnResponse&quot;:true}}">
    Delete
</a>
```

Both `disableAjaxFormSubmit` and `redirectOnResponse` are set to `true`.  However, looking at the pattern source, it appears that the place where `redirectOnResponse` is checked would never be reached if `disableAjaxFormSubmit === true`:

https://github.com/plone/mockup/blob/master/mockup/patterns/modal/pattern.js#L383

Therefore, I would say that setting `redirectOnResponse` is superfluous.

The action menu comes from here:

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/profiles/default/actions.xml#L192

fulv:
Fixes #3197

### DIFF
--- a/Products/CMFPlone/profiles/default/actions.xml
+++ b/Products/CMFPlone/profiles/default/actions.xml
@@ -189,7 +189,7 @@
     <element value="Delete objects"/>
    </property>
    <property name="visible">True</property>
-   <property name="modal" type="text">{"actionOptions": {"disableAjaxFormSubmit":true, "redirectOnResponse":true}}</property>
+   <property name="modal" type="text">{"actionOptions": {"disableAjaxFormSubmit":true}}</property>
   </object>
   <object name="rename" meta_type="CMF Action" i18n:domain="plone">
    <property name="title" i18n:translate="">Rename</property>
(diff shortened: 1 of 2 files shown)


Products/CMFPlone/profiles/default/actions.xml
```diff
@@ -269,7 +269,7 @@
     <element value="View"/>
    </property>
    <property name="visible">True</property>
-   <property name="modal" type="text">{"prependContent": ".portalMessage", "title": "Log in", "width": "26em", "actionOptions": {"redirectOnResponse": true}}</property>
+   <property name="modal" type="text">{"prependContent": ".portalMessage", "title": "Log in", "width": "26em", "actionOptions": {"ReResponse": true}}</property>
```
laulaz, line 272: There is a typo / bad remove here :wink:

PR #3208, opened 2020-11-09
Remove meta_type index and metadata from catalog.

jensens:
Both are unused, with Dexterity the metatype is almost the same. Removing has some advantages:
- smaller portal catalog
- less probability of transaction conflicts
- faster indexing


jensens:

jensens:

ale-rt:
I guess it is intended to not have a related PR in plone.app.upgrade.
BTW I never used meta_type for querying stuff around but you never know what is happening in to the wild.
I do not know what other people of the @plone/framework-team think about this change, but it seems to me a little bit to sneaky.
We should have it at least well documented (somehow).

ale-rt:
My 2 c.: I do not think that removing one small FieldIndex and the associated catalog will change dramatically the performances of a website.

frapell:
I agree... I do remember using it, but a loooong time ago. I don't have any objections to this change, however I would put some sort of warning that could be removed on a later version that if there is a "meta_type" key in the query, it would print something like "maybe you meant portal_type?"

jensens:
1. no, by intend no upgrade step
2. It has no impact on read performance
3. it has an impact on write performance and on possible conflicts (since there are with dexterity only two values, so a bucket conflict on write may happen)
4. I do not think a warning is needed. It bloats the code for real rare case (hitting it is like winning in the lottery)

jensens:
One against a warning: Someone may want to add *by intend* add a `meta_type` index.

PR #3213, opened 2020-11-18
Swallow plone.theme in Products.CMFPlone

jensens:


jensens:

jensens:
this one would need  a start over after plone.base was introduced after this PR

PR #3224, opened 2020-12-23
Fix problem to remove username and password if there was already one set

jensens:
If there was an email user set and I do not need one any more, its was impossible to blank the field.


Products/CMFPlone/controlpanel/browser/mail.py
```diff
@@ -37,8 +37,9 @@ def save(self):
             self.status = self.formErrorsMessage
             return False
         # keep password field
-        if (data.get('smtp_userid') is not None and
-                data.get('smtp_pass') is None):
+        smtp_user_id = (data.get('smtp_userid') or '').strip()
```
ale-rt, line 40: I would fix the `smtp_user_id` vs `smtp_userid` inconsistency. But it is really minor.

ale-rt (approved):
LGTM, it would be nice to have a test

mauritsvanrees:
What is the problem you are trying to solve here?
To me it looks like you mean this:

1. Set smtp user id and password in the Mail Settings control panel and save
2. Remove them and save.
3. Problem: both fields still filled in.

But this does not happen when I try it. Tried with coredev 5.2 on Py 3 (and customer project with 5.2.3 on Py 2). Looking in the portal_registry, the field values are empty after this.

Also, on Jenkins, `Products.CMFPlone.controlpanel.tests.test_controlpanel_browser_mail.MailControlPanelFunctionalTest.test_mail_controlpanel_smtp_pass_keep_on_saving` fails.

jensens:
I had the problem in Coredev 6. I can reinvestigate, but after this change all worked as excpected.

mauritsvanrees:
I cannot reproduce a problem in 6.0 either.

Products/CMFPlone/controlpanel/browser/mail.py
```diff
@@ -37,8 +37,9 @@ def save(self):
...
+        smtp_pass = (data.get('smtp_pass') or '').strip()
+        if not (smtp_user_id or smtp_pass):
```
mauritsvanrees, line 42: To match the original intent, the condition would need to be:

```
if smtp_user_id and not smtp_pass:
    del data['smtp_pass']
```

This would likely fix the failing test.

PR #3230, opened 2021-01-20
Plone 6: Removed dependency on Products.Sessions.

mauritsvanrees:
It is still pulled in by `Products.PluggableAuthService` though.
See also https://github.com/plone/Products.CMFPlacefulWorkflow/issues/35

Since it is still pulled in, this change should not matter. I will only run one Jenkins job, just for good measure.

### DIFF
--- a/news/2957.bugfix
+++ b/news/2957.bugfix
@@ -1,0 +1,4 @@
+Removed dependency on Products.Sessions.
+It is still pulled in by Products.PluggableAuthService though.
+See also `CMFPlacefulWorkflow issue 35 <https://github.com/plone/Products.CMFPlacefulWorkflow/issues/35>`_.
+[maurits]
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,6 @@ setup(
         'Products.PluggableAuthService',
         'Products.PluginRegistry',
         'Products.PortalTransforms',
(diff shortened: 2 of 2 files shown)


gforcada (approved):

gforcada:
What is one supposed to use for sessions then, if this one goes away, is that documented somewhere? :thinking:

mauritsvanrees:
You still need `Products.Sessions` if you want to use sessions. But `Products.Sessions` is only the glue between Zope/Plone and the actual session implementation. You need some extra package. Until and including Plone 5.2, by default this was `Products.TemporaryFolder`, but this was unreliable, so we will no longer ship it with Plone 6. So you need an alternative session backend.

The short overview is still in the [`Products.Sessions` readme](https://github.com/zopefoundation/Products.Sessions).
For further details it points to the [Zope book on Sessions](https://zope.readthedocs.io/en/latest/zopebook/Sessions.html) and especially its section on [alternative session backends](https://zope.readthedocs.io/en/latest/zopebook/Sessions.html#alternative-server-side-session-backends-for-zope-4).

I guess if you write an add-on that needs sessions, you should add `Products.Sessions` in your `install_requires`, and in the readme of your add-on you add links to the Zope book.

PR #3234, opened 2021-02-02
Search Page updates

petschki:


Products/CMFPlone/browser/templates/search.pt
```diff
@@ -35,11 +39,15 @@
...
                       st python:request.get('SearchableText', '');
                       DateTime python:modules['DateTime'].DateTime;
-                      navigation_root_url view/navroot_url;">
+                      navigation_root_url view/navroot_url;
+                      icons python:context.restrictedTraverse('@@iconresolver');">
 
         <input type="hidden" name="sort_on" value="" />
         <input type="hidden" name="sort_order" value="" />
+        <input type="hidden" id="search-batch-start" name="b_start:int" value="${request/b_start|string:0}" />
+        <input type="hidden" id="advanced-search-input" name="advanced_search" value="${python: view.show_advanced_search() and 'True' or 'False'}" />
```
jensens, line 48:
```suggestion
        <input type="hidden" id="advanced-search-input" name="advanced_search" value="${python: 'True' if view.show_advanced_search() else 'False'}" />
```

jensens:
While at it, I would have converted a bunch of existing TALES expression into Python expressions.

PR #3242, opened 2021-02-11
Require plone.autoinclude

tschorr:
@mauritsvanrees I have another PR for plone.autoinclude, but I don't have permission to push in https://github.com/plone/plone.autoinclude/. Could you check?


mauritsvanrees (approved):
That is good, thanks.

I have now given the plone developers team write access to plone.autoinclude.

PR #3246, opened 2021-02-12
Fix migration page

### LINKED ISSUE #3245: I cannot migrate a Plone 5.2 ZoDB to 6
I took a Plone 5.2.3 Database with some history and I tried to migrate it to Plone 6 (buildout coredev, dx site root plip cfg).

When I browse to the site root http://localhost:8080, I have this:
```
...
AttributeError: 'MountedObject' object has no attribute 'getId'

 - Expression: "view/sites"
 - Filename:   ... ne/Products/CMFPlone/browser/templates/plone-overview.pt
 ...
```

The reason is that we have the `temp_folder` which is already meant to be removed in Plone 6 but unluckily this is broken and so the meta_type attribute can be found:
```
((Pdb++)) pp root.values()                                                     
[<BrowserIdManager at /browser_id_manager>,
 <SessionDataManager at /session_data_manager>,
 <SiteErrorLog at /error_log>,
 <persistent broken Products.ZODBMountPoint.MountedObject.MountedObject instance b'\x00\x00\x00\x00\x00\x00\x00\t'>,
 <Products.SiteAccess.VirtualHostMonster.VirtualHostMonster object at 0x7f304b3984a0 oid 0xa in <Connection at 7f3051c6a490>>,
 <ZopePageTemplate at /index_html>,
 <PloneSite at /Plone>,
 <PluggableAuthService at /acl_users>]
```

ale-rt:
Fix migration when we have broken objects in the app root (e.g. the temp_folder)

Fixes #3245


ale-rt:

PR #3254, opened 2021-02-26
Add test that there is no Uninstall button in a new site.

mauritsvanrees:
I try to show for which package the Uninstall button is.
Currently you get this error because plone.app.contenttypes has an upgrade profile that is not hidden yet:

```
AssertionError: Uninstall button found. Probably an upgrade profile needs to be hidden. <form action="uninstall_products" class="float-end" method="post">
<input name="uninstall_product" type="hidden" value="plone.app.contenttypes"/>
<button class="btn btn-sm btn-danger" name="form.submitted" type="submit" value="Uninstall">Uninstall</button>
<input name="_authenticator" type="hidden" value="7783303fbea316a6b26de498519664ff59e3641e"/></form>
```

I fix that problem in https://github.com/plone/plone.app.contenttypes/pull/591
I will test the two PRs together for good measure.


ale-rt (approved):
Thanks!

PR #3258, opened 2021-03-03
Add PLONE60MARKER Python marker

sneridagh:
I already missed it when making collective.recipe.plonesite work for Plone 6. :)

### DIFF
--- a/news/3257.feature
+++ b/news/3257.feature
@@ -1,0 +1,2 @@
+Add IMREALLYPLONE6 Python marker
+[sneridagh]
--- a/news/3257.feature
+++ b/news/3257.feature
@@ -1,2 +1,2 @@
-Add IMREALLYPLONE6 Python marker
+Add PLONE60MARKER (and PLONE52MARKER) Python marker
 [sneridagh]
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -18,6 +18,7 @@ _CONTENT_PROFILE = 'plone.app.contenttypes:plone-content'
 
 # A little hint for PloneTestCase
 _IMREALLYPLONE5 = True
+_IMREALLYPLONE6 = True
 
 logger = getLogger('Plone')
 
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -16,9 +16,14 @@ _DEFAULT_PROFILE = 'Products.CMFPlone:plone'
 _TYPES_PROFILE = 'plone.app.contenttypes:default'
 _CONTENT_PROFILE = 'plone.app.contenttypes:plone-content'
 
-# A little hint for PloneTestCase
+# A little hint for PloneTestCase (pre-Plone 6.0)
 _IMREALLYPLONE5 = True
-_IMREALLYPLONE6 = True
+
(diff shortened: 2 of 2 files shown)


sneridagh:

jensens:
OMG, we have such a marker? :-D

mauritsvanrees (changes requested):
The comment "A little hint for PloneTestCase" can be removed. Maybe:

    A little hint for code that needs to know the major Plone version.

We should probaby remove `_IMREALLYPLONE5`. In commit 00e5e5e8213802b2cf1b1e348c2ea30c4c977f95 for Plone 5.0, @davisagli replaced the Plone 4 hint with the Plone 5 hint. So the original intention seems to have been to only mention the current major Plone version in here, not previous ones.

But if we remove the Plone 5 hint, then maybe some code in `plone.restapi` and `plone.app.upgrade` needs to change. Those are the only core packages that use this, and then mostly in tests.

sneridagh:
@mauritsvanrees I initially thought that we could use it the same way (incrementally) as we already do with the ZCML conditions. So Plone 6 is (implicitly) also Plone 5. This way we do not break old packages. But it's true that then we loose completely the "IMREALLY" point :) 

We could revisit and rethink the whole thing though... I think it's handy to have such a marker for things like the one in:

https://github.com/collective/collective.recipe.plonesite/pull/23 

so you don't have to do the full `pkg_resources` dance every single time... specially on majors where you break things and need to branch code here and there.

What do you think?

mauritsvanrees:
The `pkg_resources` dance may be more reliable, or at least more fine grained, instead of only a major version.
But the downside is that I won't be able to remember how to check this, so I would have to search the current packages to copy a few lines. And we are probably doing this in at least three different ways. And I wonder if corner cases like an alpha release are recognised correctly by all of them.

The [zcml way](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/meta.zcml#L13-L21) is well known, with Plone 6 having markers from plone-4 to plone-60. I can see that a similar way in Python code could help, keeping previous `_IAMREALLY*` settings. But the name does not really fit, and the underscore suggests you should not really use this.

Maybe we can start using a different name? Well, in Plone 5 we keep the current `_IAMREALLYPLONE5` of course.

- Plone 5.2: `PLONE52-MARKER`
- Plone 6.0: `PLONE52-MARKER`, `PLONE60-MARKER`
- Plone 6.1: `PLONE52-MARKER`, `PLONE60-MARKER`, `PLONE61-MARKER`

I guess the current `factory.py` is an okay place for this.

Does that make sense?

jensens:
I think that's fine now.

news/3257.feature
```diff
@@ -0,0 +1,2 @@
+Add IMREALLYPLONE6 Python marker
```
mauritsvanrees, line 1: Can you fix the news snippet to use the new `PLONE60MARKER`? After that, it is fine to merge it yourself.
Will you create a PR to add the `PLONE52MARKER` in 5.2?

sneridagh:
@mauritsvanrees created:  https://github.com/plone/Products.CMFPlone/pull/3264

jensens (approved):

Issue #3260, opened 2021-03-11
Update tests to match plone.dexterity changes - ignorePrefix


avoinea:
See https://github.com/plone/plone.dexterity/pull/145

mauritsvanrees:
Closing, see https://github.com/plone/plone.dexterity/issues/144#issuecomment-1444632183.

PR #3261, opened 2021-03-11
Update tests to match `ignorePrefix = True`

avoinea:
- See #3260 and https://github.com/plone/plone.dexterity/pull/145


mauritsvanrees:
Closing, see https://github.com/plone/plone.dexterity/issues/144#issuecomment-1444632183.

Issue #3229, opened 2021-01-20
Plone 6: Drop Python 3.6 and 3.7, add support for Python 3.9


jensens:
## Abstract

Plone 6: Drop Python 3.6 and 3.7, add support for Python 3.9

## Motivation

- Python 3.6 [support ends 12/2021](https://www.python.org/dev/peps/pep-0494/) 
- be able to use [nice features of Python 3.8+](https://realpython.com/python38-new-features/)
- less test runners on Jenkins needed

## Assumptions

It is easy to install and use Python 3.8 these days on all platforms.

## Proposal & Implementation

- remove Jenkins jobs for 3.6 + 3.7 on buildout.coredev 6
- add Jenkins job for Python 3.9
- ensure whole stack runs on 3.9
- remove troove classifiers for 3.6 + 3.7 on all core packages used only in Plone 6
- add troove classifiers for 3.9  to all core packages used in Plone 6
- fix documentation where needed
- fix installer (probably there will be a new installer anyway)

## Deliverables

- Core Python packages
- Documentation

## Risks

System Python might be not enough on some older systems to run Plone. Debian Buster comes with 3.7, but Ubuntu 20.04LTS comes with 3.8 and next Debian release bullseye is already in freeze since Jan 12th 20221, coming with 3.9. 

## Participants

- Jens Klein
- add your name

jensens:
Lets see if this is fine, poll at https://community.plone.org/t/plone-6-python-version-support/13353/2

mauritsvanrees:
For an overview, all dates for active Python releases can be found at the top of https://www.python.org/downloads/
I would add that a new major Python version is released each year now. Python 3.10 is expected in October this year, see [PEP 619](https://www.python.org/dev/peps/pep-0619/). Somewhere in the Plone 6 lifecycle we should start supporting it. I expect we have Plone 6.0 before October, so we won't have support for 3.10 then yet.

I agree it makes sense to restrict the amount of Python versions we support and test. For 5.2 we support four versions, and that is really a lot. Cannot do much about that in 5.2 though. I would say supporting two or three major versions is the maximum. But with a new Python version each year, we could easily find ourselves wanting to support more...

We could say:

- 6.0: support 3.7, 3.8. 3.9
- 6.1: support 3.8, 3.9, 3.10

In 6.0 I am +1 for dropping 3.6, +0.5 for dropping 3.7, +1 for adding 3.9.
But yes, the poll is a good idea.

[...]

PR #3282, opened 2021-06-18
Removed the docstring from methods to avoid publishing them [6.0]

mauritsvanrees:
From [Products.PloneHotfix20210518](https://plone.org/security/hotfix/20210518/reflected-xss-in-various-spots).

### DIFF
--- a/Products/CMFPlone/PloneTool.py
+++ b/Products/CMFPlone/PloneTool.py
@@ -324,11 +324,9 @@ class PloneTool(PloneBaseTool, UniqueObject, SimpleItem):
 
     @security.protected(View)
     def getReviewStateTitleFor(self, obj):
-        """Utility method that gets the workflow state title for the
-        object's review_state.
-
-        Returns None if no review_state found.
-        """
+        # Utility method that gets the workflow state title for the
+        # object's review_state.
+        # Returns None if no review_state found.
         wf_tool = getToolByName(self, 'portal_workflow')
         wfs = ()
         objstate = None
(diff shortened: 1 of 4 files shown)


mauritsvanrees:

icemac:
Why not make the methods private by using `@security.private`? There is https://github.com/zopefoundation/Zope/issues/774 – which is not yet implemented – but it would break this code again.

gforcada:
At least for Plone 6 and Zope 5 that would be the perfect timing for doing such a breaking change, otherwise we would have to wait until Plone 7 and Zope 6? 😅 that would be far too long IMHO

jensens:
> Why not make the methods private by using @security.private?

I think the idea of the PloneTool is to provide functionality to be used in untrusted templates, but are not meant to be directly published. Well, overall today one would write this different.

There are probably multiple places to improve here.

mauritsvanrees:
> I think the idea of the PloneTool is to provide functionality to be used in untrusted templates, but are not meant to be directly published. Well, overall today one would write this different.

Indeed. Untrusted code should be able to call these. But they should not be available at a url, because that would make reflected XSS easy. `@security.public` without a docstring currently makes this combination possible.

I think this is a valid use case, which would break when docstrings are not required anymore. That *does* seem the right direction though. And the use case gets less important, because less code is in skins these days. So we *could* choose to tighten the security in Plone 6, in preparation for stricter handling in Zope. Maybe start saying that doing things in skins is deprecated, and people should move to browser views and resources. (But Plone itself still uses a few skin folders.) That is a wider discussion though, not something for a random PR.

Anyway, for now I think the changes in this PR are fine. Can someone approve?

@gforcada The checks expect a Jenkins Plone 6.0 Python 3.6 job, but this is not supported, so it can be ignored. Do you think this would be fixed by calling "Re-create GitHub commit hooks" on https://jenkins.plone.org/roboto/ ? It is probably safe to try, and I can do that.

gforcada (approved):

gforcada:
@mauritsvanrees no not there, that's on the mr.roboto configuration: https://github.com/plone/mr.roboto/pull/81 I will check why 3.6 is still expected for 6.0

gforcada:
I see that the configuration looks fine, so it might be a bug on mr.roboto? 🤔

jensens:
Did we dropped Python 3.6 for Plone 6.0?

mauritsvanrees:
> Didn't we dropped Python 3.6 for Plone 6.0?

Yes we did. Well, officially this is still pending an [okay](https://github.com/plone/Products.CMFPlone/issues/3229) from the framework team. But when we sat with a few people a month or two ago to fix some Jenkins stuff, we removed 3.6 from the checks already. (Could be reactivated if the framework team decides otherwise.)

So the point now is that it is strange that 3.6 is still in the checks here.

I have called the "Re-create GitHub commit hooks" command from mr.roboto, to see if that fixes it. It is still running, because it needs to touch a lot of repositories. I see it is also adding hooks to some Guillotina repositories. I hope that is okay.

Ah, it seems finished, no further logs are coming in. Nothing has changed in this PR, but perhaps in a fresh PR it will work correctly.

Anyway, this PR is approved, I will merge.

gforcada:
@mauritsvanrees the hooks are to make repositories (not PR) to mr.roboto, so that whenever an action happens in a repository mr.roboto knows about it, it has nothing to do with which jobs a PR is expected to run 😅 

The code for that is https://github.com/plone/mr.roboto/blob/3fd95d071d62f30caf6d55028767cedbde651439/src/mr.roboto/src/mr/roboto/subscriber.py#L375 (I think 😅 I had only a brief look at the code in 2 seconds)

PR #3287, opened 2021-07-07
plip favicon settings in ControlPanel

talarias:


talarias:
https://github.com/plone/plone.app.layout/pull/255

MrTango:

MrTango:
@talarias please push your changes, so that we can help ;)

Rudd-O:
Wow, this is really cool.  Let's sort the outstanding items out.  Also, can you add tests similar to the ones related to the site logo?  Other than that, LGTM.  Thanks.

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -1005,6 +1005,19 @@ class ISiteSchema(Interface):
...
     )
 
+    site_favicon = schema.Bytes(
+        title=_('Site Favicon'),
+        description=_('This shows a custom favicon on your site.'),
+        required=False,
+    )
+    
+    site_favicon_mimetype = schema.TextLine(
+        title=_('Site Favicon Mimetype'),
```
Rudd-O, line 1015: Change to "MIME type of the site favicon"

Products/CMFPlone/interfaces/controlpanel.py
```diff
@@ -1005,6 +1005,19 @@ class ISiteSchema(Interface):
...
+        description=_('Automatically setted Mimetype of the Favicon'),
```
Rudd-O, line 1016: Change to "MIME type of the favicon (automatically set when a new favicon is uploaded)".

Products/CMFPlone/browser/favicon.py
```diff
@@ -0,0 +1,26 @@
...
+
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(ISiteSchema, prefix="plone")
+        if getattr(settings, 'site_favicon', False):
+            filename, data = b64decode_file(settings.site_favicon)
+            data = NamedImage(data=data, filename=filename)
+            self.data = data
+            self.filename = filename
+
+    def _getFile(self):  # todo add caching
```
Rudd-O, line 25: The caching is important.

Products/CMFPlone/browser/favicon.py
```diff
@@ -0,0 +1,26 @@
...
+class SiteFavicon(Download):
+
+    def __init__(self, context, request):
+        super().__init__(context, request)
+        self.filename = None
+        self.data = None
+
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(ISiteSchema, prefix="plone")
+        if getattr(settings, 'site_favicon', False):
```
Rudd-O, line 19: What's the fallback when there is no favicon?  Should be included along the logo image, right?

Rudd-O (changes requested):

Products/CMFPlone/browser/favicon.py
talarias: it looks like that im wrong and there is no caching needet. 
Why no caching ?
- If there is no favicon this view will not called
- Fallback is the Plone-default-favico at .../Plonesite/favicon.ico

talarias: it looks like that im wrong and there is no caching needet in this case.

Products/CMFPlone/browser/favicon.py
talarias: Why no fallback in this case ?
- If there is no favicon this view will not be called
- The Fallback is the Plone-default-favico at ".../Plonesite/favicon.ico" in that case that there is no favicon

Rudd-O:
There is one question that is raised by the favicon customization (which is clearly great work and should not be impeded by the following question).

What do we do about the iPhone/iPad icons?  These are normally shipped with the theme.  They are worth customizing, although they are separate PNG images that come in different sizes.  Currently, they are shipped by the theme, and included on the page by the theme.

Would it be a good idea to, in a separate PR, allow the user to add multiple touch icon images, with multiple dimensions, as a list of text lines in the registry?

talarias:
continued work in https://github.com/plone/Products.CMFPlone/pull/3389

PR #3293, opened 2021-07-27
Test fix: wait until the folder contents items are loaded.

mauritsvanrees:
With plone.app.content 4.0.0a3 this takes slightly longer because we call the safe html transform.
A few robot tests on Jenkins fail more often now, and I see this locally as well, although not on every run.


mauritsvanrees:
A test fix needs no changelog entry.

PR #3298, opened 2021-08-11
Protect @@historyview with Modify portal content permission

### LINKED ISSUE #3297: Change permission of @@historyview
The toolbar-action to see `@@historyview` and that view itself is protected by the permission `zope2.View` only.
This has the negative effects:

a) The action is shown in the toolbar but no information is visible in the view unless you can actually edit the page.
b) The view itself is rendered with the css-class `frontend` requiring it to be themed in a custom theme.

We discussed that in the plone6 classic meeting and will protect the view and the action with `Modify portal content` which fixes both issues. 

The modification-date shown on the link of the action is already visible in the byline to readers and also to anonymous if the setting `allow_anon_views_about` is enabled in `/@@security-controlpanel` and `display_publication_date_in_byline` is enabled in `@@site-controlpanel`.

pbauer:
Fixes #3297


pbauer:

pbauer:

PR #3309, opened 2021-09-11
Update security.md to forbid use of huntr.dev

MatthewWilkes:
Clarify that using this third party vuln system that buys vulns is a violation of our policy.


dataflake:
FYI, I updated the `zopefoundation` policy to match the new wording: https://github.com/zopefoundation/.github/blob/master/SECURITY.md

mauritsvanrees (approved):
Thanks. I took over the suggestion from @dataflake.
I also moved this file to the `.github` directory.

I see there is a `zopefoundation/.github` repository that contains this and other files, and GitHub uses it when a repo does not have it. We may want to do this for the Plone organisation as well. But that is a different issue.
I will merge.

PR #3311, opened 2021-09-13
Add plone.app.caching to list of add-ons to upgrade

mauritsvanrees:
This list gets upgraded when calling `@@upgrade-plone`.
In PR https://github.com/plone/plone.app.caching/pull/83 I add the first upgrade step.

### DIFF
--- a/Products/CMFPlone/MigrationTool.py
+++ b/Products/CMFPlone/MigrationTool.py
@@ -88,6 +88,7 @@ class AddonList(list):
 ADDON_LIST = AddonList([
     Addon(profile_id='Products.CMFEditions:CMFEditions'),
     Addon(profile_id='Products.CMFPlacefulWorkflow:CMFPlacefulWorkflow'),
+    Addon(profile_id='plone.app.caching:default'),
     Addon(profile_id='plone.app.contenttypes:default'),
     Addon(profile_id='plone.app.dexterity:default'),
     Addon(profile_id='plone.app.discussion:default'),
--- a/Products/CMFPlone/MigrationTool.py
+++ b/Products/CMFPlone/MigrationTool.py
@@ -102,7 +102,7 @@ ADDON_LIST = AddonList([
     Addon(profile_id='plone.app.discussion:default'),
     Addon(profile_id='plone.app.event:default'),
     Addon(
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

ale-rt (approved):

Products/CMFPlone/MigrationTool.py
```diff
@@ -88,6 +88,7 @@ def upgrade_all(self, context):
 ADDON_LIST = AddonList([
     Addon(profile_id='Products.CMFEditions:CMFEditions'),
     Addon(profile_id='Products.CMFPlacefulWorkflow:CMFPlacefulWorkflow'),
+    Addon(profile_id='plone.app.caching:default'),
```
jensens, line 91: I think you need to initialize `Addon` it with `check_module=True`, because it is not guaranteed that `plone.app.caching` is there. Like if one does not need it and only depends on `Products.CMFPlone`, instead of `Plone`.

```suggestion
    Addon(profile_id='plone.app.caching:default', check_module=True),
```

jensens (changes requested):

Products/CMFPlone/MigrationTool.py
mauritsvanrees: Good point.
Ah, but the same is true for `CMFPlacefulWorkflow` and `plone.app.iterate`, which are also in this list.
In fact, we do not need this check for these packages. I tried it out with a wrongly spelled name, and the only result is a warning from GenericSetup:

    Version of profile plone.app.katching:default is unknown, refusing to upgrade.

This is the same warning that you get for the others, unless you have actually installed them.

mauritsvanrees: Actually, if we use `check_module`, and the module cannot be imported, we get an info message, which is better than a warning. So let me change this anyway.
I will do the same for a few other packages.

mauritsvanrees: Done.
And I replaced `plone.app.iterate:plone.app.iterate` with `plone.app.iterate:default`.

mauritsvanrees:

jensens (approved):

Issue #3211, opened 2020-11-11
Mockup and resource registry redone


thet:
## Abstract

This PLIP is about reworking Mockup to be based on up-to-date JavaScript standards and to simplify the resource registry.

Mockup will be based on Patternslib version 4 and integrate many concepts from there:

- ES6 module imports instead of RequireJS.
- ES6+ compliant code base.
- Dynamic module loading for 3rd party dependencies.
- Webpack as build system instead of Grunt.
- Jest as test framework.
- Markdown based documentation.

The simplification of the resource registry is largely based on the removal of RequireJS:

- No through-the-web compilation via RequireJS and LESS.
- Bundle registration is kept but resources will be removed as there is no need for them anymore.
- Removing unnecessary attributes from the bundle registration.
- Improvements to the caching URLs.
- Reduction of number of default bundles.

Mockup and the resource registry should get much easier to understand and to work with, more up-to-date and match developer expectations better.

Additionally we want to advertise the development of new JavaScript functionality as web components but any JavaScript framework should be compatible with Plone.

## Motivation

Mockup was once and avant garde answer to the growing problem of managing JavaScript in Plone. Based on Patternslib it offered clear and defined dependencies via RequireJS, a JavaScript testing framework and a flexible build system incorporating the resource registry with through the web compilation and the possiblity to overwrite specific resources with a custom implementation. Other technologies like web components or reactive JavaScript frameworks were not available or not an option back then.

Now, Mockup lags years behind current JavaScript development. It's no big fun to develop for Mockup (although there are nice parts, like live reloading in Plone without having to build bundles). Mockup depends on some long outdated and inactive libraries - the most important here is RequireJS. RequireJS was the corner stone to solve the dependency problem with JavaScript in Plone 4. But meanwhile it prevents us from using modern JavaScript features. There are also other outdated canditates: the complex integration of the Grunt build framework, the use of React 0.10 for our documentation framework, the integration of the unmaintained jquery.drag/drop which holds us on jQuery 1.x (there is a seperate PR for fixing this) - and so on.

The resource registry is also complex and hard to understand. The through the web compilation (actully client side compilation on the web browser) works like magic but creates a hard dependency on RequireJS and a specific version of the LESS compiler. This bounds us to a complex development workflow and prevents developers to make their own choices regarding JavaScript libraries, ECMAScript versions and build frameworks.

For good reasons we Plone community failed get developers on board helping out in the JavaScript area of Plone.

[...]

PR #3325, opened 2021-09-23
Rethink resource registry (part of PLIP 3211)

jensens:
This is part of PLIP https://github.com/plone/Products.CMFPlone/issues/3211  "Mockup and resource registry redone" 
Idea:
- based on webresource https://webresource.readthedocs.io/en/latest/
- uses bundles only, resources are gone
- no TTW compile and related
- overrides are still possible, but there is no UI

TODO:
- [x] webresource unique handling
   - [x] traverser to include bundle hash (as uuid5 namespace)
   - [x] check if referenced resources (i.e.chunks) may be included in hash? if how? or is this just a non-problem?  
     - No, bundler need to be configured to change main file
- [x] new control panel
  - [x] edit existing bundles
  - [x] adding of new bundles
- [x] caching
  - [x]  cache resource tree on request in order to build only once for two viewets
  - [x] cache in tree in ram if in production mode 
- [x] support for TTW manual theme CSS/JS
- [x] support addBundleOnRequest
- [x] rewrite tests (remove most of them)


Products/CMFPlone/resources/webresource.py
```diff
@@ -0,0 +1,26 @@
...
+class PloneBaseResource:
+    @property
+    def file_data(self):
+        return get_resource(self.context, self.resource)
+
+
+class PloneScriptResource(PloneBaseResource, ScriptResource):
+    """Webresource based ScriptResource for Plone"""
+
+    def __init__(self, context, **kw):
```
rnixx, line 16: Init can be moved to PloneBaseResource

Products/CMFPlone/resources/browser/resource.py
```diff
@@ -31,39 +27,162 @@ class ResourceBase:
...
+                crossorigin=None,
+                referrerpolicy=None,
+                hreflang=None,
+                media="all",
+                rel="stylesheet",
+                sizes=None,
+                title=None,
+            )
+
+        resolver_js = ResourceResolver(root_group_js)
```
rnixx, line 178: theme_group_js needs to be passed to ResourceResolver

Products/CMFPlone/resources/browser/resource.py
```diff
@@ -31,39 +27,162 @@ class ResourceBase:
...
+        self.renderer_js = ResourceRenderer(
+            resolver_js, base_url=self.portal_state.portal_url()
+        )
+        resolver_css = ResourceResolver(root_group_css)
```
rnixx, line 182: theme_group_css needs to be passed to ResourceResolver

Products/CMFPlone/resources/browser/resource.py
```diff
@@ -31,39 +27,162 @@ class ResourceBase:
...
+        self.renderer_css = ResourceRenderer(
+            resolver_css, base_url=self.portal_state.portal_url()
+        )
```
rnixx, line 185: This function always resolves js and css, although the deriving viewlets use either js or css resources. Maybe this can be made more explicit by implemeting the plone resource iteration here and let a concrete resource factory implementation (e.g. ``create_resource_for`` or similar) on the subclass create the resource instances?

Products/CMFPlone/resources/browser/resource.py
jensens: I don't think so, its already a subgroup of root group. But since the group contains always only one element we could get rid of it at all.

Products/CMFPlone/resources/browser/resource.py
jensens: see above

Products/CMFPlone/resources/browser/resource.py
jensens: Well, building up two trees with resources works fine here. What's the advantage of your approach?

Products/CMFPlone/resources/browser/resource.py
rnixx: ok

Products/CMFPlone/resources/browser/resource.py
rnixx: just nitpicking. it simply feels odd to build js resources if only css resources are needed and vice versa

jensens: ah, ok, yes sure. if needed a future optimization can be done here.

jensens:

jensens:

Issue #3035, opened 2020-02-07
Performance issue with GlobalSection viewlet


thet:
The PR https://github.com/plone/plone.app.layout/pull/192 restored the ability to show excluded items in the global section viewlet when any of the current path's parent's have exclude_from_navigation set to True.

The PR is implemented in a way that the ``exclude_from_navigation`` catalog index (which was created for the PLIP https://github.com/plone/Products.CMFPlone/issues/2516 ) is used when ``show_excluded_items`` is set to False. This is good.
But ``show_excluded_items`` is set per default to True: https://github.com/plone/Products.CMFPlone/blob/fc45ebf97b5f10aa690e405c9acd5cfde65a5616/Products/CMFPlone/interfaces/controlpanel.py#L838

This is only an issue when the navigation depth is set to something higher than 1, so when a navigation tree is rendered with subnavigation. Plone's default is ``3``.

For a quick test I got these results:

Navigation always rendered as a tile - ``@@plone.app.standardtiles.navigation``

``navigation_depth=2``navigation tree with total 9 items:
``show_excluded_items=True`: 50-90ms.
``show_excluded_items=False`: 50-90ms.

``navigation_depth=2``navigation tree with total 40 items:
``show_excluded_items=True``: 160ms to 230ms.
``show_excluded_items=False``: 60-90ms.

``navigation_depth=3``navigation tree with total 67 items:
``show_excluded_items=True``: 300ms to 600ms.
``show_excluded_items=False``: 80-120ms.

/cc @petschki @agitator @jensens @sneridagh @ale-rt

thet:
Performance fix:
in ``registry.xml``:
```
<?xml version="1.0"?>
<registry>
  <records prefix="plone" interface="Products.CMFPlone.interfaces.controlpanel.INavigationSchema">
    <value key="show_excluded_items">False</value>
  </records>
</registry>
```

agitator:
I'd say this option should be `False` by default

jensens:
I think for Plone 6 we are save to just change the default. No migration step needed, since keeping the behavior from before upgrade is more important.

Issue #3055, opened 2020-02-26
CSS issues with InOut widgets having the same vocabulary


zopyx:
The width of the elements for three fields with identical vocabularies looks ba (Plone 5.2.1)

![screenshot-dev2 zopyx de_5080-2020 02 26-07_21_14](https://user-images.githubusercontent.com/594239/75317653-d31da580-5868-11ea-8856-9e96ddd5d629.png)

jensens:
Closing as outdated (6 years old). About Plone 5.x Mockup InOut widgets. Plone 6 uses Bootstrap 5 + rewritten widgets.
If you think this issue is still relevant please reopen it. Thank you for the original report!

PR #3329, opened 2021-09-30
Set the "Show excluded items" (``show_excluded_items``) to False per default

thet:
Setting it to ``True`` can introduce a performance problem.
``False`` should be the default, also from user expectation for the ``exclude_from_nav`` setting on conte
nt items.
No upgrade step!
Previous behavior is just kept, unless you override it manually.
See: #3055, first comment.
Use this registry snippet to set it false:
```
<?xml version="1.0"?>
<registry>
  <records prefix="plone" interface="Products.CMFPlone.interfaces.controlpanel.INavigationSchema">
    <value key="show_excluded_items">False</value>
  </records>
</registry>
```
Fixes: #3035


thet:

petschki (approved):
Lgtm

PR #3332, opened 2021-10-07
Use HTML5 meta charset

thet:
Using http-equiv is no longer the only way to specify the character set of an HTML document.

Continuation of: https://github.com/plone/Products.CMFPlone/pull/2025


mauritsvanrees:
Why would we use tal replace here? There is nothing dynamic.

PR #3344, opened 2021-10-27
On Zope root, create Volto site by default.

mauritsvanrees:
When the `plone.volto` package is not available, the buttons are the same as before,
and will create a Classic Plone site.

When `plone.volto` is there, you get an extra button for Classic:

Screenshot 2021-10-27 at 10 12 57

- 'Create a new Plone site' will create a Volto site.
- 'Create Classic Plone site' will create a Classic site.
- Advanced still gives you all options. The extra Volto profiles are selected when available.

We could pick different colors.

When installing Volto, we install the caching and Barcelonata profiles just like a Classic site,
although we could choose to not install Barceloneta.
As extra, we install `plone.volto:default` and `plone.volto:default-homepage`.
(Technically, `default-homepage` would already pull in `default`, but let's do this explicitly.)

### DIFF
--- a/Products/CMFPlone/browser/admin.py
+++ b/Products/CMFPlone/browser/admin.py
@@ -33,6 +33,14 @@ from zope.publisher.interfaces import IRequest
 from zope.schema.interfaces import IVocabularyFactory
 
 import logging
+import pkg_resources
+
+
+try:
+    pkg_resources.get_distribution("plone.volto")
+    HAS_VOLTO = True
+except pkg_resources.DistributionNotFound:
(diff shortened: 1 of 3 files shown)


jensens:

sneridagh (approved):
+1 what @tisto said.

mauritsvanrees:
@jenkins-plone-org please run jobs

ericof:
I'm a bit late to the party, but I do think we should merge this but then refactor the logic out of the browser view and into the `factory.py` module.

ericof (approved):

PR #3354, opened 2021-11-12
fix test to use document_view as default view for site root

agitator:


agitator:
merge together with https://github.com/plone/plone.app.contenttypes/pull/624

mauritsvanrees (approved):

PR #3359, opened 2021-11-18
Allow resolving of icons files stored in database

laulaz:
This PR allows to have icons stored in database, but also use `portal_resources` to override FileSystem icons without changing their url in `portal_registry` records.

I'm not sure about the risks of using `unrestrictedTraverse` in this context, but otherwise you get a traceback like this when accessing a resource stored in `portal_resources` :
```
AccessControl.unauthorized.Unauthorized: The container has no security assertions.
Access to None of (PersistentResourceDirectory object at /Plone/portal_resources/plone/my.product) denied.
```

Thank you for your feedbacks on this change !

### DIFF
--- a/Products/CMFPlone/browser/icons.py
+++ b/Products/CMFPlone/browser/icons.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from io import BytesIO
 from lxml import etree
 from OFS.Image import File
 from plone.registry.interfaces import IRegistry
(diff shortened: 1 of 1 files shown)


jensens:
I do not feel good with `unrestrictedTraverse` here. Shouldn't there be security assertions on the container? Maybe it is more an bug over there and the `unrestrictedTraverse` just do only work around here?

laulaz:
@jensens thank you for your feedback !

Accessing resource via URL works as expected, for example http://localhost:8080/Plone/++plone++my.product/icon.svg

I assume I need to dig into resources related code to find out why you can traverse to the resource via URL, but not through `restrictedTraverse` (for the same url) ...

jensens:
icons are overhauled in Plone 6, so this one is outdated, sorry.

PR #3360, opened 2021-11-18
Load adapter for index location, only if Archetypes is installed

### LINKED ISSUE #3347: Index location doesn't work for Dexterity content
### What I did:

- I created a Dexterity type with the `location` field.
- I created a `location` index of type ZCTextIndex in portal_catalog.
- I created an object of the content type that had the `location` field.

### What I expect to happen:

Hope the catalog indexing the location field of the object created in `location` index.

### What actually happened:

The `location` field is not indexed.

**Plone: 5.2.6**

This is because Plone define an `location` index but uses the `getField` method, which is for Archetype:

https://github.com/plone/Products.CMFPlone/blob/f2d8813d0a1b82cd8ef3adfc0554609452b1e51c/Products/CMFPlone/CatalogTool.py#L258-L260

As Archetype is at the end of its life, I think this index can be removed, or set to Archetype only.

wesleybl:
Adapter for index `location` only works on Archetypes content, as it uses the `getField` method. This caused that, when an `location` index  was created in the catalog, the indexing would not work for Dexterity contents, since the adapter trie to use the `getField` method on a Dexterity content.

Fix #3347

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -276,9 +276,14 @@ def mime_type(obj):
     return aq_base(obj).getPrimaryField().getContentType(obj)
 
 
-@indexer(Interface)
-def location(obj):
-    return obj.getField('location').get(obj)
+# BBB: This adapter is for compatibility and should be removed in Plone 6.
+try:
+    from Products.Archetypes.interfaces import IBaseObject
+    @indexer(IBaseObject)
+    def location(obj):
+        return obj.getField('location').get(obj)
+except ImportError:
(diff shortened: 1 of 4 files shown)


wesleybl:

ale-rt (approved):

ale-rt:
Would you be so kind to remove it completely from Plone 6?
That would be great, so we can definetely close #3347!

wesleybl:

wesleybl:
> Would you be so kind to make another PR to remove it completely from Plone 6?

Done: #3361

PR #3361, opened 2021-11-19
Remove adapter for index location

### LINKED ISSUE #3347: Index location doesn't work for Dexterity content

wesleybl:
Adapter for index `location` only worked with Archetypes content, as it used `getField` method. This caused that, when an `location` index  was created in the catalog, the indexing would not work for Dexterity contents, since the adapter trie to use the `getField` method on a Dexterity content.

Fixes #3347 in Plone 6

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -255,11 +255,6 @@ def mime_type(obj):
     return aq_base(obj).getPrimaryField().getContentType(obj)
 
 
-@indexer(Interface)
-def location(obj):
-    return obj.getField('location').get(obj)
-
-
 @implementer(IPloneCatalogTool)
 class CatalogTool(PloneBaseTool, BaseTool):
     """Plone's catalog tool"""
--- a/Products/CMFPlone/catalog.zcml
+++ b/Products/CMFPlone/catalog.zcml
@@ -17,7 +17,6 @@
     <adapter factory=".CatalogTool.is_folderish"           name="is_folderish" />
     <adapter factory=".CatalogTool.is_default_page"        name="is_default_page" />
     <adapter factory=".CatalogTool.getIcon"                name="getIcon" />
-    <adapter factory=".CatalogTool.location"               name="location" />
     <adapter factory=".CatalogTool.mime_type"              name="mime_type" />
 
 </configure>
--- a/news/3347.bugfix
+++ b/news/3347.bugfix
@@ -1,0 +1,1 @@
+Remove adapter for index location. [wesleybl]


ale-rt (approved):
Top!

wesleybl:

wesleybl:
@ale-rt do you know why the `robot-mister` did not pass this PR? Does it take a while to get through?

ale-rt:
I think there is a problem somewhere in jenkins, if you check: https://jenkins.plone.org/job/pull-request-6.0-3.7/661/console

You can read:
```
Got pbr 5.8.0.
Version and requirements information containing pyparsing:
  [versions] constraint on pyparsing: 3.0.6
  Requirement of httplib2: pyparsing<3,>=2.4.2
  Requirement of packaging: pyparsing!=3.0.5,>=2.0.2
While:
  Installing releaser.
Error: The requirement ('pyparsing<3,>=2.4.2') is not allowed by your [versions] constraint (3.0.6)
Build step 'Virtualenv Builder' marked build as failure
```

I have no idea for the moment.

CC @plone/testing-team

ale-rt:
CC @plone/ai-team

wesleybl:
> I think there is a problem somewhere in jenkins, if you check: https://jenkins.plone.org/job/pull-request-6.0-3.7/661/console

@ale-rt I run buildout locally and also saw this error.

I think it was caused by: https://github.com/zopefoundation/Zope/commit/adacc419e7ba09c1c17a2b43256432b315124923

Look: https://github.com/plone/buildout.coredev/commit/551314149ed6ceda923c2ff48e42311dab034c55

CC @mauritsvanrees

mauritsvanrees:
I have fixed it by upgrading `httplib2` to the latest version.

We indeed always checkout Zope. Bad about this is that we sometimes will suddenly run into a problem like this.  Good about this is that we run into it *now* instead of after a Zope release.

mauritsvanrees:

wesleybl:
Python 3.9 job failed with an unrelated robot test.

I create PR #3362  to fix this failed test.

ale-rt:

wesleybl:
> We indeed always checkout Zope. Bad about this is that we sometimes will suddenly run into a problem like this. Good about this is that we run into it now instead of after a Zope release.

@mauritsvanrees Is this Zope master pin only until Plone 6 is released? What will happen when Plone 6 is released? You probably don't want plone-6.0.cfg to point to the Zope master. Pointing to the latest Zope release doesn't seem right either, since there may be new pins in the master. Are you going to point to a specific commit?

mauritsvanrees:
The Zope version in a Plone release will always be a real Zope release on PyPI, not some checkout. If too much has happened since the last Zope release, we will have to check which packages may be newer in the Zope versions.

PR #3368, opened 2021-11-25
improve iconsresolver for mockup usage

jensens:


Products/CMFPlone/browser/icons.py
```diff
@@ -58,21 +52,39 @@ def _strip_id(svgtree, cfg):
...
+    def publishTraverse(self, request, part):
+        """Subpath detection
+
+        - "@@iconsresolver/plone-copy" delivers the file.
+        - "@@iconsresolver/get_file/plone-copy" delivers the file too.
+        - "@@iconsresolver/get_url/plone-copy" delivers url to the file.
+        - "@@iconsresolver/get_tag/plone-copy" delivers image tag or the SVG data.
+        - "@@iconsresolver/get_tag/plone-copy?class=my-css-info&alt=Plone-Copy-Icon" delivers image tag or the SVG data.
+        """
+        breakpoint()
```
ale-rt, line 70: ```suggestion```

jensens: well, draft, but ok :D

jensens:
Closing this in favor of a potential plone.restapi endpoint.

The `@@iconresolver` view should stay focused on serving icon files via traversal. Structured responses (tag HTML, URLs) are better placed in plone.restapi where JS clients already expect JSON endpoints.

The PR also has issues that would need fixing if pursued here:
- Stale against current main (import cleanup already done, `name = ""` class attr added)
- Multi-segment icon names (`contenttype/document`) break because `self.name = part` overwrites instead of accumulating
- Path-based variety (`get_tag`, `get_url`) risks namespace collision with icon names

PR #3369, opened 2021-11-25
Add plone.synchronize as dependency on 5.2.

mauritsvanrees:
plone.dexterity 2.10.5 has removed this dependency, by taking over the single function it contains.
See https://github.com/plone/plone.dexterity/pull/157
Core Plone does not need the package anymore, but in case someone uses it, it is not nice to lose it in a bugfix release of Plone.
I could have made a branch of plone.dexterity for Plone 5.2, but that is a hassle.
Adding the dependency in CMFPlone 5.2 is easier.
For clarity: we won't add this in CMFPlone 6.0.


mauritsvanrees:

ale-rt (approved):
Smart idea!

PR #3377, opened 2021-12-09
Optional hook to customize breadcrumbs like navigation

jensens:
Navigation already has this hook, same for breadcrumbs.

### DIFF
--- a/Products/CMFPlone/browser/navigation.py
+++ b/Products/CMFPlone/browser/navigation.py
@@ -22,7 +22,7 @@ from zope.interface import implementer
 def get_url(item):
     if not item:
         return None
-    if hasattr(aq_base(item), 'getURL'):
+    if hasattr(aq_base(item), "getURL"):
         # Looks like a brain
         return item.getURL()
     return item.absolute_url()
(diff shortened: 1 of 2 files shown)


jensens:

mauritsvanrees (approved):

ale-rt:
Fine to merge, but you could also do:
```python
class CustomPhysicalNavigationBreadcrumbs(PhysicalNavigationBreadcrumbs):

     def breadcrumbs(self):
           breadcrumbs = super().breadcrumbs()
           # Do whatever you need on the breadcrumbs
           return breadcrumbs
```

It looks to me more pythonic and powerful and basically requires the same effort (register a custom view which extends the base view plus some code to fiddle with the default Plone one)

jensens:
I could, but this way I do not have access to the original contexts of the breadcrumbs and would need to replicate all the code just for one line change. Having a callback per item is way more handy.

ale-rt:
> I could, but this way I do not have access to the original contexts of the breadcrumbs and would need to replicate all the code just for one line change. Having a callback per item is way more handy.

Good point! It definitely makes sense to me now!

PR #3384, opened 2021-12-20
Test-only fix: in test_controlpanel_relations flush the index queue.

mauritsvanrees:
Needed for https://github.com/plone/plone.app.uuid/pull/11
This uuid PR changes code to not do a full catalog query, but it uses some catalog index internals.
Result for two tests in CMFPlone is that the index queue is not flushed, which leads to inconsistent results.
This fixes it.

Internal testing detail only, so I did not add a news snippet.


mauritsvanrees:

Issue #3305, opened 2021-08-19
Plone 6: migrate ulines properties to lines, make lines contain strings


mauritsvanrees:
See https://github.com/zopefoundation/Zope/issues/987

Since Zope 5.3, `ulines` is deprecated and you should use a `lines` property. Same for a few others: utext, utoken, ustring. The unicode variants are planned to be removed in Zope 6.

Also since Zope 5.3, a `lines` property stores strings instead of bytes. But there is no migration yet.
Result is that `getProperty` on an already created `lines` property will return the old value with bytes, but a newly created `lines` property will return strings. This could lead to problems, especially when you insert a new value into the list. GenericSetup had problems with that, and I have created [a fix](https://github.com/zopefoundation/Products.GenericSetup/pull/110) which is merged. It remains to be seen how many problems will arise in other code. I guess it will mostly be okay, but it is not ideal.

We should create some migration code:

- Go over all `lines` properties and make sure they contain a tuple of strings, probably by calling the converter. This should happen in Plone 6.0 alpha or beta.
- Turn all `ulines` properties into `lines`, so they remain working in Zope 6. Could be done later, but it makes sense to do this at the same time.

The best place for this code is probably in Zope itself. `plone.app.upgrade` should call it, probably with `zopeFindAndApply`.

Also, we should look for spots in our own code where we create deprecated unicode properties, or where we add bytes in a lines field.

mauritsvanrees:
In `plone.app.upgrade` we now have an upgrade step, which was merged today.
Further steps are that we should prevent creating deprecated `ulines` and other properties, at least in Plone. Probably in Zope as well.

Looking for [`ulines` in the plone organization](https://github.com/search?q=org%3Aplone+ulines&type=code):

- [ ] `plone.dexterity`. PR: https://github.com/plone/plone.dexterity/pull/161
- [ ] [`Products.CMFDynamicViewFTI`](https://github.com/plone/Products.CMFDynamicViewFTI/blob/3c6a2d5576d8ab713b42bba78fe5cf84d9d7a020/Products/CMFDynamicViewFTI/fti.py#L69), should be fixed in the same way as `plone.dexterity`.
[...]

PR #3385, opened 2021-12-20
Removed management_page_charset support from usergroup-groupdetails

mauritsvanrees:
This is related to deprecated unicode property types, like ustring.
Core Plone should not need them, but there could be custom groups that use them,
so we keep the compatibility checks in here.
Part of https://github.com/plone/Products.CMFPlone/issues/3305


mauritsvanrees:

PR #3391, opened 2022-01-06
updated classifiers

anukaal:
added support for python 3.10

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ setup(
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
     ],
     keywords='Plone CMF Python Zope CMS Webapplication',
     author='Plone Foundation',


mauritsvanrees:
I cannot imagine Plone currently works on 3.10. For starters, you would need an unreleased version of Zope. But I have not tried it.
Have you tried running Plone on Python 3.10?

jensens:
This needs all jenkins testing for 3.10 green first.

anukaal:
> I cannot imagine Plone currently works on 3.10. For starters, you would need an unreleased version of Zope. But I have not tried it.
> Have you tried running Plone on Python 3.10?

no but I guess it will work !

jensens:
> no but I guess it will work !

Sorry, guessing is not in line with our quality policies. First:
- Zope 5 needs a release with 3.10 support.
- Jenkins needs all tests green with 3.10
 
Since we have somewhere a policy to support 3 Python versions we need to officially drop 3.7 - AFAICT.

mauritsvanrees:
Zope 5.4 has been released with Python 3.10 support, so that helps. I tried Plone 6, but it could not start. First error was an incompatibility in `pyScss`: `from collections import Iterable` fails, and should be `from collections.abc import Iterable`.
Once the ES6 PLIP is merged, this package is no longer used, so that helps. With a quick local fix I worked around it and then I could actually create a Plone Site and view it, so the basis does work.
But running the tests, I quickly ran into lots of errors. `robotframework` needs a similar fix. Probably there are more places.

Anyway, thanks for your PR, but more work is needed, so I will close this.

mauritsvanrees:
@jensens wrote:

> Since we have somewhere a policy to support 3 Python versions we need to officially drop 3.7 - AFAICT.

I don't remember this policy. 3.7 is still supported for one and a half years, so it would be good to keep supporting it.
Gets heavier on testing though. :-/

jensens:
> I don't remember this policy.

Or not a real policy, well, I think we spoke in Sorrento about it and came up with the idea. We indeed should write such down. It's OK to me to support 3.7 anyway as long as it is supported by Python Release Management.

PR #3393, opened 2022-01-06
Jmevissen 3241 enhance error log form

jmevissen:
* update error log form layout
* rename error log form


mauritsvanrees (approved):
Tested, and looks good to me.

Note that this PR removes a few buttons: Refresh, Clear Displayed Entries, Show All Entries. This was discussed by a few people. Let's cc @plone/framework-team to see if someone wants to push back on that.

I think I only used the Refresh button occasionally (using Ctrl-R or similar works too), and the others never. I had to start a Plone 4.3 and 5.2 site to check if these buttons where even there in those versions (they were).

mauritsvanrees:
This belongs together with https://github.com/plone/plone.app.upgrade/pull/266, but fine to test separately, as long as they are merged close together.

jmevissen:

jensens:
The buttons do not really make sense. We discussed it and it seems they were never used.

ale-rt:
> The buttons do not really make sense. We discussed it and it seems they were never used.

Fine for me, I hope the other @plone/framework-team are ok with that as well.

If you could add a link to the place were that discussion happened or if you could summarize the reasons that lead to this it would be great.

jensens:
Its easy to bring them back if one is in urgent need. That said, the error log is anyway just the log of one instance after startup and has limited value.

Issue #3395, opened 2022-01-09
PLIP 3395 `plone.base`: Untangles existing circular dependencies (imports/zcml) in Plone.


jensens:
## PLIP (Plone Improvement Proposal)

Dear @plone/framework-team - I am happy to submit this PLIP

## Abstract

This PLIP is a major step toward the cleanup of our dependency mess in Plone core. It untangles the (implicit) dependencies on import/ ZCML level and ensures a clean dependency graph.

## Motivation

After discussion around PLIP #3355, the result was to start with a new `plone.base` package to resolve existing circular dependencies.
Current circular dependencies between `Product.CMFPlone` and several `plone.app.*` and some `Products.*` packages are inconvenient and wrong. My heart is bleeding with this architecture. In #3355 

## Assumptions

`Product.CMFPlone` is consumer of (IOW it imports) most `plone.*` and some `Products.*` packages. If it provides functionality, this is done by adapters and utilities only.

Most circular dependencies are between other core and `Product.CMFPlone` modules/packages 
- `interfaces.*`
- `permissions`
- `utils`
- PloneMessageFactory in NS-root.
- `PloneBatch`
- `defaultpage`
- `i18nl10n`

That said, some packages have special needs, here an analysis of all packages:
 
- [x] `plone.app.contentlisting`: utils, interfaces and `Batch` is used. Latter could live in `plone.base` or moved to `plone.batching`. https://github.com/plone/plone.app.contentlisting/pull/43
- [x] `plone.app.contentmenu`: utils, interfaces https://github.com/plone/plone.app.contentmenu/pull/34
- [ ] `plone.app.contentrules`: 
   - [x] utils, interfaces https://github.com/plone/plone.app.contentrules/pull/71
   - [ ]  breadcrumbs
   - [ ] strange five_template usage
- [ ] `plone.app.contenttypes`: 
  - [x] utils, interfaces, also Batch 
  - [ ] and new: `CMFPlone.browser.syndication.adapters.CollectionFeed`.
- [x] `plone.app.dexterity` utils, interfaces, Batch https://github.com/plone/plone.app.dexterity/pull/348
- [x] `plone.app.discussion` utils, interfaces, Batch, DISCUSSION_ANNOTATION_KEY (could be moved to pa.discussion) https://github.com/plone/plone.app.discussion/pull/195
- [x] `plone.app.layout` https://github.com/plone/plone.app.layout/pull/302
  - [ ] utils, interfaces, 
[...]

PR #3396, opened 2022-01-09
Plip 3395 plone base

jensens:
see #3395


mauritsvanrees:
Slightly easier: `bin/test -u` fails with the same traceback, during test layer setup. End result is:

      Ran 2040 tests with 6 failures, 79 errors, 2 skipped in 16.880 seconds.

mauritsvanrees:
I confirm the tests pass locally (at least the ones that failed earlier for me). I have restarted the PLIP jobs.

jensens:
FTR: link to plip-jobs/3.9 https://jenkins.plone.org/view/PLIPs/job/plip-3395-plone.base-part-1-3.9/

jensens:
And PLIP jobs are green!

mauritsvanrees:
For me it would be good to merge.

@plone/framework-team  have a look at the implementation please.

ale-rt:
I had a look to the `plone.base` package but I am struggling to find the time for an in depth review.

It seems well crafted but I have some doubts.

The [`defaultpage`](https://github.com/plone/plone.base/blob/6e26a377f7153ea6631a6ad58b7a9008fd3d6e53/src/plone/base/defaultpage.py) module pulls in quite a lot of dependencies that otherwise could be dropped and the content is not really something that survived well the time.
Does it really make sense to move it to a new package?

I also noted this [`deprecation`](https://github.com/plone/plone.base/blob/6e26a377f7153ea6631a6ad58b7a9008fd3d6e53/src/plone/base/batch.py#L28) in the `batch` module with a long explanation.

That also feels weird, either we revert the deprecation or remove the code.

That said I already asked if somebody of the @plone/framework-team has some time to review this one.
If in the next week:
1. the tests are green again
2. you think the things I mentioned above are not blockers or something which is worth investing time fixing
3. nobody else from the @plone/framework-team jumps in to (at least) propose review this one

I would just say that you should feel free to push the merge button.

ale-rt:
One other thing... I think we lost some tests.
See https://github.com/plone/plone.base/blob/6e26a377f7153ea6631a6ad58b7a9008fd3d6e53/src/plone/base/utils.py#L69

It mentions `safe_unicode` which is not something we have in that package.
Probably it is way easier to move those doctests in some unitest.

jensens:
> The [`defaultpage`](https://github.com/plone/plone.base/blob/6e26a377f7153ea6631a6ad58b7a9008fd3d6e53/src/plone/base/defaultpage.py) module pulls in quite a lot of dependencies that otherwise could be dropped and the content is not really something that survived well the time.

Indeed. But this might get refactored later. OTOH defaultpage is used by a bunch of plone.app.* packages, so it has to be moved to somewhere. Probably to `Products.CMFDynamicViewFTI`? But we can move it further later, right? I am happy having it that far down for now.

> I also noted this [deprecation](https://github.com/plone/plone.base/blob/6e26a377f7153ea6631a6ad58b7a9008fd3d6e53/src/plone/base/batch.py#L28) in the batch module with a long explanation. 
> 
> That also feels weird, either we revert the deprecation or remove the code.

This is true, but IMO out of scope here. We can tackle this on a different PR later.

jensens:
> One other thing... I think we lost some tests. See https://github.com/plone/plone.base/blob/6e26a377f7153ea6631a6ad58b7a9008fd3d6e53/src/plone/base/utils.py#L69
> 
> It mentions `safe_unicode` which is not something we have in that package. Probably it is way easier to move those doctests in some unitest.

It  seems in a former copy paste session the docstring was copied, but never reviewed. The test does not make any sense for `safe_text` and been probably never run. This would need indeed some love, but also IMO out of scope here.

mauritsvanrees:

mauritsvanrees:
Oops, starting Jenkins via a comment here is not very useful. I stopped the jobs and started the proper PLIP jobs instead.

jensens:
All green. IMO ready for merge?

PR #3397, opened 2022-01-10
Update error_message.pt

yurj:
Use label_site_administration instead of label_site_admin because of the missing article. Must be checked over other languages but italian and english would be ok.


yurj:

PR #3408, opened 2022-01-19
Add option to enable TinyMCE inline mode

pbauer:


pbauer:
This may require changes in mockup (or widget). Currently it fails with:

```
Could not initialize inline editor on invalid inline target element 
<textarea id="id06d8" class="form-control" name="form.widgets.IRichTextBehavior.text">
```

It also requires a upgrade-step that updates the registry.

pbauer:
Obsolete since more mork would be needed. See https://github.com/plone/Products.CMFPlone/issues/3465

thet:
@pbauer I would want to give this another try.
This PR got out of sync with the base branch. Is the last commit the one which is relevant for this PR?

I recently did an example for a different project to enable inline editing among some other tiny customizations (to actually enable TinyMCE in collective.z3cform.datagridfield). This is for Plone 5 - maybe we need some adaptions for the tinymce pattern in Plone 6 to allow inline editing.

This is it:

```
from plone.app.textfield import RichText
from plone.app.z3cform.widget import RichTextWidget
from plone.autoform.directives import widget

[...]

    # inline: True necessary for reordering rows, otherwise tinymce would not render correctly.
    # tiny config to reduce UI
    widget("description", RichTextWidget, pattern_options={
        "inline": True,
        "tiny": {
            "toolbar": "bold italic bullist unlink plonelink",
            "plugins": ["plonelink", "lists", ],
            "menubar": "",
        },
    })
    # RichText for TinyMCE
    # allowed_mime_types to not get the mimetype selector and for tinymce to actually work.
    # the mimetypeselector has a config for the textfield, but here we are cloning textfields and the mimetypeselector config isn't updated.
    description = RichText(title=u"Beschreibung", required=False, allowed_mime_types=["text/html"])

```

thet:
Follow up: https://github.com/plone/Products.CMFPlone/issues/3465

PR #3416, opened 2022-01-27
remove viewlet

erral:
See https://github.com/plone/plone.app.layout/issues/286


Rudd-O:
https://github.com/plone/plone.app.layout/issues/286#issuecomment-1027454879

Products/CMFPlone/profiles/default/viewlets.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <object>
   <order manager="plone.portaltop" skinname="Plone Default">
-    <viewlet name="plone.header" />
```
ale-rt, line 4: This requires an upgrade step, anyway right?
In addition I do not see any point in having an empty:
```
<order manager="plone.portaltop" skinname="Plone Default">
</order>
```

erral: right: https://github.com/plone/plone.app.upgrade/pull/268/files

Yeah... I wasn't sure if we need the order manager empty in order to configure it properly or not... I have never been a big fan of viewlets.xml and never could get it to work properly

MrTango (approved):
LGTM

Products/CMFPlone/profiles/default/viewlets.xml
mauritsvanrees: If a viewlet is no longer registered for a manager in zcml, then it should not be needed to remove it in xml. But does not hurt. I just tried.

mauritsvanrees (approved):

PR #3418, opened 2022-01-29
 Allow access to the favicon via conventional /favicon.ico, simplifying both customizable and fallback favicon handling.

Rudd-O:
There is the justified use case of apps that don't know what favicon URL should be accessed in order to render the URL, such as RSS readers (deployed in the field) and others like podcast subscribers which don't at all parse HTML and shouldn't need to.

This addresses that problem and fixes https://github.com/plone/plone.app.layout/issues/288 .


Rudd-O:

Rudd-O:
Folks, I tested this locally and it works correctly.

Rudd-O:
See also https://github.com/plone/Products.CMFPlone/issues/3419 .

Rudd-O:
One question I haven't resolved is how do I make the view `SiteFavicon` return the data for the **default Plone logo** in the case where the icon is in fact not set up.  Context:

```
        if getattr(settings, 'site_favicon', False):
            filename, data = b64decode_file(settings.site_favicon)
            data = NamedImage(data=data, filename=filename)
            self.data = data
            self.filename = filename
```

This needs an `else:` case that handles that.  But I would not know how to obtain that icon!  Working on it.

Rudd-O:
OK.  I have fixed the default case.

Products/CMFPlone/browser/favicon.py
```diff
@@ -18,9 +20,16 @@ def __init__(self, context, request):
...
         if getattr(settings, 'site_favicon', False):
             filename, data = b64decode_file(settings.site_favicon)
-            data = NamedImage(data=data, filename=filename)
+            mimetype = getattr(settings, 'site_favicon_mimetype', "image/vnd.microsoft.icon")
+            data = NamedImage(data=data, contentType=mimetype, filename=filename)
             self.data = data
             self.filename = filename
+        else:
+            basedir = os.path.dirname(os.path.dirname(__file__))
+            with open(os.path.join(basedir, "skins", "plone_images", "favicon.ico"), "rb") as icon:
```
Rudd-O, line 29: How fortunate that we ship the icon ourselves!

Rudd-O:

Rudd-O:
There is a question whether the favicon views should apply to `*` or to `INavigationRoot`.

Older Plones used acquisition to basically plaster a `favicon.ico` on every folder.

Do we want to continue this in Plone 6?  Or do we want to **only** support `/favicon` and `/favicon.ico` in the navigation roots of the site?

Please let me know what you think!

Products/CMFPlone/browser/favicon.py
Rudd-O: Had to rename the icon, otherwise the skin would trump the view.  That's fine tho — we have a view, so the skin element being absent is no longer a problem anymore.

jensens:
> There is a question whether the favicon views should apply to `*` or to `INavigationRoot`.

Always `INavigationRoot` (b/c of p.a.multilingual, Lineage etc.)

Rudd-O:
OK, I think this is good.  The test failing above appears to be a regression wholly unrelated to this work.

Rudd-O:

Products/CMFPlone/browser/favicon.py
```diff
@@ -18,9 +20,16 @@ def __init__(self, context, request):
...
+            mimetype = getattr(settings, 'site_favicon_mimetype', "image/vnd.microsoft.icon")
+            data = NamedImage(data=data, contentType=mimetype, filename=filename)
             self.data = data
             self.filename = filename
+        else:
+            basedir = os.path.dirname(os.path.dirname(__file__))
+            with open(os.path.join(basedir, "skins", "plone_images", "default-favicon.ico"), "rb") as icon:
+                data = NamedImage(data=icon.read(), contentType="image/vnd.microsoft.icon", filename="favicon.ico")
+            self.data = data
+            self.filename = "favicon.ico"
```
ale-rt, line 32: I am not sure I like this fallback.

First it reuses the skins folder (and we are trying to move away from that, see ).
Second, it uses that in a way that cannot be overwritten TTW.

Rudd-O: These are reasonable objections to which I'd like to answer.

1. We are in fact reusing content from the skins folder, *but we are not using the skins folder or infrastructure here*.  When we finally delete all skins, all we need to do is simply move the single icon to an assets folder somewhere else in the product.  We don't at all depend in this PR on the soon-to-be-removed skins feature.
2. `favicon.ico` can certainly be overwritten through the Web.  All you have to do is upload the icon to the Site Control Panel in Plone.  The infrastructure is already checked in for that — adding that feature was not my call or work, but I was in agreement with those PRs.
3. Additionally, this PR removes the `favicon.ico` on every folder that was (in Plone 5) obtained from the skin via Acquisition, so now any and every folder (other than the INavigationRoot folders) can have `favicon.ico` overwritten through the Web.

Please ack.

ale-rt: What I meant is that it cannot be overwritten changing it in the skins folder, I know that this is a fallback ;)

I am fine removing the old `favicon.ico` from the skins folder I have no problem with that.

I am not fine renaming the favicon inside the skins folder and using that with a low level `open` call.

I still have to ponder what could be a better approach, but for the moment I do not like this approach.

ale-rt: CC @plone/framework-team do you have an opinion on this one?

Rudd-O: I would gladly contemplate alternatives.  That said, at some level, the data for the favicon must be opened from disk, then retrieved, then served.  How that is done specifically, I am happy to change if a better alternative is presented.

One thing I know almost for certain: since we are eliminating skins, we should probably not invest any effort in accessing any data using skins-related mechanisms.

frapell: I am no longer part of the FWT, but I also dislike this fallback solution... If the favicon is not in the settings, then I think it should return None. If a favicon has to be returned, I would rather have a b64 encoded Plone logo instead of a hard-coded path in the filesystem. 
Also, why is the default mt `image/vnd.microsoft.icon` when the field default defined in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/interfaces/controlpanel.py#L1012 is `image/x-icon` ?

Rudd-O: @frapell 

> Also, why is the default mt image/vnd.microsoft.icon when the field default defined in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/interfaces/controlpanel.py#L1012 is image/x-icon ?

Because image/x-icon is not really a thing.  The registered MIME type for Microsoft ICO format is `image/vnd.microsoft.icon`.  I can fix the default in that line if you want.

>  If the favicon is not in the settings, then I think it should return None.

Favicon data has to be returned.

> I would rather have a b64 encoded Plone logo instead of a hard-coded path in the filesystem.

Happy to.  I only read the icon file because it was already checked in.  Where should I check in the base64-encoded data?

frapell:
@Rudd-O Sorry I didn't pay more attention to this when I replied before... So, several things here:

1. The way this works, is that there's a viewlet in https://github.com/plone/plone.app.layout/blob/dc7da67ad17f6139e0c3f22b48466bef390902c2/plone/app/layout/links/viewlets.py#L39-L64 that will check if there is a favicon defined in the control panel, and based on that whether use the `/favicon` or `/favicon.ico` path
2. When getting a feed that supports icons, like an atom.xml, this method https://github.com/plone/Products.CMFPlone/blob/adadf26aecdc7e5cbd01b034b822182b257df7df/Products/CMFPlone/browser/syndication/adapters.py#L129 returns `/favicon.ico` by default
3. I presume an RSS reader, by default will try to fetch the icon from `/favicon.ico` since there is no favicon specified in the feed that Plone returns...

With all of this said, in my opinion, this is getting a bit complex and it really doesn't need to be... It could be solved by:

1. Creating a `favicon.ico` view (like you did here)
2. Copying the `favicon.ico` file from the `skins/plone_images` folder, either directly in `/browser` or in `/browser/static`
3. Have the `if...else` logic you already created and simplifying the `open()` call by having the file right there (no need to b64 encode).
4. Remove the `favicon` view, leaving only the `favicon.ico`, and change the favicon viewlet in plone.app.layout to always use `/favicon.ico`, In fact the whole python logic can be removed from there, leaving only the `favicon.pt` with the hardcoded `$portal_url/favicon.ico` as it was before.

This would allow anyone to be able to either customize the favicon by uploading an icon through the control panel, as well as to override the `favicon.ico` view from an Add-On, if so desired.

Anyway... This is my opinion, not saying it should be done this way ;) would like to hear thoughts from the other reviewers

Rudd-O:
OK.  What you are suggesting sounds like a reasonable plan of action.

Rudd-O:
Aha!  I'm now fixing this issue on this PR, almost by accident:

https://github.com/plone/Products.CMFPlone/issues/3419

Rudd-O:
The complement of the PR is now done: https://github.com/plone/plone.app.layout/pull/290

Rudd-O:

Rudd-O:
It is my pleasure to confirm that the favicon URL is now served with HTTP 200 OK both at all INavigationRoots as well as the site root.

Rudd-O:

frapell:
This LGTM, thanks @Rudd-O !

@ale-rt Do you want to have a look? part of this is in https://github.com/plone/plone.app.layout/pull/290

Rudd-O:
Still need approval.  If any of you request that two people approve, I will hold off on merging until two have.

Products/CMFPlone/browser/favicon.py
```diff
@@ -1,26 +1,44 @@
...
 from zope.component import getUtility
-from plone.memoize import ram
-
 
-class SiteFavicon(Download):
 
+class SiteFavicon(DisplayFile):
     def __init__(self, context, request):
         super().__init__(context, request)
         self.filename = None
```
ale-rt, line 15: Minor: given that you are assigning `self.filename` below anyway there is no need for this line

Products/CMFPlone/browser/favicon.py
```diff
@@ -1,26 +1,44 @@
...
         self.data = None
+        # The following attribute disables the use of an allowlist that
+        # would otherwise cause image/vnd.microsoft.icon MIMEtyped files
+        # to be served as downloads.  This allowlist list is sadly not
+        # complete, at the top of the plone.namedfile.browser.py, but
+        # fixing that is beyond the scope of this pull request.
+        self.use_denylist = True
```
ale-rt, line 22: Minor: This should be a class attribute, there is not point in assigning it every time to the same value

Products/CMFPlone/browser/favicon.py
```diff
@@ -1,26 +1,44 @@
...
 
         registry = getUtility(IRegistry)
         settings = registry.forInterface(ISiteSchema, prefix="plone")
-        if getattr(settings, 'site_favicon', False):
+        mimetype = "image/vnd.microsoft.icon"
+        filename = "favicon.ico"
```
ale-rt, line 27: Minor: these two lines should go in the else block

ale-rt (approved):
LGTM
I approve it with minor comments feel free to ignore them

Products/CMFPlone/browser/favicon.py
```diff
@@ -1,4 +1,4 @@
-from os.path import dirname
+from os.path import dirname, join
```
jensens, line 1: one import per line, alphabetical sorted please

Rudd-O: Done.  Please stamp approve when you see fit.

frapell (approved):

PR #3420, opened 2022-01-29
Removed deprecated meta_bundle from combine resources script

petschki:
Fixes removal of `logged-in` because of the deprecated `merge_with` option

### DIFF
--- a/Products/CMFPlone/resources/browser/combine.py
+++ b/Products/CMFPlone/resources/browser/combine.py
@@ -80,7 +80,7 @@ def get_resource(context, path):
     return result
 
 
-def write_js(context, folder, meta_bundle):
+def write_js(context, folder):
     registry = getUtility(IRegistry)
     resources = []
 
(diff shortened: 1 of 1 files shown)


pbauer:
Mixed messages...

When I remove the bundle-registration from `plone/app/event/profiles/default/registry.xml` the test still fail. 
Btw: like before this MR, the file only needs to hold the string `IBundleRegistry` for tests to start passing. This is enough:

```
<?xml version="1.0"?>
<registry>
  <!-- IBundleRegistry -->
</registry>
```

It also seems that removing the registration of `plone.bundles/plone-logged-in` from `plone/staticresources/profiles/default/registry/bundles.xml` no longer breaks the tests with this MR. 
Maybe @thet or @jensens can bring light to the issue.

jensens:
`write_js` and `write_css` is dead code. @thet brought it back here https://github.com/plone/Products.CMFPlone/commit/cd6f4aee687c7f16607f322948bd28815cba6dc7 because some tests where failing w/o it and it was the lazy solution to work further. But in fact it is not needed anymore and @thet his commit should be reverted and possible problems resulting from this need a real fix.

jensens (commented):
Not the way this should be solved. Close this and revert cd6f4aee687c7f16607f322948bd28815cba6dc7 instead.

PR #3445, opened 2022-02-22
1letter/issue#3031

1letter:
This PR remove two Configlets:

- the "Change Password" Configlet 
- the "Member Prefs" Configlet

Both are not really needed in the Controlpanel Overview. The Views behind the Configlets are available via Usercontrolpanel. That make more sense.
The Condition of "Change Password" Configlet had a Bug an throw an Error, another reason to remove the Panel.


1letter:

mauritsvanrees (approved):
LGTM, agreed that these should be removed.

PR #3447, opened 2022-02-25
fix initial login detection

MrTango:
If one has set login_time member properties via generic setup, this is usually set to 2000-01-01 and this value is used to detect initial login in PlonePAS.tool.memberhship.MembershipTool.loginUser() but in CMFPlone the detection compares to None which is tru is a new Plone side but if one ever has set properties via generic setup, this is never None for new users.

### DIFF
--- a/Products/CMFPlone/browser/login/password_reset.py
+++ b/Products/CMFPlone/browser/login/password_reset.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from AccessControl.SecurityManagement import getSecurityManager
+from DateTime import DateTime
 from email.header import Header
 from plone.app.layout.navigation.interfaces import INavigationRoot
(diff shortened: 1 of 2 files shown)


MrTango:

Products/CMFPlone/browser/login/password_reset.py
```diff
@@ -109,8 +110,9 @@ def _auto_login(self, userid, password):
             # with the given userid
             user = getSecurityManager().getUser()
 
-        login_time = user.getProperty('login_time', None)
-        if login_time is None:
+        default = DateTime('2000/01/01')
+        login_time = user.getProperty('login_time', default)
+        if login_time == default:
```
jensens, line 115: If I understand your PR description correctly you say both is possible, None and default time. But here you check only default time.

PR #3448, opened 2022-02-25
Fix detection of initial login time

MrTango:
If one has set login_time member properties via generic setup, this is usually set to 2000-01-01 and this value is used to detect initial login in PlonePAS.tool.memberhship.MembershipTool.loginUser() but in CMFPlone the detection compares to None which is tru is a new Plone side but if one ever has set properties via generic setup, this is never None for new users.

### DIFF
--- a/Products/CMFPlone/browser/login/password_reset.py
+++ b/Products/CMFPlone/browser/login/password_reset.py
@@ -1,4 +1,5 @@
 from AccessControl.SecurityManagement import getSecurityManager
+from DateTime import DateTime
 from email.header import Header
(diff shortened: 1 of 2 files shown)


MrTango:

jensens:
I commented already on the other PR #3447 for Plone 5.2 and my question there applies here too.

gbastien:
Hi,

I was just hit by this bug, when initial login is done by the passwordreset view (auto login after password is defined) the InitialLogin event is not fired.

Would be nice to see this in next alpha ;-)

Gauthier

Issue #3279, opened 2021-06-13
Support for modern image scales (Plone 6)


tisto:
## Abstract

Modernize the default image scales that Plone 6 uses by default.

## Motivation

Modern websites and web design heavily rely on large-scale and high-resolution images. Supporting retina image scales was a step in the right direction. Though, (Plone's current image scales)[https://github.com/plone/Products.CMFPlone/blob/fe810d633b96213cc8555099b0d4e9e585613686/Products/CMFPlone/interfaces/controlpanel.py#L1479] currently uses 768px as maximum image scale. It is fair to assume that every Plone integrator adds larger image scales for every project and that the smaller image scales are barely used.

## Assumptions

Plone Classic would gain from new default image scales as much as Volto. Therefore it makes sense to PLIP this for Plone core and not outsource that to the new Volto policy product. It would certainly be possible to use the new scales for Volto only and not to touch the Plone default scales.

## Proposal & Implementation

Add new image scales to Plone core profile:

```
huge 1600:65536
great 1200:65536
large 800:65536
preview 400:65536
mini 200:65536
thumb 128:128
tile 64:64
icon 32:32
listing 16:16
```

Source: https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6

We tried to keep the old scales (listing, icon, tile, thumb, mini, preview, large) and only slightly amend them to fit scales that can easily be calculated (this is important for srcset calculations).

We added new scales (large, great, huge) to reflect the requirements of modern image-heavy websites.

The current proposal is a compromise to keep some kind of backward compatibility. It would certainly be possible to drop all old scales and start from scratch. Though, that would possibly cause more issues when upgrading.

## Deliverables

Add new image scales to Plone core.

## Risks

* Database size might grow since we store more image scales. Since image scales are still created on the fly this risk is pretty low. If integrators do not use those image scales, there will be no difference.
* Upgrading from Plone 5 to Plone 6 and keeping the old scales only will slightly amend the scales that are created. Though, since we only slightly amended the image sizes (e.g. 768 -> 800), the risk is bearable. 
* 
## Participants

* Timo Stollenwerk
* Victor Fernandez de Alba
* Asko Soukka

datakurre:
[...]

PR #3450, opened 2022-03-04
PLIP #3279: Implement modern image scales

tisto:


tisto:

ale-rt (approved):
Thanks @tisto for taking care of this!

I would also add an upgrade step to add the **new** scales if missing.
Not sure if we should edit the existing ones.
Also some documentation about the scales might be added somewhere.

Please also wait for the approval of some people from the classic team (CC @agitator @MrTango @petschki @thet)

yurj:
Hi!
 can we just empty the default and let the profile handle it? @ale-rt introduced the upgrade steps need. So there would be an upgrade steps that delete "large", "preview" and "mini" when upgrading from 5.2 and add the new values. For 6.0, I would empty the default and have a profile with this new values. Hardcoded defaults (and they will change in the future) are hard to maintain, as we've seen, with a lot of work and discussion just to change a single value.

This values should be a config, as many other values in the control panel.

tisto:
I am currently looking into the failing unit tests in plone.restapi. My time is rather limited these days and I would highly appreciate it if someone (@yurj @ale-rt @fredvd?) could help with the upgrade steps.

fredvd:
@tisto @mauritsvanrees  I'm not sure if we need to provide an upgrade step here.  New Volto or Classic sites will get the proper default values when they are created.  integrators can and should overwrite the image scales with a GS profile. 

And when you upgrade A Plone 5.2 python3 Classic site to Plone 6.0 using in place migration, you'll need to look carefully at your probably customised theme and the impact here anyway. 

large will move from 768 width to 800.  But also all image heights in richtext area's that were bound to the max height will now scale according to the aspect ratio and could become much higher. Which will be most visible by portrait images. If you had a 400x800 image bound by the 400:400 scale, the image in the site would become 200x400 width. That's something you'll need to discuss with site users/owners about what to do here.

I don't think the impact is that big, but it will need a manual intervention anyway.  Better be explicit about it in the upgrade notes and not change/break things silently by rewriting the image scales in that case

fredvd:
@tisto Failing tests in plone.restapi are 2 tests checking hardcoded against the 'existing' image scales in a default Plone5.2 sites.  

* test_namedimage_field_serialization_returns_dict_with_original_scale
* test_namedblobimage_field_serialization_returns_dict_with_original_scale

Maybe those tests should be rewritten to first insert a known image scale set and then test for only those scales coming back. Or query the image scale registry  field and do some basic math to calculate which scales at which sizes should be returned. 

--------

* https://github.com/plone/plone.restapi/blob/96e0bc119ecb847112b1310860b43e5c86cd5a54/src/plone/restapi/tests/test_dxfield_serializer.py#L376-L445

* https://github.com/plone/plone.restapi/blob/96e0bc119ecb847112b1310860b43e5c86cd5a54/src/plone/restapi/tests/test_dxfield_serializer.py#L504-L544

ale-rt (approved):

Issue #3273, opened 2021-05-12
update WCAG info on our accessibility-info default content


tkimnguyen:
https://demo.plone.org/en/accessibility-info currently says WCAG v2.0 twice. That should be updated to 2.1. See also https://github.com/plone/Products.CMFPlone/issues/2408

Note: https://www.w3.org/WAI/standards-guidelines/wcag/#versions "WCAG 2.2 is scheduled to be published in 2021"

mohitbalwani:
Can I work on this issue?

marwanatef2:
@tkimnguyen I think this issue should be closed? Or what do you think?

tkimnguyen:
thank you @marwanatef2 !

PR #3452, opened 2022-03-10
updated WCAG info on our accessibility-info default content (Issue #3273)

Hrittik20:


mauritsvanrees (approved):
Thanks.
Technically you should fill in the Plone contributor agreement. But as release manager I happily accept a small textual fix without it.

(Note: this was also done on the 5.2.x branch in https://github.com/plone/Products.CMFPlone/pull/3479.)

Issue #3453, opened 2022-03-11
Update the ADDON_LIST of core profiles with upgrades


mauritsvanrees:
See also [my comment in `Products.PlonePAS`](https://github.com/plone/Products.PlonePAS/pull/66#issuecomment-1064889122).

In the `MigrationTool` we have a [list of add-ons](https://github.com/plone/Products.CMFPlone/blob/6.0.0a3/Products/CMFPlone/MigrationTool.py#L90). At the end of running `@@plone-upgrade`, all [profiles in this list get upgraded](https://github.com/plone/Products.CMFPlone/blob/2a1eb791eb94628a22e1f5deacd2fd44d01a63aa/Products/CMFPlone/MigrationTool.py#L325).

In my opinion, all core packages that provide their own upgrade steps, should be in this list. That is what I intended it for. The list is missing a few profiles now. At least `Products.PlonePAS` and `plone.volto`.

I can have a look at updating it. It is on my list to have a look at anyway, for this issue: https://github.com/plone/Products.CMFPlone/issues/3346.

rpatterson:
Good catch!  I started digging into this, and now I'm at the bottom of a rabbit hole.  Oh wait, I still haven't hit bottom.  Expect a missive from me tomorrow.  :-/

rpatterson:
> Good catch!

Actually now I kinda resent you for motivating me to read this code.  ;-)  It seems like there may be several messes here.

[...]

PR #3454, opened 2022-03-11
fix(upgrade): Recursive dependency profiles

rpatterson:
Proof of concept for doing away with custom/copied/duplicated `Products.GenericSetup`
logic and delegating both import and upgrade behavior to profile dependency definitions
in the profile metadata.  See the individual commit messages for discussion of each
discreet change.

Refs #3453


rpatterson:
Note that these changes are intended to demonstrate this approach.  They be used as the
basis for the real changes, but the impact of the changed should be reviewed by someone
familiar with the reasons for the original approach, tests should be added and resulting
issues resolved.

rpatterson:
Oh, forgot to specify that I will *not* have time to do this work.  I just wanted to capture it in case someone else wanted to pick this up.

rpatterson:
Closing just to clean up my user PR list, feel free to re-open if it's actually going to be used or built upon.

PR #3456, opened 2022-03-15
[Master] Fix test after PortalTransforms uses lxml html method to serialize (instead of xml method).

gotcha:


mauritsvanrees (commented):
Why is this needed? I do not see failing tests on Jenkins before this.
I do not see how both the old and new test can pass.

gotcha:
Failure happens only if https://github.com/plone/Products.PortalTransforms/pull/43 is merged.

I did not question the validity of the test.

mauritsvanrees:
Ah, understood.

mauritsvanrees (approved):

PR #3470, opened 2022-03-29
Fixed plone.session resource, and evaluating expressions on resources

mauritsvanrees:
Fixes https://github.com/plone/plone.session/issues/23.
Main problem there was that the code could not traverse to the plone.session css resource, resulting in a resource None, which caused a traceback, resulting in an unstyled site.

After fixing this part, I found another problem:
the `plone.session` css resource had an expression `python: member is not None`, but was still loaded for anonymous users as well.
It turned out that the expression was never checked. More correctly said: it looks like the expression of the last resource was used for all resources, even when this expression was empty.
To fix this, I had to pass the expression to the webresource, and move some code from `browser/resource.py` to `webresource.py`.

(I should probably say "bundle" wherever I say "resource", but you get the idea.)

I added tests with a couple of resource bundles with different expressions.

### DIFF
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -1,21 +1,13 @@
 from ..webresource import PloneScriptResource
 from ..webresource import PloneStyleResource
-from Acquisition import aq_base
-from Acquisition import aq_inner
-from Acquisition import aq_parent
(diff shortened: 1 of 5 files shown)


mauritsvanrees:

ale-rt (approved):

ale-rt:
Nice catch!

PR #3489, opened 2022-04-07
Added support for images in liveSearch results.

agitator:


mauritsvanrees (approved):
Tested and it works.

mauritsvanrees:
All green.

mauritsvanrees:
I have merged all related branches.
Thanks!

yurj:
KeyError: 'Interface `plone.base.interfaces.controlpanel.ISearchSchema` defines a field `search_show_images`, for which there is no record.'
My custom theme/package, upgrading a site from alpha3 to alpha4 (buildout-coredev). I saw this:
https://github.com/plone/plone.app.upgrade/pull/285/commits/9d1a564a2723c683d7f56e8cb9f33d4f87fb1298
and then re run the upgrades by hand for 6004 (even if I upgraded the Plone site with the upgrade button in Zope). It is because it is alpha?

mauritsvanrees:
I copied a Plone 5.2 Python 3 database to 6.0.0a4, ran `@plone-upgrade`, and this was fine.

But this can happen when you had previously already applied the upgrade to 6004. For example like this:

- Start with a 6.0.0a3 site.
- You used the coredev buildout yesterday. Or you updated to https://dist.plone.org/release/6.0-dev/ yesterday. Or you can reproduce this now by adding that link to the `find-links` and pinning `plone.app.upgrade = 3.0.0a4.dev0`.
- You ran `@@plone-upgrade`. This runs the upgrades to 6004, but it is lacking my commit for the `ISearchSchema`.
- You upgrade to 6.0.0a4. `@@plone-upgrade` has nothing to do, because you are already to 6004.
- Or you upgrade to latest coredev. `@@plone-upgrade` upgrades you from 6004 to 6005, which is a dummy upgrade.

With such a scenario, you would indeed be missing this key from the search schema. Rerunning the upgrade to 6004 fixes it. Or explicitly run import step `plone.app.registry` from profile `plone.app.upgrade.v60.to6004`.

Does this explain it?
Or is something else going on? I do hope that the move of `ISearchSchema` from `Products.CMFPlone` to `plone.base` does not negatively influence this. But so far it seems good for me.

mauritsvanrees:
I had to check:

- In Plone 5.2 Py 3 I changed all fields in the search control panel.
- Copied database to Plone 6 and ran upgrade.
- The new field is there.
- The existing fields have the same values as in 5.2.

So that is good.

PR #3518, opened 2022-05-02
Fix for quoted search terms

### LINKED ISSUE #3517: searching for phrases within quotes doesn't work
the `munge_search_term` method doesn't work when searching for quoted phrases.

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/search.py

**Reproduce:**

- go to https://6-classic.demo.plone.org/en
- search for `"Welcome to Plone"`
- you get an empty search result

**Background**

the `munge_search_term` returns this for the phrase above:

`'Welcome AND to AND Plone*'`

**Solution**

keep phrases within quotes untouched.

petschki:
fixes #3517

### DIFF
--- a/Products/CMFPlone/browser/search.py
+++ b/Products/CMFPlone/browser/search.py
@@ -46,23 +46,24 @@ def quote(term):
     return term
 
 
+def munge_search_term(query):
+    for char in BAD_CHARS:
+        query = query.replace(char, ' ')
+
+    # extract quoted phrases first
+    r = re.findall(r'"[^"]*"', query)
+    for qp in r:
+        query = query.replace(qp, "")
+
+    r += map(quote, query.strip().split())
+    r = " AND ".join(r)
+    r = quote_chars(r) + ('*' if r and not r.endswith('"') else '')
+    return r
+
+
 class Search(BrowserView):
 
     valid_keys = ('sort_on', 'sort_order', 'sort_limit', 'fq', 'fl', 'facet')
-
-    def munge_search_term(self, q):
-        for char in BAD_CHARS:
-            q = q.replace(char, ' ')
-
-        # extract possibly quoted phrases first
-        r = re.findall(r'"[^"]*"', q)
-        for qp in r:
-            q = q.replace(qp, "")
-
-        r += map(quote, q.strip().split())
-        r = " AND ".join(r)
-        r = quote_chars(r) + ('*' if not r.endswith('"') else '')
-        return r
 
     def results(self, query=None, batch=True, b_size=10, b_start=0,
                 use_content_listing=True):
@@ -113,7 +114,7 @@ class Search(BrowserView):
             if v and ((k in valid_keys) or k.startswith('facet.')):
                 query[k] = v
         if text:
-            query['SearchableText'] = self.munge_search_term(text)
+            query['SearchableText'] = munge_search_term(text)
 
         # don't filter on created at all if we want all results
         created = query.get('created')
--- a/Products/CMFPlone/browser/search.py
+++ b/Products/CMFPlone/browser/search.py
@@ -16,6 +16,7 @@ from zope.publisher.browser import BrowserView
 from ZTUtils import make_query
 
 import json
+import re
 
 _ = MessageFactory('plone')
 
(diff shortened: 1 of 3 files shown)


petschki:

mauritsvanrees (changes requested):
Thanks!
Could you add some tests for `munge_search_term` please? It looks like it may be doing the right thing, but it is hard to say from the current tests: they only check how many results there are.
`munge_search_term` could actually be a standalone function. Could be useful for third party code that needs this.
I guess it could also be a `classmethod` and/or `staticmethod`. I can never remember the difference. :-) So feel free to ignore that.

petschki:
@mauritsvanrees I also thought about refactoring as a standalone function so addons coul make `from Products.CMFPlone.browser.search import munge_search_term` and reuse it ... or should we even place it in `Products.CMFPlone.utils` ?

mauritsvanrees:
> @mauritsvanrees I also thought about refactoring as a standalone function so addons coul make `from Products.CMFPlone.browser.search import munge_search_term` and reuse it ... or should we even place it in `Products.CMFPlone.utils` ?

Or `plone.base.utils`? cc @jensens

jensens:
Where else do we need this? It could go into plone.base.utils, since it has zero dependencies. But I would like to avoid adding there random stuff not to be reused. It spreads then logic over multiple packages and makes readability worse.

petschki:
I also would not move it to plone.base.utils. the munge_search_term logic belongs to the search view and therefore the right place is in Products.CMFPlone.browser.search for me ...

petschki:

petschki:
found a duplicate which could be resolved with plone.base.utils? see https://github.com/plone/plone.app.querystring/blob/master/plone/app/querystring/querybuilder.py#L199

mauritsvanrees (approved):

jensens (approved):

yurj (changes requested):

petschki:

petschki:
One more thing ...

should we add `*` to every word in an unquoted term? so `welcome to plone` would get `welcome* to* plone*` ...

@yurj @mauritsvanrees @agitator @jensens

yurj:
> One more thing ...
> 
> should we add `*` to every word in an unquoted term? so `welcome to plone` would get `welcome* to* plone*` ...
> 
> @yurj @mauritsvanrees @agitator @jensens

Some stop words should be escluded, otherwise ZCTextIndex would search for 'to*' instead of removing it. Beware that the live search and the search clicking on "Search" can differ. For example, on https://6-classic.demo.plone.org/en type "to". Zero results on live search, while clicking on search you will find 5 items...

petschki:
> > One more thing ...
> > should we add `*` to every word in an unquoted term? so `welcome to plone` would get `welcome* to* plone*` ...
> > @yurj @mauritsvanrees @agitator @jensens
> 
> Some stop words should be escluded, otherwise ZCTextIndex would search for 'to*' instead of removing it. Beware that the live search and the search clicking on "Search" can differ. For example, on https://6-classic.demo.plone.org/en type "to". Zero results on live search, while clicking on search you will find 5 items...

there is no difference in generating the search query for `@@search` and `@@ajax_search` both use the same logic (https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/search.py#L239). This difference is because the livesearch only starts when at least 4 characters are given ... results are (and should be) the same.

agitator:
Are there existing stop word lists to build on? 
Not sure if a minimal character approach would work in a multilingual context.

petschki:
whoa ... I've opened pandoras box here 😅

multilingual is a good point. but if I read the docs here correctly, this is done by the ZCTextIndex Lexicon: https://zope.readthedocs.io/en/latest/zopebook/SearchingZCatalog.html#lexicons

so it would be interesting if the `to` in `Welcome to Plone` is removed by the index lexicon ...

petschki:

yurj:
I think this can be merged, can improved later

yurj (approved):

petschki:

petschki:

PR #3477, opened 2022-04-01
Add image srcset's configuration to TinyMCE pattern settings

MrTango:
All PR's:

- https://github.com/plone/Products.CMFPlone/pull/3477
- https://github.com/plone/plone.base/pull/5
- https://github.com/plone/mockup/pull/1156
- https://github.com/plone/plone.outputfilters/pull/49

needs an upgrade step in plone.app.upgrades to load the config

PR #3521, opened 2022-05-04
[WIP] Image scales metadata

cekk:
As discussed today at Beethoven sprint.

We need a way to pass image scales to Volto without waking up objects and calculating them every time.

This is a poc to see if we can store them in catalog.
We store a PersistentDict into metadata so it is only loaded from db if accessed.

There are two new adapters:
- for context (IImageScalesAdapter)
- for fields (IImageScalesFieldsAdapter)

The first one iterate over context's schema and try to call the second over schema fields.
If there is an adapter registered for that field, it can return the proper scales.
This allows future customizations like for example if miniatures are stored in an external service and not in Plone.

I copied the code for getting scales from plone.restapi utils (we could move that code from restapi to here if needed).


Products/CMFPlone/image_scales/adapters.py
```diff
@@ -0,0 +1,158 @@
...
+        scales = self.get_scales(self.field, width, height)
+        return {
+            "filename": image.filename,
+            "content-type": image.contentType,
+            "size": image.getSize(),
+            "download": url,
+            "width": width,
+            "height": height,
+            "scales": scales,
+        }
```
jensens, line 70: We want to have a list here (think of multi-image fields, like a relation to images), right?

Products/CMFPlone/image_scales/indexer.py
```diff
@@ -0,0 +1,19 @@
...
+from zope.component import queryMultiAdapter
+from zope.globalrequest import getRequest
+
+
+@indexer(IDexterityContent)
+def image_scales(obj):
+    """
+    Indexer used to store in metadata the image scales of the object.
+    """
+    adapter = queryMultiAdapter((aq_base(obj), getRequest()), IImageScalesAdapter)
```
jensens, line 15: Why aq_base here?

cekk: i was thinking it was a good idea "isolate" the context, but probably not. removed it

Products/CMFPlone/image_scales/adapters.py
cekk: Yes, sorry. Now it returns a list.
Should i add additional adapters in base implementation or not? Because multi-image fields are related to custom implementations.

jensens: We could do so for a RelationList, not sure if we want to have it in Core?

cekk:
> Is this implying that we generate all the scales whenever a new image is uploaded?

isn't this already made when we create a new content on Volto? When you get its data from restapi, you get also its miniatures.

cekk:
@ale-rt @plone/framework-team if this is ok for you and you don't have any complaints, i will split that pr into the right packages

cekk:
@jensens this pr needs also an upgrade-step because we're adding a new metadata to the catalog and catalog needs to be updated.
That could be a "problem" in large installations because for metadata we need to reindex the whole objects, not only the index and could take time.
Is it ok to do it in a Plone upgrade-step?

ale-rt:
> > Is this implying that we generate all the scales whenever a new image is uploaded?
> 
> isn't this already made when we create a new content on Volto? When you get its data from restapi, you get also its miniatures.

IIRC correctly @datakurre made a PR to some package that enables that behavior by setting some env variable, but with a quick search I was not able to find it. 

Anyway this is not the general case and I would like to keep it like that.
The big argument against creating all the scales immediately is that we could upload N images at a time and block the site while computing the scales.

Also I think that the upgrade step that adds the `image_scales` index lead to a lot of undesired effects.

I am not deeply in to your use case and this might be the perfect solution, but I would ask you to think more about this.

datakurre:
https://github.com/plone/plone.formwidget.namedfile/blob/e1a51ce119c1912f431aaeecc498153b352a5414/plone/formwidget/namedfile/datamanager.py#L20

cekk:
Ok, i've found some problems with this implementation.
The first one is that when i create a new object with image, or i update the image with a new one, when the indexer is called i don't have the final blob file in content-type field, but a temporary file.
That means that when the object is indexed, we generate a wrong set of scales that will be invalidated at first object GET call.
The catalog keeps the old ones, and restapi call returns the right ones generated from the proper blob.
When i save again the content without touching the image, everything has fixed.

Probably adding two event handlers that listen to object added and modified events can have the right blobs, but i need a way to check into my indexer (actually into the field adapter) if the current image is temporary or not.

@datakurre @ale-rt  any suggestions?

I don't know if this is the perfect solution also because we have to re-index the catalog.

Let me better explain our use-case: when dealing with listing blocks (and other blocks) we could have a list of results with images.
With listing blocks (and search endpoint too) we can choose to get the fullobjects or the brains with a custom set of metadata (catalog metadata).
To speedup the queries and not have too big responses we decided to get brains by default. Most blocks needs images infos. For frontend optimizations, we need the full list of available scales for each image and their width and height.
We can get these data from the real object (that means waking up each brain) or from a metadata stored in catalog.

An alternative solution that we discussed was to create a separate catalog (maybe just a BTree into the site root) that we are going to populate with the same data that i'm trying to store in metadata, and when we serialize the brains, we'll get scales infos from there without touching real objects.

ale-rt:
I do not know if I understand completely your issue and your "frontend optimizations".

When working with a brain usually the URL to access the scale looks like: `${brain/getURL}/@@images/${scale}/?t=${brain/modified}`.

Knowing the scale you are pretty sure to know the width of the image but not the height given that usually the scales size look like `$WIDTH:65536`.

You already see that puts emphasis on the width rather than the height and in my opinion this is a good thing.
The two dimensions are not exactly equal when we talk about rendering a page: the vertical one is much more delicate and browser dependent.

That said, with a brain in your hands there are good chances that you already know what is the with width of your scale and the URL to display it.

That should be enough info to build a web page that can be rendered nicely across different browsers.

If your "frontend optimizations" have desperate need to know the height of an image I strongly doubt they are fit for the Plone core.
Can you reconsider them?

pnicolli:
@ale-rt A couple explanations are probably due here:

- the height of the image is required for properly rendering it, the browser uses it to reserve the image space on the page, thus avoiding Cumulative Layout Shift
- in addition to that, in Volto we don't have access to the scales configured in the backend, at this time, so an alternative to have those would need to be implemented

datakurre:
@cekk Ouch. The rabbit hole gets deeper.

A think that has troubled me forever is that scale uids are not deterministic, but just random uuid4s: https://github.com/plone/plone.scale/blob/09b3598b3843363260a263c65435bd2f463249ab/plone/scale/storage.py#L198

If were able to use deterministic uids, this should be among the issues that get fixed. Like hash of the original file by its contents and full scale configuration.

ale-rt:
> @cekk Ouch. The rabbit hole gets deeper.
> 
> A thing that has troubled me forever is that scale uids are not deterministic, but just random uuid4s: https://github.com/plone/plone.scale/blob/09b3598b3843363260a263c65435bd2f463249ab/plone/scale/storage.py#L198
> 
> If were able to use deterministic uids, this should be among the issues that get fixed. Like hash of the original file by its contents and full scale configuration.
> 

This is super true, having the possibility to get the proper URL out of some brain metadata would be really awesome.

About the hashing, I think it is enough to consider the `_p_modified` timestamp of the blob rather than hashing the full file content which can be several MB.

ale-rt:
> @ale-rt A couple explanations are probably due here:
> 
> * the height of the image is required for properly rendering it, the browser uses it to reserve the image space on the page, thus avoiding Cumulative Layout Shift
> * in addition to that, in Volto we don't have access to the scales configured in the backend, at this time, so an alternative to have those would need to be implemented

1. I do not see it **required** for properly rendering. I grok it is a plus but it is not required by any means.
2. That seems to me a really easy thing to solve. I guess you already have some constants about the site already defined somewhere (e.g. the `portal_url`). That would be just another one.

ale-rt:
> Knowing the scale you are pretty sure to know the width of the image but not the height given that usually the scales size look like `$WIDTH:65536`.
> 

And if you want to know the height given the width you can just get it knowing the original images and doing a proportion or if you prefer you can divide it by the aspect ratio

cekk:
Ok, so we could get rid of real scales when storing the data (also because i suspect that if we call the view on temp files when we are uploading new image, we'll get blobs twice).

I think that we should store something in the catalog anyway, because we need to know at least original image sizes.
Then we could manually calculate different scales sizes based on what @ale-rt suggested (width from the registry and height calculated).
And generate urls without blobs UIDS but appending modified date to invalidate cache.

The big question now is: **WHERE** and **HOW**? In catalog as proposed? In a separate tree?
In both of the options we need to make a "reindex" of the objects to populate new data.
If this is not only a Volto need, we should made it in the core.

An alternative solution could be: in summary serializer, if you are requesting an image-ish field (we need to define a way to set that list) with additional_metadata, we wake up the brain and get proper scales (as we already do when serializing the object).
Probably this will slow down a bit query performances, but doesn't add any extra data in the catalog. And we're talking about  batched results, so usually we have ate least 20/30 brains each query.

yurj:
> The first one is that when i create a new object with image, or i update the image with a new one, when the indexer is called i don't have the final blob file in content-type field, but a temporary file.

Hi! this is a bug, because when indexing the object on save, the object should already be finalized. I think it has to do with the uncommit state of the blob:
https://github.com/zopefoundation/ZODB/blob/22d1405d413a0c210ecbf43575d6bb15b34ecdc9/src/ZODB/tests/blob_transaction.txt

can you check if the `committed` call over the blob fails like in https://github.com/zopefoundation/ZODB/blob/22d1405d413a0c210ecbf43575d6bb15b34ecdc9/src/ZODB/tests/blob_transaction.txt#L321?

Also, @1letter had this:

```python
# first attempt
(Pdb++) pp blob._p_blob_committed 
None

(Pdb++) dir(blob)
['__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__getstate__', '__hash__', '__implemented__', '__init__', '__module__', '__new__', '__providedBy__', '__provides__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_create_uncommitted_file', '_p_activate', '_p_blob_committed', '_p_blob_ref', '_p_blob_uncommitted', '_p_changed', '_p_deactivate', '_p_delattr', '_p_estimated_size', '_p_getattr', '_p_invalidate', '_p_jar', '_p_mtime', '_p_oid', '_p_serial', '_p_setattr', '_p_state', '_p_status', '_p_sticky', '_uncommitted', 'closed', 'committed', 'consumeFile', 'open', 'opened', 'readers', 'writers']

# second attempt
(Pdb++) pp blob._p_blob_committed 
u'/Plone5/py2/var/blobstorage/0x00/0x00/0x00/0x00/0x00/0xaa/0x68/0xbc/0x03d9906455146bcc.blob'
```
He solved with:

```python
blobfile = obj.file
blob = blobfile._blob
blob._p_activate()
```

Hope this can help.

pnicolli:
> > @ale-rt A couple explanations are probably due here:
> > 
> > * the height of the image is required for properly rendering it, the browser uses it to reserve the image space on the page, thus avoiding Cumulative Layout Shift
> > * in addition to that, in Volto we don't have access to the scales configured in the backend, at this time, so an alternative to have those would need to be implemented
> 
> 1. I do not see it **required** for properly rendering. I grok it is a plus but it is not required by any means.
> 2. That seems to me a really easy thing to solve. I guess you already have some constants about the site already defined somewhere (e.g. the `portal_url`). That would be just another one.

@ale-rt 
1. It **is** required if we want good scores on core web vitals. I realize it also looks good without, but we should want our frontend to be the best possible one in order to have a solid product (Plone itself). Also, several customers tend to look at the lighthouse score more than they look at the actual look of the site.
Good source of info on the topic: https://web.dev/optimize-cls/#images-without-dimensions

2. We could indeed save the Plone scales in Volto config, but then we would need to manually keep those in sync with changes in the backend config, in the actual scales calculated by Plone. Of course, this requirement is especially valid if we get width and height of images from the catalog like suggested in the PR, because then the code sync problem can bite us anytime.

@cekk 
Regarding where and how to generate image urls without uuids like ale-rt suggested, which is fine to me, I would suggest doing it directly in the frontend, that means in Volto and in Classic UI templates. The info to build the url are there, it's just a matter of using those. This is true unless we decide that we want the backend to decide how the url is built and the frontend to just use the one provided by the backend, and this option is not to be discarded probably, since it is probably more future proof.

cekk:
@pnicolli we need to generate them in the backend because otherwise in brains you don't know what are the actual image sizes, and you need both if i understood correctly.

pnicolli:
@cekk I just meant the url of the image, which in the frontend could be generated like `{brain.getURL}/@@images/{brain.image_field}/preview?modified={brain.modified}`. Of course we would need to assume `preview` exists instead of knowing it from the catalog, but I believe there is no escaping from this.
Of course we still need image width and height from the backend.

fredvd:
From what I learned and proposed for last friday at the Beethovensprint my current understanding is that the most useful and probably only thing to store in the catalog and what is needed to create the correct image sizes in the srcset is the width of the original image stored on the plone.namedfile field or of the images available on a context. 

What you don't want to do is to generate a srcset with for example image widths  (300w, 600w, 1200w, 1600w) when the stored image is only 980 pixels wide. 

All other information can be generated independent of waking up the image object. And with the current knowledge for every site there will be a limited identical set of 'fixed' image widths that can be used for all images that are served and rendered in the frontend, either volto or classic. 

So then you could filter the image widths  that are not available (for example for hidpi usage) because it would mean the image gets upscaled when that image is pulled from the srcset.

fredvd:
> A thing that has troubled me forever is that scale uids are not deterministic, but just random uuid4s: https://github.com/plone/plone.scale/blob/09b3598b3843363260a263c65435bd2f463249ab/plone/scale/storage.py#L198
> 
> If were able to use deterministic uids, this should be among the issues that get fixed. Like hash of the original file by its contents and full scale configuration.
> 
> (I have had a similar patch for that for JYU for years, but don't have access to the code until next year. We did it for better caching, because the builtin 24h scale invalidation on any scale modification with continuously changing scale URLs was bad.)

One problem that we never had to deal with, but became relevant with plone.restapi and decoupled frontend is opening up the server for Denial of Service attacks by requestiong thousands of flexible image scales.  When the templates are processed only server side and a uuid url is generated this is not possible. 

Some stand alone image servers that were looked at for inspiration or integration (also in 2020 at the Dresden Plonetagung) allow to 'sign' a image request with public/private key, effectively generating a signed 'ticket' with which the frontend is allowed to request a certain image scale/format/quality/crop . 

@datakurre  I'm not saying you are proposing this: you are only discussing about deterministic scales, but if one more step would be  to generate the image based on decoding the information in the url without any security checks we need to be aware of DoS. 

----

I'm surprised to learn that scales are 24h lived, I'd expect they are there permanently until the image data is refreshed or in the current implementation the scale definitions is altered.  It was  actually because of stale image scale annotations that we run a purge image scales script on Plone sites   (https://github.com/zestsoftware/plonescripts/blob/master/purge_image_scales.py)

----

What I missed in a solution for problems we encountered the last few years with plone.restapi image responsse triggering generation of all image scales at once because returning the 'stable' url with uuid needs the generation of the image: 

why did nobody until now propose to make the generation of the image scaled to a certain width lazy by:
* registering the uuid in an servide side dict/list with the requested processing parameters so it is 'pre registered' for creation when restapi or a page template needs the width(s). 
* generate the image scale upon the first incoming request for this uuid when it is pre-registered and then cache it on an annotation

This avoid rendering scales that are not requested but still provide some safety measure against DoS.

datakurre:
> @datakurre I'm not saying you are proposing this: you are only discussing about deterministic scales, but if one more step would be to generate the image based on decoding the information in the url without any security checks we need to be aware of DoS.

I don't propose decoding scale from the URL. Only that the scale URL remains the same as long as the input and parameters are the same.

In this thread is would fix the weird blob related issues.

Else where it would fix that when any update to scales after 24h clears old scales, the new scales would have the same URLs.

MrTango:
@jensens and I where discussing this also today and came to the same conclusion, that we would like to generate the URL in the front-end. Because than the browser will only trigger scaling for individual images and only for one scale at a time.
To archive this we where thinking about having a timestamp for each scale on the object and use it in the URL. Other than that the URL will be similar to what we had without the UID-based version.

@@images/image/TIMESTAMP/large

I also did some testing earlier today, to research what information we need.
We need the one pair of width/height for example from the original image when it's uploaded. 
Would be also good to have that in the catalog metadata. 
With this we have the original aspect ratio. 
For the srcset to be generated on client side, we only need to put one pair, it could be even something like width="3" and height="2". The browser will adjust this to the real width for us, if height="auto" is set.

If we later set the width in CSS, that is fine.
It also works well together with max-with="100%".
The Browser will keep updating it's calculation and loads the fitting image from the list we provided.
It works well with "px" or "vw" on images, but not with "%", as the later will make the image jump around the others are not.

datakurre:
> why did nobody until now propose to make the generation of the image scaled to a certain width lazy by:

I almost did this two years ago. Only that images were generated immediately in parallel threads and not on-demand and meanwhile the stable URLs would return redirection to the original.

On-demand generation would imply write on read. That's probably why I did not do it.

jensens:
> On-demand generation would imply write on read. That's probably why I did not do it.

At the moment we have write-on-read already. The image-scales are not created on save, but on the first page demanding the scales, which can be an anonymous access.

datakurre:
Ping me is there’s interest to revive https://github.com/plone/plone.namedfile/tree/datakurre-image-scaling-queue (as such or with some on-demand-changes) before rewriting everything to support external scaling services 🤷‍♂️ (We’ve been running that in production for a couple of years without anyone noticing it is there.)

fredvd:
I'd like to repeat something @pnicolli mentionned this morning during the Volto-meeting:  Let's keep the actual requirements in mind (and work from there).

I think wat we want to achieve in Plone 6.0 is a better content editor experience: how they place images where in the content (pages, mosaic layouts, volto block layout) and a better 'designer'/'integrator' experience so that you can decide per project what to limit what editors can do with images (limited number of aspect ratio's in a project based on context, sizes, layout options). 

What would be bad if a developer needs to look at the image handling code in his frontend or backend code when upgrading from Plone 6.0 to Plone 6.1. 

What would be even worse and very bad is if a Plone 6.1 would change the default Image placement control panel for End Users.   

I'd love to have the option for an external source image source external to the ZODB, the optimal img srcset calculations  on the backend/frontend, stable image size links, etc. etc. but with the limited time we have, can we maybe agree that whatever we do, we should not force integrators/developers to change the default UX for editors between Plone 6.0 and 6.X ?

(Unless they override the handling already with the hooks we provide in Plone 6.0 already in their projects customisations, but then also their customisations also shouldn't heavily break before Plone 7 or 8 appears)

wesleybl:
> If the client can generate the srcset url's without triggering scales creation, the browser will load only some of them and with lazy loading even less.

We are thinking of creating a behavior with 3 images, to use in srcset. Is this a bad idea? I see you talk about scales, but generally the mobile image has to be different from the desktop image.

MrTango:
@wesleybl 

> > If the client can generate the srcset url's without triggering scales creation, the browser will load only some of them and with lazy loading even less.
> 
> We are thinking of creating a behavior with 3 images, to use in srcset. Is this a bad idea? I see you talk about scales, but generally the mobile image has to be different from the desktop image.

have a look at:
https://github.com/plone/Products.CMFPlone/pull/3477
https://github.com/plone/plone.base/pull/5

you can test it via plip file in buildout.coredev
https://github.com/plone/buildout.coredev/blob/6.0/plips/plip-image-srcsets.cfg

Currently the following is supported:
- srcset list of allowed scales, the browser can choose from (image optimization)
- multiple source tags with media queries to deliver different images based a viewport size (art direction)

raw documentation here:
https://deploy-preview-1252--6-dev-docs-plone-org.netlify.app/classic-ui/images.html#srcset-configuration

let me know if you need anything else.

sneridagh:
I completely agree with @mauritsvanrees point of view. This is a nice to have and quite a hot issue, but still in the drawing board and a lot of discussions need to be done and decisions to be made. Let's not jeopardise the release because of this. There is always a "this would be nice to be included" in store... we should have a realistic vision now.

However, it's a hot topic, and need to be addressed, and yes, now that we are all on it, stick next to the drawing board until we figure out how to do that, and implement a PoC then iterate over it. +1 to create an add-on. This will allow us to use it in old versions as well, not only in Plone 6. 

I'm not going to enter in any tech discussion, I have enough food in my plate! However I suggest that we explore @datakurre PoC and continue from there. Also, I have serious doubts about when the scales are _really_ generated.

I'd say we need a champion for this as well, so the current efforts keeps momentum going? I'd say that it's a perfect topic for Buschenschanksprint ! Any taker?

It seems to me that once we get the gist and we decide how to proceed, the implementation should not be that difficult.

cekk:
+1 for the separate experimental package and proper plip basically because probably we are going to do a massive catalog reindex or filling a separate tool.

I tried some (maybe stupid) perf tests with restapi endpoint (querystring-search) to understand if could be acceptable going back using fullobjects when we need images. I've also tried to implement the "additional metadata" logic also for fullobjects (this means returning less not needed data like blocks for example), but response times are still longer than dealing with brains also for small batches:

- query with _fullobjects=false_, _b_size=200_ and _all metadata_: ~ **0.17ms** and **647.1Kb** size
- query with _fullobjects=false_, _b_size=200_ and only _default metadata_: ~ **0.04ms** and **95.5Kb** size
- query with _fullobjects=true_, _b_size=200_ and _all fields_: ~ **2.7ms** and **2.9Mb** size
- query with _fullobjects=true_, _b_size=200_ and _filtered fields_ (default ones + image scales) : ~ **1.7ms** and **890Kb** size

- query with _fullobjects=false_, _b_size=50_ and _all metadata_: ~ **0.06ms** and **166Kb** size
- query with _fullobjects=false_, _b_size=50_ and only _default metadata_: ~ **0.016ms** and **25Kb** size
- query with _fullobjects=true_, _b_size=50_ and _all fields_: ~ **0.7ms** and **667Kb** size
- query with _fullobjects=true_, _b_size=50_ and _filtered fields_ (default ones + image scales) : ~ **0.4ms** and **230Kb** size

- query with _fullobjects=false_, _b_size=10_ and _all metadata_: ~ **0.02ms** and **34Kb** size
- query with _fullobjects=false_, _b_size=10_ and only _default metadata_: ~ **0.012ms** and **5.4Kb** size
- query with _fullobjects=true_, _b_size=10_ and _all fields_: ~ **0.15ms** and **157Kb** size
- query with _fullobjects=true_, _b_size=10_ and _filtered fields_ (default ones + image scales) : ~ **0.1ms** and **48Kb** size

I also tried to get fullobjects with filtered fields without images (adding description field for example):

- query with _fullobjects=true_, _b_size=10_ and filtered fields (_default ones + description_) : ~ **0.05ms** and **39Kb** size
- query with _fullobjects=true_, _b_size=10_ and filtered fields (_default ones + image scales_) : ~ **0.1ms** and **48Kb** size

It seems (as supposed) that scaled generation took several time that can cause slow requests with fullobjects.
If one of your proposal can speedup this call, we could also get rid of storing additional data to the catalog or using an external storage, and maybe say: if you need images in your block (or listing template), you can fetch fullobjects, otherwise not.

@MrTango @jensens @ale-rt let me know if you're going to discuss about that during the sprint.

ale-rt:
We did discuss the topic at the sprint and we identified some way forward... stay tuned

sneridagh:
@ale-rt guys you are having us in Suspense (no React pun intended) :D

Any update/development?

mauritsvanrees:
I add a News Item in Classic. Iadd an image. With restapi I ask for the news listing with `image_scales` as extra metadata. Url is: `http://localhost:8080/Plone/++api++/news/?metadata_fields=image_scales`. Result: the image scales are included as dictionary. So success!

mauritsvanrees:
Sample dictionary of `image_scales`:

```
{
    "image": [
        {
            "filename": "dierproeven.jpg",
            "content-type": "image/jpeg",
            "size": 14939,
            "download": "http://localhost:8080/Plone/news/news-item-1/@@images/image-395-c209c546deec3eaad5c1a0d94fe39a28.jpeg",
            "width": 395,
            "height": 300,
            "scales": {
                "mini": {
                    "download": "http://localhost:8080/Plone/news/news-item-1/@@images/image-200-f31a54fb13dfd9cfa20b7d9a91f97d69.jpeg",
                    "width": 200,
                    "height": 151,
                },
                "thumb": {
                    "download": "http://localhost:8080/Plone/news/news-item-1/@@images/image-128-5e4fa67362c36dad14730ad9dd35b397.jpeg",
                    "width": 128,
                    "height": 97,
                },
                "tile": {
                    "download": "http://localhost:8080/Plone/news/news-item-1/@@images/image-64-64be18a12a086a60cd62feeb846c748b.jpeg",
                    "width": 64,
                    "height": 48,
                },
                "icon": {
                    "download": "http://localhost:8080/Plone/news/news-item-1/@@images/image-32-0bb98da006620ef36361a910aeaf95b0.jpeg",
                    "width": 32,
                    "height": 24,
                },
                "listing": {
                    "download": "http://localhost:8080/Plone/news/news-item-1/@@images/image-16-c879d46734689d31b8338ddee99a772c.jpeg",
                    "width": 16,
                    "height": 12,
                },
            },
        }
    ]
}
```

mauritsvanrees:
> @mauritsvanrees Thanks for the wrapup of the latest work on this PR. I am not sure I entirely understand what the next steps are, in order to finish this work after Buschenschanksprint. Can we maybe make some bulleted list or some recap of what's missing?

Tests need to pass. One PR that could be independently merged for this, is this one in `plone.restapi` once green:
https://github.com/plone/plone.restapi/pull/1433

There might be some more test failures left in other packages. The robot tests are not very stable the past few weeks, especially with high load on Jenkins, so it is hard to say which packages still need test fixes that are related to these changes.

I think the easiest order is:

- [ ] Merge https://github.com/plone/plone.restapi/pull/1433
- [ ] Test the `pre-scale` branches of `plone.scale` and `plone.namedfile` again, and fix any remaining test failures.
- [ ] Merge those two branches.

Then we can independently test/review/merge the two features that depend on this:

- [ ] This CMFPlone PR for the image_scales metadata.
- [ ] Maik's work in https://github.com/plone/Products.CMFPlone/pull/3477 and [related PRs](https://github.com/plone/buildout.coredev/blob/6.0/plips/plip-image-srcsets.cfg).

> In addition to this, in order to use the latest version of this work, as of today, the correct way is to use this branch plus the `pre-scale` branches of `plone.scale` and `plone.namedfile`, am I right?

Correct.

> One last question about the write-on-read situation that was discussed. Is it the default behavior in the branches mentioned above? Is there a way to opt-out of the write-on-read behavior?

No, there is no opt out.

I guess you could add an event handler on added/modified of an Image, News Item, etc. Then get the `@@images` view on the context and call the `scale` method on it with all scale names that you want. By default this passes `pre=False`, which gets you the old behavior.

You could also add an environment variable and hook that into the indexer in this PR: by default call the scale method with `pre=True`, but do not do this when an environment variable is set.

pnicolli:
> > No, there is no opt out.
> 
> what is with the environment parameter Volto people are using already? That still exists right? So nothing would change there. So if that flag is set, scales are generated directly on modify and there shouldn't be any write on read anymore. But to be clear, the PR's are not intruducing write on read, we have this behavior for a long time in Plone. With the current optimizations it will be less of a problem and for big sites one can go the prerender way with the flag.

Yeah I guess you're right, I found out that I wasn't aware of the current state of this specific part, actually, so that's why I asked. Thanks for the clarification :)

jensens:
Since we support SVG as image - how is this reflected in the metadata?

mauritsvanrees:
I have rebased and updated the PR a bit, after the merge of the needed changes in `plone.scale` and `plone.namedfile`.

I am trying to add a test for a dexterity FTI with multiple image fields. Plone totally fails to find my fields though when iterating over the schemata. Does someone know how to fix this? See aaa207c3706f0d30ebd07cfdd412400bb9d3b294. My next try would be doing this in a GenericSetup profile that I register only in the tests but it feels like overkill.

1letter (commented):
@mauritsvanrees 
i think here is a typo, the fieldnames in images.xml should be unique. image1, image2 ...

mauritsvanrees:
> i think here is a typo, the fieldnames in images.xml should be unique. image1, image2 ...

Thanks, fixed now, but that was not it. Still no fields are found.

jensens:
(I removed WIP from title: We have labels and draft state, its enough).

Products/CMFPlone/image_scales/adapters.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # Get the @@images view once and store it, so all methods can use it.
+        self.images_view = getMultiAdapter((self.context, self.request), name="images")
+
+        width, height = image.getImageSize()
+
+        url = self.get_original_image_url(self.field.__name__, width, height)
+
+        scales = self.get_scales(self.field, width, height)
+        return [
```
pbauer, line 76: Please document why this is a list (because it could be used for IRelationList if the target is a image).

mauritsvanrees: I have documented it.

sneridagh (approved):

sneridagh:
LGTM! Although I might also be missing something, I've been looking at this from the fence. 
We need more feedback!

Products/CMFPlone/image_scales/interfaces.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class IImageScalesFieldAdapter(Interface):
+    """ """
+
+    def __init__(field, context, request):
+        """Adapts field, context and request."""
+
+    def __call__():
+        """Returns JSON compatible python data."""
```
jensens, line 23: The interfaces here need to go to plone.base before merge. @cekk put them here to simplify prototyping. It was never planned to keep them here.

Products/CMFPlone/image_scales/adapters.py
```diff
@@ -0,0 +1,142 @@
...
+    """Returns list of (name, width, height) of the available image scales."""
+    registry = getUtility(IRegistry)
+    imaging_settings = registry.forInterface(IImagingSchema, prefix="plone")
+    allowed_sizes = imaging_settings.allowed_sizes
+    return [_split_scale_info(size) for size in allowed_sizes]
+
+
+@implementer(IImageScalesFieldAdapter)
+@adapter(INamedImageField, IDexterityContent, Interface)
+class ImageFieldScales:
```
jensens, line 57: The specific field adapter should go to plone.namedfile. It was never planned to keep it here, in Bonn we agreed to put it here for rapid prtotyping. cc/ @cekk

jensens (changes requested):
After moving all parts to its respective packages as planned, I would give green light for merge here.

ale-rt:
I think I already wrote that somewhere else but:

- I do not have time to review it
- I trust the work you guys did and I even saw the cool stuff coming out of this work

That's a +1 from my point of view.

It would be cool to have some other @plone/framework-team member or core contributor to say something about that.
If that does not happen consider this work approved.

mauritsvanrees:
Jenkins job with all PRs is [green](https://jenkins.plone.org/job/pull-request-6.0-3.9/1449/).
Meanwhile the PLIP jobs can be run, and they are green as well: [3.8](https://jenkins.plone.org/job/plip-plip-image-scales-metadata-3.8/4/) and [3.9](https://jenkins.plone.org/view/PLIPs/job/plip-plip-image-scales-metadata-3.9/).

jensens (approved):

PR #3524, opened 2022-05-05
sort addons by title in addon controlpanel

### LINKED ISSUE #3523: Case insensitive sort of addon controlpanel items
In the addon controlpanel we are presenting the list of addons without sorting it.

Volto sorts them [case-insensitively by title](https://github.com/plone/volto/blob/master/src/reducers/addons/addons.js#L28) which makes sense. 

We should do the same here in CMFPlone.

erral:
Fixes #3523

### DIFF
--- a/Products/CMFPlone/configure.zcml
+++ b/Products/CMFPlone/configure.zcml
@@ -5,53 +5,69 @@
            xmlns:i18n="http://namespaces.zope.org/i18n"
            i18n_domain="plone">
 
+  <!-- basic zope/cmf -->
   <include package="zope.app.locales" />
   <include package="Products.CMFCore" />
   <include package="Products.GenericSetup" />
-  <include package="plone.app.contentmenu" />
+
+  <!-- Our controlpanel config defines several permissions, so load it early. -->
+  <include package=".controlpanel" />
+
+  <!-- plone.* -->
+  <include package="plone.batching" />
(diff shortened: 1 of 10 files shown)


erral:

avoinea (approved):

mauritsvanrees (changes requested):
Sorting by title is fine. The problem is that this sorts on the _English_ title. Result it that in Dutch it seems just as random as before:

Screenshot 2022-05-06 at 14 45 59

Maybe translate it explicitly in Python before sorting?

Also, optionally do the sorting in `get_addons` itself, so you only have to do it at one point.

erral:
I tried to do it in `get_addons` but it's a dictionary and required to do more changes, that's why I chose to do it in a separate place.

I didn't realised about the translated title indeed, and will take it into account.

erral:
I have reworked this a bit and now the titles are translated. BTW I realised that plone.restapi's `@addons` endpoint does not return translated addon titles so I worked on that too (https://github.com/plone/plone.restapi/pull/1413)

erral:

mauritsvanrees (approved):

mauritsvanrees:
Thanks! Approved.

Optional request: could you sort it case insensitively?

erral:

Issue #3188, opened 2020-09-29
Remove autoinclude entry points from core packages


mauritsvanrees:
In Plone 5.2 coredev, Python 3, the zcml of these packages is being loaded via their `z3c.autoinclude` entry points. I expect that most of these are also loaded explicitly by `Products.CMFPlone`, so the entry points should not be needed. Some of these have started outside of core, and then they needed the entry point, but not any more, for example `plone.app.contenttypes`.  There may be exceptions, especially perhaps some packages that are not dependencies of `CMFPlone` but of `Plone`. 

In alphabetical order:

- [ ] plone.app.caching
- [ ] plone.app.contentlisting
- [ ] plone.app.contenttypes
- [ ] plone.app.dexterity
- [ ] plone.app.discussion
- [ ] plone.app.event
- [ ] plone.app.intid
- [ ] plone.app.iterate
- [ ] plone.app.linkintegrity
- [ ] plone.app.lockingbehavior
- [ ] plone.app.multilingual
- [ ] plone.app.querystring
- [ ] plone.app.theming
- [ ] plone.app.upgrade
- [ ] plone.app.users
- [ ] plone.app.versioningbehavior
- [ ] plone.app.widgets
- [ ] plone.batching
- [ ] plone.formwidget.namedfile
- [ ] plone.formwidget.recurrence
- [ ] plone.outputfilters
- [ ] plone.portlet.static
- [ ] plone.resource
- [ ] plone.rest
- [ ] plone.restapi
- [ ] plone.staticresources
- [ ] plone.stringinterp
- [ ] plone.subrequest
- [ ] plonetheme.barceloneta
- [ ] Products.CMFDiffTool
- [ ] Products.CMFEditions
- [ ] z3c.formwidget.query

Maybe Plone 6.0 is a natural time to clean this up. But should be okay in 5.2 as well. Should be easy: simply remove a few lines from `setup.py`.

Next step is probably to split this list up into a list of packages that are indeed explicitly included in CMFPlone and a list for which this is not the case.  Perhaps CMFPlone should start explicitly including the packages from the second list if available.

mauritsvanrees:
These packages are already included explicitly in `Products.CMFPlone` or elsewhere, so their entry point can be removed immediately:

- [ ] plone.app.discussion
- [ ] plone.app.event
- [ ] plone.app.intid
- [ ] plone.app.lockingbehavior
- [ ] plone.app.querystring
- [ ] plone.app.versioningbehavior
[...]

PR #3525, opened 2022-05-05
Explicitly include zcml of more packages.

mauritsvanrees:
Part of https://github.com/plone/Products.CMFPlone/issues/3188.
All packages in the `install_requires` that have zcml, should be loaded explicitly.

Also reorder the zcml loading more logically: first `plone.*`, then `plone.app.*`.

Require `plone.resource`: this is used by various packages, but only required by `plone.staticresources`, which did not seem right to me.

### DIFF
--- a/Products/CMFPlone/configure.zcml
+++ b/Products/CMFPlone/configure.zcml
@@ -5,53 +5,72 @@
            xmlns:i18n="http://namespaces.zope.org/i18n"
            i18n_domain="plone">
 
+  <!-- basic zope/cmf -->
   <include package="zope.app.locales" />
   <include package="Products.CMFCore" />
   <include package="Products.GenericSetup" />
-  <include package="plone.app.contentmenu" />
+
+  <!-- plone.* -->
+  <include package="plone.batching" />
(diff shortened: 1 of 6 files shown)


mauritsvanrees:

mauritsvanrees:
Okay, we should not try to load `plone.app.caching` zcml as it causes test failures. I have removed the conditional includes, except `plone.app.upgrade` which was already there.

jensens:
Question is, do we want to include everything here? For instance plone.batching is included in plone.base, isnt it plone.base which should get included here and job done?

mauritsvanrees:
> Question is, do we want to include everything here? For instance plone.batching is included in plone.base, isnt it plone.base which should get included here and job done?

We could indeed do this for the packages that are in the `install_requires` of `plone.base`. Currently only these three: `plone.batching`, `plone.registry`, `plone.z3cform`. Since `plone.base` has no zcml yet, I would say `Products.CMFPlone` is the correct place.
We could move some though. Perhaps also register permissions in `plone.base`.

jensens (approved):

PR #3538, opened 2022-05-21
Rename passwordreset into password-reset

cekk:
Related to https://github.com/plone/plone.volto/pull/61

Can be nice to have the same name in both Plone and Volto.
I prefer "password-reset" but if Plone endpoint is more important for backward compatibility, we can change it in Volto.

### DIFF
--- a/Products/CMFPlone/browser/login/configure.zcml
+++ b/Products/CMFPlone/browser/login/configure.zcml
@@ -121,7 +121,7 @@
         />
 
     <browser:page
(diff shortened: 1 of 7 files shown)


cekk:

jensens (approved):
I dont mind the name here. Existing customization/addons may fail if there are overrides, that's the only issue I see here.

cekk:
I agree..let's change it into volto and close this pr.
Thanks

Issue #3537, opened 2022-05-20
Robot tests are not always reported as failures


mauritsvanrees:
On current coredev 6, Python 3.10, with a few packages on other branches but I think this does not matter:

```
$ bin/test --all -s Products.CMFPlone -m test_robot -t 'Simple Livesearch'
...
  Set up plone.app.event.testing.PAEventLayer in 0.055 seconds.
  Set up plone.app.contenttypes.testing.PloneAppContenttypes in 0.013 seconds.
  Set up Products.CMFPlone.testing.ProductsCMFPloneLayer in 0.095 seconds.
  Set up plone.app.robotframework.remote.CMFPloneRobotRemoteLibrary:RobotRemote in 0.000 seconds.
  Set up plone.testing.zope.WSGIServer in 0.031 seconds.
  Set up Products.CMFPlone.testing.CMFPloneLayer:Acceptance in 0.000 seconds.
  Running:
    1/1 (100.0%)/Users/maurits/shared-eggs/cp310/selenium-4.1.3-py3.10.egg/selenium/webdriver/chrome/webdriver.py:70: DeprecationWarning: desired_capabilities has been deprecated, please pass in a Service object
  super(WebDriver, self).__init__(DesiredCapabilities.CHROME['browserName'], "goog",
<frozen importlib._bootstrap>:283: DeprecationWarning: the load_module() method is deprecated and slated for removal in Python 3.12; use exec_module() instead

==============================================================================
Test Livesearch                                                               
==============================================================================
Scenario: Simple Livesearch                                           | FAIL |
Keyword 'Element should contain' failed after retrying for 5 seconds. The last error was: Element with locator 'css=.livesearch-results li a .heading' not found.
------------------------------------------------------------------------------
Test Livesearch                                                       | FAIL |
1 test, 0 passed, 1 failed
==============================================================================
Output:  /Users/maurits/community/plone-coredev/6.0/parts/test/test_livesearch/Scenario_Simple_Livesearch/output.xml

                
  Ran 1 tests with 0 failures, 0 errors, 0 skipped in 10.727 seconds.
Tearing down left over layers:
  Tear down Products.CMFPlone.testing.CMFPloneLayer:Acceptance in 0.000 seconds.
  Tear down plone.testing.zope.WSGIServer in 0.002 seconds.
  Tear down plone.app.robotframework.remote.CMFPloneRobotRemoteLibrary:RobotRemote in 0.000 seconds.
[...]

PR #3540, opened 2022-05-23
Add GitHub action to test PR on coredev.

mauritsvanrees:
This is a gitHub Action for testing a package PR on buildout.coredev.
For now (May 2023) it is an experiment, but I think we need it.
Reasons:

1. The robot tests on Jenkins are flaky/unstable, which most likely is due to several robot jobs running parallel on one node.
2. Python 3.10 is not working with the 'Python Shining Pandas' plugin we use on Jenkins.  We had it running with a different script for a while, but these changes got lost.  Should be restorable, but let's try GHA.

One thing to watch out for, is that robot tests are not always reported as failures.
On the one hand there is config for that on Jenkins:
https://github.com/plone/jenkins.plone.org/issues/297
On the other hand I see the same problem locally:
https://github.com/plone/Products.CMFPlone/issues/3537

TODO: put this in a reusable workflow in plone/.github:
https://docs.github.com/en/actions/using-workflows/reusing-workflows
But first we should see if it actually works.

Aftwerwards, we could install it in all relevant Plone repos using https://github.com/asottile/all-repos
But nothing too hasty. :-)


mauritsvanrees:
@fredvd @gforcada What do you think of this approach as (partial) alternative to Jenkins? For starters we could move the robot tests and the 3.10 tests to GHA.
We should definitely look at reusable workflows. With a bit of luck, the yaml file per repository only needs about five lines. I have not tried this yet though.

Note that we suffer from some [unstable folder contents tests](https://github.com/plone/Products.CMFPlone/issues/3542), just like Jenkins.

mauritsvanrees:
I have moved to a reusable workflow, which is easy to do, once you realise the workflow must be in a specific directory. Now the yaml file in this repo is basically just five lines. We can make a [starter workflow](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization) to make that even easier.
Tests are currently running.

mauritsvanrees:
If we need something to replace parts of `mr.roboto`, the [GitHub CLI is available](https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows) in GitHub Actions. You can for example use it to add a comment, like we do with "To start Jenkins jobs, copy-paste the next comment.".
I use it locally for simple things like getting a list of open issues and PRs before I make a release of a package.

gforcada:
I still did not have time to look at it properly, but my fear, of having to duplicate lots and lots of configuration is down to only 5 lines, that sounds quite good to me ✨ 💯 

The idea then is to get every repository to run its tests on its own as well as the whole test suite? 🤔 

For simple PR that sounds enough, if there are several PRs working together, that might be a bit more problematic, but I guess we will find a solution 👍🏾 

As for 3.10, sorry, indeed we had it working, but somehow it got lost, it should be easy, as in directly calling python3.10 and creating a venv manually, so it should be fairly quick to get it back to work, unfortunately I'm rather busy for the next few weeks still 😖

mauritsvanrees:
For multiple related PRs we could create a plip config in coredev.

mauritsvanrees:
I was glancing at my open pull requests, and this is the oldest one.
This was an experiment from 2022.  Meanwhile we have `plone.meta` for this.
I have removed the related [workflow in the `plone/.github` repo](https://github.com/plone/.github/commit/59b99cf424ca39b86ee9637cb31fbdaa6a344cf7).
The nice thing here is that you do a coredev checkout, so you really have the latest updates to all packages, without asking me to make some releases and update list.plone.org.  I suppose it takes longer.
Anyway, I will close this.  If we want something like that, it should be done in `plone.meta`.

PR #3552, opened 2022-06-05
try fixing flaky robottests

petschki:
Jenkins Parameter:
```
https://github.com/plone/Products.CMFPlone/pull/3552
https://github.com/plone/plone.app.widgets/pull/218
```

PR #3553, opened 2022-06-07
More robust linkintegrity robot tests

fredvd:


mauritsvanrees:
Moved into PR #3552 which is now merged.

PR #3554, opened 2022-06-07
Moved brand new wait-for-element keywords to plone.app.robotframework.

mauritsvanrees:
Needs https://github.com/plone/plone.app.robotframework/pull/136
Use the new keywords in more tests.
This should hopefully again bring a bit more stability.

### DIFF
--- a/Products/CMFPlone/tests/robot/keywords.robot
+++ b/Products/CMFPlone/tests/robot/keywords.robot
@@ -33,23 +33,3 @@ patterns are loaded
 a folder with a document '${title}'
   ${folder_uid}=  Create content  type=Folder  title=folder
   Create content  type=Document  container=${folder_uid}  title=${title}
-
-Wait For Element
-  [Arguments]  ${element}
-  [Documentation]  Can contain css=, jquery=, or any other element selector.
-  Wait Until Page Contains Element  ${element}
-  Set Focus To Element  ${element}
-  Wait Until Element Is Visible  ${element}
-
-Wait For Then Click Element
-  [Arguments]  ${element}
-  [Documentation]  Can contain css=, jquery=, or any other element selector.
-  Wait For Element  ${element}
-  Click Element  ${element}
-
-Wait For Then Click Hidden Element
-  [Arguments]  ${element}
-  [Documentation]  Meant for invisible elements.  Can contain css=, jquery=, or any other element selector.
-  Wait Until Page Contains Element  ${element}
-  Set Focus To Element  ${element}
-  Click Element  ${element}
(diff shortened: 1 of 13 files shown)


mauritsvanrees:
I am testing this plus https://github.com/plone/plone.app.robotframework/pull/136 on [Plone 6 Py 3.8](https://jenkins.plone.org/job/pull-request-6.0-3.8/1452/).

petschki (approved):
LGTM

I've fixed two more deprecated "Press Key" keywords here ...

fredvd (changes requested):
Fine to move this to robotframework instead. 

I'm a bit concerned about the split between
"Wait For Then Click Element" and "Wait For Then Click Hidden Element".

* Are devs editing the robot tests easily going to understand the difference between the two variants and that the latter is adding an extra set Focus?  That doesn't have to do with hidden, but being 'selected'?  (And it should maybe be called "Wait For Then Focus Click Element"

* Are we sure that setting the focus is not part of the solution/stop gap that really makes the tests more stable? Then it doesn't make sense to offer 2 variants, I'd prefer to have just 'Wait For Then Click Element' and let it always do the focus step as well.

mauritsvanrees:
> I'm a bit concerned about the split between "Wait For Then Click Element" and "Wait For Then Click Hidden Element".
> 
> * Are devs editing the robot tests easily going to understand the difference between the two variants and that the latter is adding an extra set Focus?

That is not the difference.

`Wait For Then Click Element` is:

```
    [Arguments]  ${element}
    Wait Until Page Contains Element  ${element}
    Set Focus To Element  ${element}
    Wait Until Element Is Visible  ${element}
    Click Element  ${element}
```

`Wait For Then Click Hidden Element` is:

```
    Wait Until Page Contains Element  ${element}
    Set Focus To Element  ${element}
    Click Element  ${element}
```

So for the hidden one, the line `Wait Until Element Is Visible` is missing.
I am using this in two places in `test_querystring.robot` on the element `div#select2-drop-mask`. The element is on the page, you can focus on it, but you cannot see it. When I try it myself with `bin/instance` I see this element has no content. That is probably why `Wait Until Element Is Visible` fails.

I could rename this to `Wait For Then Click Invisible Element`. Add an extra line of documentation.

But maybe this is just too specific and uncommon. Shall I remove this keyword and copy those lines to `test_querystring.robot`?

fredvd:
>     * Are we sure that setting the focus is not part of the solution/stop gap that really makes the tests more stable? Then it doesn't make sense to offer 2 variants, I'd prefer to have just 'Wait For Then Click Element' and let it always do the focus step as well.   (edit: just like "Wait For Element" already does the Focus as well)

Ignore this one. I didn't see that "Wait For Then Click Element" calls "Wait For Element" abwhich has the Focus line..  

> I could rename this to Wait For Then Click Invisible Element. Add an extra line of documentation.
> But maybe this is just too specific and uncommon. Shall I remove this keyword and copy those lines to test_querystring.robot?

Not it is not too specific, I see the use case now.  If you add the documentation there that this is a special case when "Wait For Then Click Element" doesn't work, it is fine to have it on the tool belt.

mauritsvanrees:
I have renamed the keyword to `Wait For Then Click Invisible Element` and improved the inline documentation. New job:
https://jenkins.plone.org/job/pull-request-6.0-3.8/1457/console

PR #3560, opened 2022-06-21
Resource registry accordions initially open

petschki:
To avoid inaccessible resources when JS errors occurs on the page, define accordions as initally open and hide them with JS.

NOTE: right now this works only with JS disabled. More updates to check for correctly initialized `bootstrap.Collapse` are about to come.


petschki:

petschki:
OK ... this is ready for review. As said right now only for JS disabled support ... cannot check for proper initialized Bootstrap until we have globals registraion back ... see https://github.com/plone/mockup/pull/1177

petschki:

jensens (approved):

jensens:
The other check can be implemented in a second PR. This one is already valuable.

PR #3562, opened 2022-06-22
Improve error handling in Products.CMFCore.factory.addPloneSite

### LINKED ISSUE #3561: addPloneSite swallows exceptions while importing extension profiles
If there's an exception while Products.CMFPlone.factory.addPloneSite is installing an extension profile, it is logged but does not abort execution. This causes problems if addPloneSite is used in a script and something is relying on the exit status of the script to determine whether the site was created successfully (like in https://github.com/plone/plone-backend/blob/5.2.x/skeleton/scripts/create_site.py). It also leads to confusion since it can hide the fact that the site was not fully configured.

Even worse, the exception is obscured in the log because Zope hits a different exception while trying to render a page showing the original exception. This happens because the not-yet-fully-configured Plone site is still set as the active zope.component site.

davisagli:
- Avoid swallowing exceptions while importing extension profiles
- Remove the status message reporting these errors (no longer relevant since the addsite form no longer offers selecting addon profiles)
- Make sure the component site gets cleared if there is an exception

Fixes #3561

### DIFF
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -5,8 +5,8 @@ from Products.CMFPlone.events import SiteManagerCreatedEvent
 from plone.base.interfaces import INonInstallable
 from Products.CMFPlone.Portal import PloneSite
 from Products.GenericSetup.tool import SetupTool
-from zope.component.hooks import site as global_site
 from zope.component import queryUtility
+from zope.component.hooks import setSite
 from zope.event import notify
 from zope.interface import implementer
 from zope.lifecycleevent import ObjectCreatedEvent
@@ -138,7 +138,9 @@ def addPloneSite(context, site_id, title='Plone site', description='',
     setup_tool = site[_TOOL_ID]
 
     notify(SiteManagerCreatedEvent(site))
-    with global_site(site):
+    setSite(site)
+
+    try:
         setup_tool.setBaselineContext('profile-%s' % profile_id)
         setup_tool.runAllImportStepsFromProfile('profile-%s' % profile_id)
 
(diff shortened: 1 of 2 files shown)


pbauer (approved):

mauritsvanrees (approved):
> Remove the status message reporting these errors (no longer relevant since the addsite form no longer offers selecting addon profiles)

You *can* still select add-on profiles in the advanced form.

But the change makes sense.
Tried it out like this:

- Go to the advanced form.
- Select all add-ons. I have this snippet to paste in the javascript console to check all inputs: `inputs = document.getElementsByTagName('input');for (var index = 0; index < inputs.length; index++) inputs[index].checked = true`
- On master this leads to two status messages: "Could not install plone.volto:default-homepage-drafjs: ('en-us',)", and same for `slatejs`. (Standard Volto add form works fine BTW.)
- On your branch I get a proper traceback.

There is one thing I wonder about. I added this code in `browser/admin.py` and printed it before and after calling `addPloneSite`:

```
from zope.component.hooks import getSite
from zope.component.hooks import getSiteManager

print(getSite())
print(getSiteManager())
```

In both cases, before calling it you get:

```
None
<BaseGlobalComponents base>
```

On master, afterwards you have this:

```
<PloneSite at Plone>
<PersistentComponents /Plone>
```

On your branch, since you use a context manager for setting the site, afterwards we are back to:

```
None
<BaseGlobalComponents base>
```

I would say this is fine. Maybe even better than before. After this, the browser view only does `self.request.response.redirect(site.absolute_url())` and returns, which works fine in both cases.
I am just wondering if this change can cause problems, maybe in other use cases.
An option could be to call `siteSite(site)` at the end of `factory.addPloneSite` when all is well. Then we would have the same behavior as before. What do you think?

It is probably fine the way it is now, so I will approve. But if someone can comment on this small concern, it would be good.

davisagli:

PR #3564, opened 2022-06-26
do not use f-strings in translations, use proper _ calls and parameters

erral:
f-strings are not properly interpolated by gettext, we need to keep using proper parameters and pass them to _ calls.


erral:

jensens (approved):

PR #3567, opened 2022-06-29
Removed earlypatches/expressions.py.

mauritsvanrees:
This was a patch to avoid some too strict checks by Zope / Products.PageTemplates.
But in Plone 6 it should be fine to be stricter.


mauritsvanrees:

news/3567.breaking
```diff
@@ -0,0 +1,5 @@
+Removed our expressings patch.
```
davisagli, line 1:
```suggestion
Removed our expressions patch.
```

mauritsvanrees:
A test in `five.customerize` was failing, but this was already fixed on its master branch. I have made a release of it and updated coredev. Should be green now:

Could someone from @plone/security-team also review please?

jensens (approved):

PR #3572, opened 2022-07-04
Cleanup dependencies

jensens:
Reduce dependencies here, when already fulfilled in the packages where in use.
I do not think we have to declare Zope and plone.base dependencies or those of packages which are already pulling in their stuff correctly. 
 
I hope this speeds up install a bit. 
See the single commits.


jensens:

jensens:

mauritsvanrees (approved):
I added some inline comments, because not all reasons for removing a dependency were correct. But in the end they can all be removed . LGTM.

PR #3575, opened 2022-07-06
Erral i18n fixes

erral:


erral:

Products/CMFPlone/browser/login/password_reset.py
```diff
@@ -196,7 +196,7 @@ def getErrors(self):
         if state:
             state['status'] = 'failure'
             state['portal_status_message'] = _(
-                'Please correct the indicated errors.',
```
mauritsvanrees, line 199: Why do you remove this comma? I have seen you do it in earlier PRs, but I can't imagine that this has any effect.
I tried `i18ndude rebuild-pot`, but it does not mind the comma.

I can't seem to trigger showing the error message, but when I add a status message like this in code, it gets translated to Dutch just fine:


```
        IStatusMessage(self.request).addStatusMessage(
            _(
                'Please correct the indicated errors.',
            )
        )
```

If removing the comma somehow *is* needed, then I expect it is also needed in line 194 right above it.

BTW, I was worried that running `black` would put the comma back, but that does not happen. So for that part we seem safe.

mauritsvanrees (approved):
Fine to merge, but I have one inline comment where I wonder if removing the comma is really needed.

Products/CMFPlone/browser/login/password_reset.py
erral: What I found is that in some cases the message used there is used in 2 places. Somehow, and I don't know why, this last comma creates a different "default message" in the po file and when running i18ndude to create the po files it complains because this message has different default message values.

The message is translated correctly but it may create different  default messages in the PO file.

mauritsvanrees: I see. I confirm the problem and have created an issue: https://github.com/collective/i18ndude/issues/97
Looks like it is not a big problem: the msgid ends up only once in the pot file, and there is no default value. The order in which the msg ids are found, is also no problem. But should be solved in `i18ndude` indeed. I have not looked at how to do this.

PR #3585, opened 2022-08-03
Set portal title in registry when creating a new Plone Site

### LINKED ISSUE #3584: Site title entered in the @@plone-addsite form is not set as site title
Plone site title is set in the @@site-controlpanel form and saved in the registry. But the @@plone-addsite form gets the site title from the user and saves it as a Zope property in the portal object.

Related to https://github.com/plone/plone.app.layout/issues/317

erral:
Fixes #3584

### DIFF
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -149,6 +149,7 @@ def addPloneSite(context, site_id, title='Plone site', description='',
         reg['plone.available_timezones'] = [portal_timezone]
         reg['plone.default_language'] = default_language
         reg['plone.available_languages'] = [default_language]
+        reg['plone.site_title'] = title
 
         # Install default content types profile if user do not select "example content"
         # during site creation.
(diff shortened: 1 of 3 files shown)


erral:

erral:
I give up with those robot tests...

jensens:
> I give up with those robot tests...

I think those are unrelated

jensens:
Nice catch. We need to drop setting the old Zope property then too, right? IMO it does not make sense to have the title in both places.

yurj:
> Nice catch. We need to drop setting the old Zope property then too, right? IMO it does not make sense to have the title in both places.

The Zope `title` and `description` properties are the same as the SiteRoot dexterity object fields, if you change in edit via Plone UI or using `/manage`, they both update each other. I don't know why. It seems a special case for every Plone object, the zope `title` and `description` properties are the same as the Plone object. This does not happen for other properties, just for `title` and `description`.

`description` is also used in the `<meta name="description" content="` html head attribute and can be edited, as `title`, in both zope and Plone edit. They are the same property.

The title of the SiteRoot DX object is the title of that object but it is not used (no listing, no catalog). It is used when using the Zope interface:
siteroot_title

Surprisingly, the zope `title` and `description` properties of the Site Root dx object cannot be deleted via the Zope interface (but you can do it in all the other Plone objects).

erral:
In Plone 5.2 the <plone_root>/edit redirect to @@site-controlpanel which is not like that in Plone 6, and neither the site root title is synced to the registry :thinking: 

I think that it doesn't make sense to have it in two places.

mauritsvanrees (approved):
I tested it and it works. Thanks!

PR #3588, opened 2022-08-06
change test to make sure email is sent in utf-8

erral:
See https://github.com/plone/plone.app.layout/issue/317

To be tested and merged with https://github.com/plone/plone.app.layout/pull/318


erral:
These changes are needed because with the changes in the referenced PR of plone.app.layout, the navigation_root_title is the portal title set in the registry and not the Title of the root object.

And in the referenced test the title is set to `Tëst Portal` which apparently is not correct us-ascii as it was previously tested.

PR #3590, opened 2022-08-09
Robot tests: be more specific when clicking some elements.

mauritsvanrees:
Test these together:

```
https://github.com/plone/plone.app.robotframework/pull/140
https://github.com/plone/plone.schemaeditor/pull/94
https://github.com/plone/Products.CMFPlone/pull/3590
```

The plone.app.robotframework PR is stricter: when you click an element, this element should exist only once. This leads to seven failures in CMFPlone, which I fix here.

### DIFF
--- a/Products/CMFPlone/tests/robot/keywords.robot
+++ b/Products/CMFPlone/tests/robot/keywords.robot
@@ -33,3 +33,7 @@ patterns are loaded
 a folder with a document '${title}'
   ${folder_uid}=  Create content  type=Folder  title=folder
   Create content  type=Document  container=${folder_uid}  title=${title}
+
+folder contents pattern loaded
+    Page should contain element  css=.pat-structure
+    Wait For Condition  return !!document.querySelector('.pat-structure div.navbar')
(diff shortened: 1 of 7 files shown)


petschki (approved):
I gave it one more try on Jenkins but this looks very good to me. When once more green I'll merge all three.

petschki:
This one I also saw very often. It seems that tinymce doesn't save the text given.

The line `Wait Until Keyword Succeeds 5s, 1s, Execute Javascript, tinyMCE.activeEditor.setContent('${input}');`  in the keyword `Input RichText` is not failing, but it also doesn't set anything. Maybe there should be an extra test after `setContent` if the content is really written to the textarea and if not, repeat it.

mauritsvanrees:
@petschki I have added another fix for stale references in the linkintegrity tests. I will stop the jobs you just started, and restart them.

mauritsvanrees:
> Maybe there should be an extra test after `setContent` if the content is really written to the textarea and if not, repeat it.

I would be in favour of that. What you changed now will make it clear if this is indeed the problem. Likely it is.

petschki:
Seems quite stable now ... I'd merge it.

mauritsvanrees:
All merged. Thanks!

PR #3616, opened 2022-08-19
Fix actions

jensens:
After adding explicit deprecation messages to before only soft-deprecated methods on ploneview, this step is needed.
I expect a minor performance improvement as well.
Anyway, this here helps only on new sites, an upgrade step is needed too.

### DIFF
--- a/Products/CMFPlone/ActionsTool.py
+++ b/Products/CMFPlone/ActionsTool.py
@@ -2,55 +2,58 @@ from AccessControl import ClassSecurityInfo
 from AccessControl.class_init import InitializeClass
 from Products.CMFCore.ActionInformation import ActionInfo
 from Products.CMFCore.ActionsTool import ActionsTool as BaseTool
+from Products.CMFCore.interfaces import IActionCategory
 from Products.CMFCore.interfaces import IActionProvider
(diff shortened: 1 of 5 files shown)


jensens:

jensens:
As this needs an upgrade step, is it enough to take a copy of `actions.xml` with the non-changed fields removed? It could then overwrite customizations. Does somebody have a better idea?

cc @mauritsvanrees @petschki @ale-rt @cekk

mauritsvanrees (approved):
LGTM

mauritsvanrees:
> As this needs an upgrade step, is it enough to take a copy of `actions.xml` with the non-changed fields removed? It could then overwrite customizations. Does somebody have a better idea?

That is the right approach.

jensens:

ale-rt:
> As this needs an upgrade step, is it enough to take a copy of `actions.xml` with the non-changed fields removed? It could then overwrite customizations. Does somebody have a better idea?
> 

This could theoretically wipe out user customizations...

I know it is more tedious, but IMO the right approach is to cicle the portal actions objects and replace `url_expr` with the new value if and only if it matches the old value.

jensens:
> This could theoretically wipe out user customizations...

As I mentioned above. But trust my release manager here (see above).

PR #3629, opened 2022-08-26
fixed alert message flowing in <body>

### LINKED ISSUE #3598: Portal status message for User/Group settings displayed with H1 and body
Plone 6.0.0b1

Bildschirmfoto 2022-08-09 um 21 19 59

The portal message should appear consistently above the H1 tag as here:

Bildschirmfoto 2022-08-09 um 21 19 47

rohnsha0:
fixed #3598

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/controlpanel_usergroups_layout.pt
+++ b/Products/CMFPlone/controlpanel/browser/controlpanel_usergroups_layout.pt
@@ -8,15 +8,15 @@
 
 <body>
 
+<div metal:use-macro="context/global_statusmessage/macros/portal_message">
+      Portal status message
+    </div>
+
 <div metal:fill-slot="prefs_configlet_main">
 
   <header>
     <h1 class="documentFirstHeading"
         i18n:translate="">User and Groups Settings</h1>
-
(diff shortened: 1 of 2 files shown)


rohnsha0:

Products/CMFPlone/controlpanel/browser/controlpanel_usergroups_layout.pt
```diff
@@ -8,16 +8,16 @@
 
 <body>
 
+<div metal:use-macro="context/global_statusmessage/macros/portal_message">
+      Portal status message
+    </div>
```
davisagli, line 13: @rohnsha0 This won't do anything. The template is using a macro (see line 6) and filling a slot in that macro (see line 15). Anything in this file that is not inside the slot will not be used.

But, after looking for a couple minutes I don't yet understand why this template needs to customize the normal display of the status message at all, so this might need more investigation.

rohnsha0: As the alert message is placed above the nav bar.... 
But see in the screenshots for the issue for reference!
I think that's misplaced!
Correct me if I'm wrong!

PR #3638, opened 2022-09-08
Add the styles CSS from the theme to the content CSS to load.

Rudd-O:
Without this, the theme creator's styles CSS is never loaded by TinyMCE.  `importcss_file_filter` acts only as a filter lambda "permit these paths to load", but if these paths are not added to the `content_css` setting, then they are never loaded by TinyMCE.

The official Plone theming documentation states this should work — it doesn't without this patch, it does with it.


Rudd-O:

MrTango:
Why do you use yoru own fork for the PR? it's much easier to test your PR when it's direktlich a branch inside the original repository ;)

MrTango:
@Rudd-O did you add your CSS in your manifest.cfg tinymce-content-css?

Rudd-O:
> Why do you use yoru own fork for the PR? it's much easier to test your PR when it's direktlich a branch inside the original repository ;)

I completely agree and until today I did not know I could just push to a branch in the org!  Sorry, I am super green in this stuff!

Rudd-O:
> Also we have to make sure that the theme css is not used for the format list in tinymce.

Imma save you some time.  With this PR, if you don't specify `tinymce-styles-css`, the content CSS of the theme (which used to slow down the formats dropdown enormously) *is* applied to the content, but is **not** loaded in the formats dropdown anymore.  Of course, the URL you specify in `tinymce-styles-css` **is** (with this PR) loaded in the formats dropdown.

Rudd-O:
For some reason the merge commit went into my own repo, instead of this one.

jensens:

Rudd-O:

Rudd-O:
Chex passed @jensens @thet  @MrTango .  Thanks!

jensens (approved):

PR #3639, opened 2022-09-10
Require Python 3.8 or higher.

mauritsvanrees:
We already dropped official support for 3.7. This makes it concrete.
Actually the syntax is unchanged. `pyupgrade` made no extra changes when using `--py38-plus`.

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -151,8 +151,8 @@ def allowedRolesAndUsers(obj):
 @indexer(Interface)
 def object_provides(obj):
     return tuple(
-        [i.__identifier__ for i in providedBy(obj).flattened()
-         if i.__identifier__ not in DENIED_INTERFACES]
+        i.__identifier__ for i in providedBy(obj).flattened()
+         if i.__identifier__ not in DENIED_INTERFACES
     )
 
 
--- a/Products/CMFPlone/tests/testInterfaces.py
+++ b/Products/CMFPlone/tests/testInterfaces.py
@@ -297,5 +297,5 @@ import unittest
 def test_suite():
     suite = unittest.TestSuite()
     for test in tests:
-        suite.addTest(unittest.makeSuite(test))
+        suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(test))
     return suite
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup(
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
     ],
+    python_requires='>=3.8',
     keywords='Plone CMF Python Zope CMS Webapplication',
     author='Plone Foundation',
     author_email='releasemanager@plone.org',
--- a/Products/CMFPlone/Portal.py
+++ b/Products/CMFPlone/Portal.py
@@ -116,7 +116,7 @@ class PloneSite(Container, SkinnableObjectManager, UniqueObject):
 
     # From PortalObjectBase
     def __init__(self, id, title=''):
-        super(PloneSite, self).__init__(id, title=title)
+        super().__init__(id, title=title)
         components = PersistentComponents('++etc++site')
         components.__parent__ = self
         self.setSiteManager(components)
@@ -135,7 +135,7 @@ class PloneSite(Container, SkinnableObjectManager, UniqueObject):
             pass
         self.setupCurrentSkin(REQUEST)
 
-        super(PloneSite, self).__before_publishing_traverse__(arg1, arg2)
+        super().__before_publishing_traverse__(arg1, arg2)
 
     # Concept from OFS.OrderSupport
     @security.protected(permissions.AccessContentsInformation)
--- a/Products/CMFPlone/WorkflowTool.py
+++ b/Products/CMFPlone/WorkflowTool.py
@@ -258,7 +258,7 @@ class WorkflowTool(PloneBaseTool, BaseTool):
                                 objects_by_path[absurl] = (o.modified(), o)
 
         results = objects_by_path.values()
-        return tuple([obj[1] for obj in sorted(results)])
+        return tuple(obj[1] for obj in sorted(results))
 
     security.declareProtected(ManagePortal, 'getChainForPortalType')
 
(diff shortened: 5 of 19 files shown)


mauritsvanrees:

Products/CMFPlone/tests/testInterfaces.py
```diff
@@ -297,5 +297,5 @@ class KlassInterfaceTest(zope_interface_test):
 def test_suite():
     suite = unittest.TestSuite()
     for test in tests:
-        suite.addTest(unittest.makeSuite(test))
+        suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(test))
     return suite
```
davisagli, line 301: The Zope testrunner has supported discovery of test classes without a test_suite function since 2009.

setup.py
```diff
@@ -24,6 +24,7 @@
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
     ],
+    python_requires='>=3.8',
```
davisagli, line 27: @mauritsvanrees Adding this means that if someone tries to pip install Products.CMFPlone on an older Python, they'll get the most recent version that did not have this python_requires value.

Well, hopefully they are installing with a constraints file, and will see an error since the specified version isn't available for their Python. So it is probably okay. If they aren't using constraints, there are bigger problems.

davisagli (approved):
LGTM. It looks like there are some `test_suite` functions that can be removed, but that can also happen separately.

Products/CMFPlone/tests/testInterfaces.py
davisagli: Never mind, I see this is related to some test classes that are generated above, so maybe it is still needed.

davisagli:
The Python 3.8 and 3.9 builds on Jenkins were quite broken, so I restarted them.

setup.py
mauritsvanrees: I have noticed that recent pip versions complain when you do not have python_requires at all.

rohnsha0 (approved):

jensens (approved):
yay!

Products/CMFPlone/CatalogTool.py
```diff
@@ -151,8 +151,8 @@ def allowedRolesAndUsers(obj):
 @indexer(Interface)
 def object_provides(obj):
     return tuple(
-        [i.__identifier__ for i in providedBy(obj).flattened()
-         if i.__identifier__ not in DENIED_INTERFACES]
+        i.__identifier__ for i in providedBy(obj).flattened()
+         if i.__identifier__ not in DENIED_INTERFACES
```
ale-rt, line 155: Note that the previous code was faster:
```
Python 3.8.12 (default, Sep 29 2021, 09:48:53) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: my_range = tuple(range(100))

In [2]: %timeit tuple([i for i in my_range])
1.51 µs ± 3.83 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit tuple(i for i in my_range)
2.42 µs ± 16.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

mauritsvanrees: Strange. I simply ran `pyupgrade`.
Do you know why the old way is faster?

ale-rt: This is something I learned from @jamadden in some comment of a PR in the zopefoundation organisation that I am not able to find anymore.
IIRC that was caused by some C level optimisation.

And yeah, I was surprised from that as well!

yurj: Because it is done with less frame switching, see the first comment in:

https://stackoverflow.com/questions/43808180/python-3-x-list-comprehension-vs-tuple-generator

yurj:
```
>>> my_range = tuple(range(100))
>>> def f1():
...  tuple([i for i in my_range])
... 
>>> def f2():
...  tuple(i for i in my_range)
... 
>>> import dis
>>> dis.dis(f1)
  2           0 LOAD_GLOBAL              0 (tuple)
              3 BUILD_LIST               0
              6 LOAD_GLOBAL              1 (my_range)
              9 GET_ITER            
        >>   10 FOR_ITER                12 (to 25)
             13 STORE_FAST               0 (i)
             16 LOAD_FAST                0 (i)
             19 LIST_APPEND              2
             22 JUMP_ABSOLUTE           10
        >>   25 CALL_FUNCTION            1
             28 POP_TOP             
             29 LOAD_CONST               0 (None)
             32 RETURN_VALUE        
>>> dis.dis(f2)
  2           0 LOAD_GLOBAL              0 (tuple)
              3 LOAD_CONST               1 (<code object <genexpr> at 0x7ffbd76a31b0, file "<stdin>", line 2>)
              6 MAKE_FUNCTION            0
              9 LOAD_GLOBAL              1 (my_range)
             12 GET_ITER            
             13 CALL_FUNCTION            1
             16 CALL_FUNCTION            1
             19 POP_TOP             
             20 LOAD_CONST               0 (None)
             23 RETURN_VALUE        
>>> 
```

disassembling it shows in the second case it is using the generator, maybe the tuple conversion find it and doesn't call it until a value is requested (next())?

ale-rt: > Strange. I simply ran `pyupgrade`. Do you know why the old way is faster?

I opened an issue to pyupgrade: https://github.com/asottile/pyupgrade/issues/704

wesleybl:
@mauritsvanrees @gforcada 

It would be nice to remove Python 3.7 from Jenkins' Plone 6 jobs. Tests would be faster.

Products/CMFPlone/CatalogTool.py
davisagli: @ale-rt good catch!

jamadden: Perhaps thinking of this?

https://github.com/zopefoundation/zope.interface/blob/009c3818662a9f15696ee4b8c2beda55b1a97cb8/src/zope/interface/adapter.py#L34-L63

(https://github.com/zopefoundation/zope.interface/commit/1e720c3819b8daa07aaad0ee0258d9035597f27d)

gforcada:
Seems someone beat me at removing them 😄

PR #3644, opened 2022-09-15
isort

mauritsvanrees:
Also, see how Jenkins and mr.roboto are doing. They are getting more and more into shape.


mauritsvanrees:
Python 3.7 was initially showing up again as pending in the checks. I have edited the branch protection rule for master to no longer require it (and added 3.10 there instead). Now it looks okay. Let's see on Jenkins:

davisagli:
@mauritsvanrees The github status update from the 3.10 job is still broken; I should have time to work on it this afternoon.

davisagli:
Wait, I am confused. The 3.10 job shows a "Pending -- Job started" status, which means it was updated from jenkins. But if I look at the job log in jenkins, I see this error when installing pygithub:
```
Collecting pygithub==1.43.2
  Using cached PyGithub-1.43.2.tar.gz (2.7 MB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      error in PyGithub setup command: use_2to3 is invalid.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
```

Ah, I think I know what is wrong. pygithub is installed in 2 different ways for the update that runs at the beginning and the update that runs at the end, and this error is probably for the one that runs at the end. I think pygithub is not yet compatible with Python 3.10

mauritsvanrees:
> Collecting pygithub==1.43.2

There is a newer version which should be better: https://pypi.org/project/PyGithub/1.55/
With the classifiers it claims compatibility with Python 3.6-3.9. 

On coredev 6.0 we pin this latest version and it ends up in `bin/fullrelease`, which I use a lot, and I use 3.10 locally. So should work.

davisagli:
@mauritsvanrees great, I will try updating to that one later today

davisagli:
@mauritsvanrees I updated the python 3.10 PR job to use pygithub 1.55, and I'm going to start a retry of it now so we can see if it helped

mauritsvanrees:
That looks good, all jobs finished with success status, and correctly reported back here:

Screenshot 2022-09-16 at 10 50 43

Thanks David!

mauritsvanrees:
Main point of this PR was testing that Jenkins works again. But I guess it is fine to merge the actual `isort` changes. :-)

PR #3645, opened 2022-09-15
Deprecate the portal_properties tool

ale-rt:
The tool gets most of the methods from OFS.PropertyManager.

The most relevant are using internally the hasProperty method.

Deprecating this should be enough to finally start removing this obsolete tool and a bunch of code that depends on that.

Refs. #125


ale-rt:

ale-rt:

gforcada:
Is there a porting guide from properties_tool to the registry? It would be great to add a link on those deprecations, so one can easily follow the link and get it ported ✨

ale-rt:
I do not think so.
What about linking #125, the issue where the discussion about removing the tool happened?

jensens:
Please always tell in depreciation message when it will be removed, like in Plone 7 (or 6.1)...

ale-rt:
The discussion about removing portal properties dates back in 2013 and [the plan was to remove it in 5.1](https://github.com/plone/Products.CMFPlone/issues/125#issuecomment-141768802).

I would target 6.1: there is really a lot of junk code that we can drop, especially in the tests.

ale-rt:

jensens:

jensens (approved):

PR #3653, opened 2022-10-05
Fix password used in a test

davisagli:


news/3653.internal
```diff
@@ -0,0 +1 @@
+- Fix password used in a test. [davisagli]
```
mauritsvanrees, line 1: The `.internal` suffix is not supported in `Products.CMFPlone`. See the difference in `pyproject.toml` with `plone.restapi` where this suffix *is* supported. `plone.restapi` is the exception here.

We could add it here as well, that is certainly a good option. But then we should add it in **all** Plone repositories. And agree on the best suffixes that we will use.
I sometimes think of creating an own version of `towncrier` specifically for Plone, where we have all shared settings. `towncrier` has all kinds of features that we do not use, because `zest.releaser` already has it.

Anyway, on master I have renamed this file to bugfix, and have removed the dash, so we are good for now. :-)

PR #3659, opened 2022-10-19
Fix the test coverage for getNavigationRoot

ale-rt:
When moving the navtree properties to the registry part of the test coverage was lost.
This patch shows that even if the registry does not contain the plone.root record a sane value is returned.

Some tests were still using the navtree properties and they were passing by chance.


ale-rt:

ale-rt:
I ended up touching this ancient code because I am cleaning up some bogus references to the unused portal_properties tool (I am on a mission to close #125 one day...).

I do not think this is worth a changelog file (correct if I am wrong).


Cherry on the cake... this test should be moved to plone.app.layout, where the `getNavigationRoot` function was moved years ago.

In short, I think it is worth merging this PR, at least I will find less things when grepping for `navtree_properties` ;)

davisagli (approved):

PR #3665, opened 2022-10-28
Don't create default content folders for Volto

### LINKED ISSUE #3628: Site Creation Volto creates Folder-type content
### What I did:

I created a Volto Site.

### What I expect to happen:

That contents of the Folder type are not created, since the Folders were removed in Volto.

### What actually happened:

Initial contents, News, Events and Users are created as Folder.

When I navigate in Volto and click on one of these items in the menu, I get the error in javascript console:

```
GET https://localhost/++api++/Members/@types/Folder 401
```

This seems to be done in:

https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/setuphandlers.py

Plone 6.0b1

davisagli:
When creating a Volto site, don't add the news, events, or users folders. They are not set up in a way that makes sense for Volto. (They are Folders, which is a disabled content type. For Volto they should be Documents with a listing block.)

In the future we can add default content that makes sense for Volto, but for now let's just avoid adding the existing default content that doesn't make sense.

Fixes #3628

### DIFF
--- a/Products/CMFPlone/browser/admin.py
+++ b/Products/CMFPlone/browser/admin.py
@@ -160,10 +160,8 @@ class AddPloneSite(BrowserView):
     # Let's have a separate list for Volto.
     volto_default_extension_profiles = (
         'plone.app.caching:default',
-        # We could choose to not install Barceloneta:
         'plonetheme.barceloneta:default',
         'plone.volto:default',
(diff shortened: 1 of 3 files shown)


davisagli:

ericof (approved):
LGTM

PR #3671, opened 2022-11-06
The class TestBaseNavTree is not used anymore

ale-rt:
The class TestBaseNavTree is not used anymore and its tests are not run. It is still assuming that the configuration is stored in the portal_properties tool, while it is in the registry since quite some time.

Another step towards #125


ale-rt:

PR #3672, opened 2022-11-06
Modernize the test testNavigationRootDoesNotOverrideExplicitPath

ale-rt:
The test was still assuming that the navigation root configuration is stored in the portal properties while it is in the registry since quite some time.

Adapt the test to check that nothing breaks when tweaking the registry.

Refs. #125


ale-rt:
I think that a changelog entry is an overkill

ale-rt:

Issue #3663, opened 2022-10-26
Doctests with HTTPError fail on Python 3.11 with traceback problem


mauritsvanrees:
TLDR: two doctest files should be rewritten to unit tests.

Long version:

I am trying the coredev buildout on Python 3.11. See https://github.com/plone/buildout.coredev/issues/821 for some troubles with this. There are actually not many test failures left. Latest full run has [5 failures](https://jenkins.plone.org/job/plone-6.0-python-3.11/120/#showFailuresLink). Three in PortalTransforms are being handled in https://github.com/plone/Products.PortalTransforms/pull/49.

This leaves two failures that seem to have the same root cause. Both `bin/test -s Products.CMFPlone -m test_functional -t redirection.txt` and `bin/test -s plone.browserlayer` fail with:

```
Error in test /Users/maurits/community/plone-coredev/311/src/Products.CMFPlone/Products/CMFPlone/tests/redirection.txt
Traceback (most recent call last):
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/doctest.py", line 2215, in runTest
    failures, tries = runner.run(
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/doctest.py", line 1497, in run
    return self.__run(test, compileflags, out)
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/doctest.py", line 1372, in __run
    exc_msg = traceback.format_exception_only(*exception[:2])[-1]
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/traceback.py", line 159, in format_exception_only
    te = TracebackException(type(value), value, None, compact=True)
  File "/Users/maurits/.pyenv/versions/3.11.0rc2/lib/python3.11/traceback.py", line 696, in __init__
    self.__notes__ = getattr(exc_value, '__notes__', None)
[...]

PR #3673, opened 2022-11-08
Fix 'KeyError: file' in browser tests on Python 3.11.

mauritsvanrees:
Fixes part of https://github.com/plone/Products.CMFPlone/issues/3663


mauritsvanrees:

mauritsvanrees:
Two checks are not reported for some reason, but they should have passed. All Jenkins tests are green, including Python 3.11!

gforcada:
I was looking at these failures a week ago and couldn't really find the root cause of it, but this re-write on plain python test seems to be working 🎉 nice job! 👍🏾

gforcada (approved):

PR #3684, opened 2022-11-29
Fix rendering of resources with expressions

petschki:


petschki:

PR #3686, opened 2022-12-01
Deprecate `plone.app.widgets`

petschki:
~~this needs https://github.com/plone/plone.app.z3cform/pull/155~~ merged 👍🏼


petschki:
After sleeping over this, I do not think it's the correct way to import from `plone.app.z3cform` ... maybe we should move the pattern configuration code out of CMFPlone to `plone.base`? Or the other way round: implement pattern specific configuration here and let other packages reuse it ?

jensens (approved):

jensens:

jensens (commented):
actually, the base idea is good and worth a try in my opinion, but maybe more eyes on this can help to get it done right?

petschki:

setup.py
```diff
@@ -75,6 +75,7 @@
         "plone.app.viewletmanager",
         "plone.app.vocabularies",
         "plone.app.workflow",
+        "plone.app.z3cform >= 4.1.0",
```
petschki, line 78: @jensens is it ok to define dependency here? Maybe this causes more circular import troubles ...

petschki:

setup.py
jensens: This is probably a problem. Need to check. A good little task for ArtSprint!

mauritsvanrees: This PR was merged, but was this conversation resolved somewhere?

petschki: Sorry for silence ... @jensens looked at it here at ArtSprint and it didn't have any circular issues.

PR #3694, opened 2022-12-05
Search controlpanel form checklist fix

danalvrz:
This PR is meant to be merged with: https://github.com/plone/plone.base/pull/24

Issue: https://github.com/plone/volto/issues/3670


danalvrz:

news/6011.bugfix
```diff
@@ -1 +0,0 @@
-Updated metadata version to 6011.  [maurits]
```
davisagli, line 1: @danalvrz This file should not be removed in this PR.

news/3694.bugfix
```diff
@@ -0,0 +1 @@
+Replace ReverseCheckBoxFieldWidget with CheckBoxFieldWidget in SearchControlPanelForm
```
davisagli, line 1: @danalvrz release notes should focus on what the end result is for the user, not on how the implementation was done. Also please include your github username for credit. So I would suggest:
```suggestion
Change the search control panel to select types not searched instead of searchable types. This fixes an inconsistency with Volto. [danalvrz]
```

mauritsvanrees:
I had no idea we had a special `ReverseCheckBoxFieldWidget` in CMFPlone for this. :-)

For the record (and for people who like to update screen shots in documentation), this is before:

Screenshot 2022-12-08 at 10 18 04

and this is after:

Screenshot 2022-12-08 at 10 15 18

Quite a UI change, a few days before Plone 6 final. But as release manager I approve. :-) Much more intuitive.
Thanks!

yurj:
The registry will contain type to be searched while the ui will show the reverse using actual types? :-/

mauritsvanrees:
> The registry will contain type to be searched while the ui will show the reverse using actual types? :-/

No. The field is `types_not_searched` and only the types that should not be searched, are stored here. That was always the case, and has not changed.
But until now, the UI showed this in reverse. With this PR, that is fixed: the UI is a one-on-one representation of what is actually stored.

davisagli:
> the UI is a one-on-one representation of what is actually stored

And now it is the same in both the Classic UI and Volto, instead of being reversed in the Classic UI only, and one-on-one with incorrect help text in Volto. 😜

PR #3696, opened 2022-12-06
Fix accessibility info

yurj:
Use label_site_administration instead of label_site_admin because of the missing article. Must be checked over other languages but italian and english would be ok and the fix is the same of pull https://github.com/plone/Products.CMFPlone/pull/3397

### DIFF
--- a/Products/CMFPlone/browser/templates/accessibility-info.pt
+++ b/Products/CMFPlone/browser/templates/accessibility-info.pt
@@ -47,7 +47,7 @@
         If this web site does not validate correctly, please contact the
         <span i18n:name="site_admin"><a href="/contact-info"
               tal:attributes="href string:${context/portal_url}/contact-info"
-              i18n:translate="llabel_site_administration">site administration</a></span>.
+              i18n:translate="label_site_administration">site administration</a></span>.
         </p>
 
         <p i18n:translate="description_wcag_aa_rating">
--- a/Products/CMFPlone/browser/templates/accessibility-info.pt
+++ b/Products/CMFPlone/browser/templates/accessibility-info.pt
@@ -47,7 +47,7 @@
         If this web site does not validate correctly, please contact the
         <span i18n:name="site_admin"><a href="/contact-info"
               tal:attributes="href string:${context/portal_url}/contact-info"
-              i18n:translate="label_site_admin">site administration</a></span>.
+              i18n:translate="llabel_site_administration">site administration</a></span>.
         </p>
 
         <p i18n:translate="description_wcag_aa_rating">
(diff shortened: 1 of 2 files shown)


yurj:

erral:
Why is this needed?

yurj:
> Why is this needed?

In Italian I get:
"Il sito usa assistive technology, come i ruoli WAI-ARIA, nel rispetto delle migliori pratiche definite ad oggi. In ogni caso il sito, gli standard e i contenuti variano nel tempo. Se questo sito web non passa correttamente la validazione, per favore contatta l'l'amministrazione del sito." <- there's a double l'l'

with the commit, I get:
"Il sito usa assistive technology, come i ruoli WAI-ARIA, nel rispetto delle migliori pratiche definite ad oggi. In ogni caso il sito, gli standard e i contenuti variano nel tempo. Se questo sito web non passa correttamente la validazione, per favore contatta l'amministrazione del sito." <- no double

The same fix was done in https://github.com/plone/Products.CMFPlone/pull/3397 with the same scenario. Other languages seems already ok but the Italian has this article with the apostrophe in front of feminine nouns. I don't know if this is the right fix or the right place to fix it.

erral (approved):

erral:
@mauritsvanrees should we merge this?

mauritsvanrees:
I wondered what "article" meant, but I slowly realised this means: "the" or "an".

This change is not good for Dutch. There we have these translations:

```
msgid "label_site_admin"
msgstr "websitebeheerder"

msgid "label_site_administration"
msgstr "Website beheer"

msgid "description_validation"
msgstr "... de ${site_admin}."
```

Here, "de" is a Dutch article.

So with the code from master we get: "...de websitebeheerder", which is correct Dutch.
With your code we get "de Website beheer", which is wrong. So I would need to edit the Dutch `description_validation` message to remove the "de" article. Or maybe replace "de" with another article "het", but that reads a bit awkward.

So: please do not merge this. It is bad for one language, and maybe more.

Instead I would say, just edit the Italian po file: instead of ending the `description_validation` translation with `contatta l'${site_admin}.`, let it end with  `contatta ${site_admin}.`

yurj:
I agree, can you also check https://github.com/plone/Products.CMFPlone/pull/3397 if it has to be reverted/changed too?

mauritsvanrees:
> I agree, can you also check #3397 if it has to be reverted/changed too?

Yes, a revert for this would be better for Dutch, although I should change one Dutch related message anyway.

The danger is that some translators have already adapted the new translations to work better together, and they might need to change them again after we revert this. :-/

Also, for me "site administration" sounds both like an activity ("I do site administration") and like a person or team ("I am the site administration" or "I belong to the site administration"). If you just translate the po files without context, you don't really know which of the two it is. That makes it difficult.

I wonder if there are many po files where the translations for these two strings are different. If they are only a few, then we could also pick either `label_site_admin` or  `label_site_administration` and use only that one everywhere.

PR #3700, opened 2022-12-10
Update for Plone 6 final release

stevepiercy:
Fix broken URLs. This PR still needs a URL to a form where people can submit a success story. I just pointed to existing content that I could find.


stevepiercy:
I made a changelog entry through the GitHub UI, so why does it complain?

davisagli:
@stevepiercy you put the changelog at the top level instead of in the news folder

stevepiercy:
> @stevepiercy you put the changelog at the top level instead of in the news folder

Derp. Of course.

I've changed this to draft and removed reviewers, as I also learned that I need to update translations, I made other mistakes changing control panel links from relative to absolute to root, and I won't have time to get to it until next week at the earliest. If someone else has free time and wants to take it over, please be my guest!

fredvd:
I have reverted the absolute links to their relative versions and moved the changelog item to the news folder. 

If I'm not mistaken translations will have to be extract first from this page template when it is merged, so that will be a separate and subsequent task,  @mauritsvanrees ,  @erral ?

fredvd:

erral:
> I have reverted the absolute links to their relative versions and moved the changelog item to the news folder.
> 
> If I'm not mistaken translations will have to be extract first from this page template when it is merged, so that will be a separate and subsequent task, @mauritsvanrees , @erral ?

That's right. After merge we will need to regenerate the translation files, ask for translations modifications and so on.

We can release a new plone.app.locales but it will contain just 2 or 3 translations. 

We can release more plone.app.locales version for the subsequent 6.0.1, 6.0.2, etc.

fredvd:
The 3.9 test failure is fluke.

erral (approved):
LGTM

mauritsvanrees:
@erral Does updating `plone.app.locales` with this mean that a lot of translations will be marked as fuzzy and thus not work? In that case I would prefer not including this in `plone.app.locales`.
Or maybe remove the fuzzy markers temporarily.

erral:
I don't know what will happen,let's see what happens when we resync them

erral:
As plonefrontpage domain translations are handled manually, I have manually updated the new default text in the po files so there are no fuzzy markers.

If some translation is not updated it will show the old text.

mauritsvanrees:
> As plonefrontpage domain translations are handled manually, I have manually updated the new default text in the po files so there are no fuzzy markers.
> 
> If some translation is not updated it will show the old text.

Perfect, thanks!

PR #3702, opened 2022-12-12
Add explanation and refer to documentation for Volto/ClassicUI sites

fredvd:
And relabel Classic to ClassicUI


erral:
Adding this requires resyncing and releasing a new version of plone.app.locales

erral (approved):

PR #3719, opened 2023-02-02
Show `modal` property when editing an action

### LINKED ISSUE #3709: Can't edit the modal property of an action in Plone UI
Most settings of actions can be edited in the actions control panel in the Plone UI. See the classic demo:
https://6-classic.demo.plone.org/@@actions-controlpanel

Some actions have a modal property, for example the `object_buttons/delete` action. You can see it in the ZMI:
https://6-classic.demo.plone.org/portal_actions/object_buttons/delete/manage_propertiesForm
The property value is this:

```
{"actionOptions": {"disableAjaxFormSubmit":true, "redirectOnResponse":true}}
```

It would be nice if this could also be edited in the control panel.  Currently this property is now shown at all.
I guess there could be other non-standard properties as well, so bonus points if this shows all properties.

My use case today was actually on Plone 5.2 where I wanted to change the modal property of `user/login`. In Plone 6.0.0 this is an empty dictionary, but on 5.2 it was this:

```
{"prependContent": ".portalMessage", "title": "Log in", "width": "26em", "actionOptions": {"redirectOnResponse": true}}
```

I had to change the width to 18em in a client project today because their base font size was a lot bigger, which led to the login modal being only half visible on mobile. :-)

With my release manager hat on: no I don't want this changed in 5.2, people will have to use the ZMI there. Depending on scale and impact of the needed changes, a fix for this could be done either in a 6.0.x bugfix release, or in 6.1.

petschki:
fixes #3709

This needs https://github.com/plone/plone.base/pull/27

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/actions.py
+++ b/Products/CMFPlone/controlpanel/browser/actions.py
@@ -14,6 +14,8 @@ from zope.event import notify
 from zope.interface import implementer
 from zope.lifecycleevent import ObjectCreatedEvent
 
+import json
+
 
 class ActionListControlPanel(BrowserView):
     """Control panel for the portal actions."""
@@ -162,6 +164,14 @@ class ActionControlPanelAdapter:
         self.current_category._objects = tuple(all_actions)
 
     position = property(get_position, set_position)
+
+    def get_modal(self):
+        return json.loads(self.context.modal)
+
+    def set_modal(self, value):
+        self.context.modal = json.dumps(value)
+
+    modal = property(get_modal, set_modal)
 
 
 class ActionControlPanel(AutoExtensibleForm, form.EditForm):
--- a/Products/CMFPlone/controlpanel/browser/actions.py
+++ b/Products/CMFPlone/controlpanel/browser/actions.py
@@ -1,6 +1,7 @@
 from plone.autoform.form import AutoExtensibleForm
 from plone.base.interfaces import IActionSchema
 from plone.base.interfaces import INewActionSchema
+from plone.base.utils import base_hasattr
 from Products.CMFCore.ActionInformation import Action
 from Products.CMFCore.interfaces import IAction
 from Products.CMFCore.interfaces import IActionCategory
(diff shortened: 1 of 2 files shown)


Products/CMFPlone/controlpanel/browser/actions.py
```diff
@@ -163,6 +165,14 @@ def set_position(self, value):
 
     position = property(get_position, set_position)
 
+    def get_modal(self):
+        return json.loads(self.context.modal)
+
+    def set_modal(self, value):
+        self.context.modal = json.dumps(value)
```
petschki, line 172: @mauritsvanrees not sure if this loads/dumps is needed for the textinput here ... the JSON is validated in `plone.base.interfaces.controlpanel.IActionSchema`

mauritsvanrees: No, it is not needed. With this code, I actually get a validation error when I edit the Delete action and save it without changes: "Expecting property name enclosed in double quotes."
Ah, in the ZMI the value is:

    {"actionOptions": {"disableAjaxFormSubmit":true, "redirectOnResponse":true}}

With the current code in this PR, this shows up in the Plone control panel as:

    {'actionOptions': {'disableAjaxFormSubmit': True, 'redirectOnResponse': True}}

And this is no json, due to the single quotes and the Python-style booleans.
Funny. :-)

So just use `return self.context.modal` and `self.context.modal = value`. Then it works for me.
I have pushed a fix.

mauritsvanrees (approved):
Thanks for working on this!
I answered your remark in a comment, and pushed a fix.

But the setter is still not completely right yet. It works when the property already exists, but not all actions have the modal property. I tried it with the Paste action in Object buttons. In the control panel I can set a modal value and save it, but it does not show up in the ZMI.

Other parts of the code here call `self.context._setPropValue`, but that does not do it either. An attribute `modal` is set on the context, with the value, but it does not show up in the ZMI. The modal is then not active when I use the paste action: having an attribute is not enough, it has to be a property.

Okay, I have fixed that too, and now the property is there in the ZMI, and the modal is used when pasting.
Can you try it out?

petschki:
works here too ... I'll start Jenkins.

PR #3722, opened 2023-02-09
Removed dependencies on development versions. [5.2]

mauritsvanrees:
For example we were depending on `plone.app.contentmenu >= 1.1.6dev-r22380` and the 1.1.6 version was released in 2008. With setuptools 67 this development version is illegal and cannot be parsed, even though in `versions.cfg` or `constraints.txt` we explicitly request version 2.3.4.

Sample error is in this [exportimport test run](https://github.com/collective/collective.exportimport/actions/runs/4135262761/jobs/7147478436#step:9:55):

```
pkg_resources.extern.packaging._tokenizer.ParserSyntaxError: Expected end or semicolon (after version specifier)
    plone.app.contentmenu>=1.1.6dev-r22380
```

I updated a few alpha/beta versions as well. They should not cause problems, it just does not look so nice.


mauritsvanrees:

ale-rt (approved):

PR #3726, opened 2023-02-24
Apply Barceloneta upgrades when upgrading Plone.

mauritsvanrees:
An upgrade step was added in plonetheme.barceloneta in September 2022. Also an upgrade step in plone.app.upgrade to re-apply the registry from plonetheme.barceloneta. But in portal_setup Upgrades the package was still marked as upgradable, because its own upgrade step was not run. Just run any upgrade automatically from now on.


mauritsvanrees:

ale-rt (approved):

PR #3747, opened 2023-03-21
Fix circular dependency in p.a.theming on ZCML level

jensens:
Move its controlpanel permission over there. see https://github.com/plone/plone.app.theming/pull/217
Needs test/merge together with this issue.


gforcada (approved):

mauritsvanrees:
I have merged both PRs.

PR #3752, opened 2023-03-28
Show warning in Site Setup when using unsupported Plone/Python version [5.2]

mauritsvanrees:
This is based on dates in the [Plone release schedule](https://plone.org/download/release-schedule) and the [Status of Python versions](https://devguide.python.org/versions/).

I want something similar in Plone 6.0, but there are no definitive dates there yet, so let's leave that for now.

This is how it looks on 5.2 when I temporarily hack it so all warnings are shown.  This gets ugly, but I have no problems with that.

Screenshot 2023-03-28 at 18 47 42

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/overview.pt
+++ b/Products/CMFPlone/controlpanel/browser/overview.pt
@@ -23,6 +23,69 @@
 
     <div class="portalMessage warning"
          role="status"
+         tal:condition="view/python2_warning">
+        <strong i18n:translate="">
+            Warning
+        </strong>
+        <span tal:omit-tag="" i18n:translate="text_python2_warning">
+            You are using Python 2.
+            This should only be used temporarily while preparing to migrate to Python 3.
+        </span>
+    </div>
+
+    <div class="portalMessage warning"
+         role="status"
+         tal:condition="view/python_warning">
+        <strong i18n:translate="">
+            Warning
+        </strong>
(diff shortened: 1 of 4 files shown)


mauritsvanrees:

davisagli:
Nice idea. Can we provide links to more information for more of the notices?

mauritsvanrees:
The last warning in the screen shot links to https://plone.org/download/release-schedule.  This warning is always shown as extra warning when at least one of these warnings is shown.  That link should contain all the info that is needed.

I went a bit back and forth between showing this link embedded in all of the warnings, but that got rather verbose, especially for translators.

BTW, we could also link to https://plone.org/release-schedule, which is shorter.  This is a Link content type and points to the real content at `/download/release-schedule`.  I suppose it is possible that during some restructuring years from now the `/download/release-schedule` gets moved somewhere else, and it is slightly more likely that `/release-schedule` will still be at the same spot, where we can edit the link.  But moving content of course is tracked in the redirection tool, and if needed we can manually add such a redirect as well.  So I opted for the slightly longer url, avoiding a redirect.

jensens:
+1 for this.

gforcada (approved):
Great idea, I do like to get reminders on jenkins.plone.org about new versions, or plugins that need updates, etc etc, so having these warnings is a great idea! 👍🏾

PR #3753, opened 2023-03-28
Don't set "path" for Plone site root

malthe:
It doesn't make sense to add a path query targeting the site root – that's a "noop", only making things slower.

Note that this also helps [collective.multilingual](https://github.com/collective/collective.multilingual/) which has "path" as one of its default exclusion query arguments (that is, if path is provided, we don't add a language filter.)

### DIFF
--- a/Products/CMFPlone/browser/search.py
+++ b/Products/CMFPlone/browser/search.py
@@ -2,11 +2,11 @@ from DateTime import DateTime
 from plone.app.contentlisting.interfaces import IContentListing
 from plone.app.layout.navigation.interfaces import INavigationRoot
 from plone.base.batch import Batch
+from plone.base.interfaces.siteroot import IPloneSiteRoot
 from plone.base.interfaces import ISearchSchema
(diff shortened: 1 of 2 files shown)


malthe:

Products/CMFPlone/browser/search.py
```diff
@@ -6,6 +6,7 @@
 from plone.registry.interfaces import IRegistry
 from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone.browser.navtree import getNavigationRoot
+from Products.CMFPlone.interfaces import IPloneSiteRoot
```
petschki, line 9: Interfaces should be imported from `plone.base.interfaces`

malthe: Fixed in df97eceab7257855935d95e2f2064690857e9e8f.

petschki:

gforcada (approved):

petschki:

PR #3758, opened 2023-03-30
Fix repairing relations

### LINKED ISSUE #3757: Fix broken relations

Content with broken relations should be repaired by `Products.relationhelper.rebuild_relations` (flush of intIds or no flush of intIds)

### What I did:

Create page 1
Create page 2
Make page 2 related content of page 1.
Delete page 2.
Execute fix of relations (flush or no flush of intIds) via control panel relations. This calls `Products.relationhelper.rebuild_relations`.

### What I expect to happen:

The broken relation is removed by `Products.relationhelper.rebuild_relations`. 

### What actually happened:

The broken relation is NOT removed by `Products.relationhelper.rebuild_relations`. 

Plone 6.0.3

ksuess:
Fix #3757

### DIFF
--- a/Products/CMFPlone/relationhelper.py
+++ b/Products/CMFPlone/relationhelper.py
@@ -40,7 +40,7 @@ else:
 
 logger = logging.getLogger(__name__)
 
-RELATIONS_KEY = 'ALL_REFERENCES'
+RELATIONS_KEY = "ALL_REFERENCES"
 
 
 def rebuild_relations(context=None, flush_and_rebuild_intids=False):
@@ -62,8 +62,8 @@ def get_relations_stats():
         try:
             rel = relation_catalog.resolveRelationToken(token)
         except ObjectMissingError:
-            broken['Object is missing'] += 1
-            logger.info(f'Token {token} has no object.')
+            broken["Object is missing"] += 1
+            logger.info(f"Token {token} has no object.")
             continue
 
         if rel.isBroken():
@@ -74,8 +74,13 @@ def get_relations_stats():
 
 
 def get_all_relations():
-    """Get all data from zc.relation catalog.
-    Logs some useful statistics.
+    """Get relations from zc.relation catalog.
+
+    Log statistics.
+    Return list of dictionaries:
+        from_uuid: source UID
+        to_uuid: target UID
+        from_attribute: relation name
     """
     results = []
     info = defaultdict(int)
@@ -85,55 +90,53 @@ def get_all_relations():
         try:
             rel = relation_catalog.resolveRelationToken(token)
         except ObjectMissingError:
-            logger.info(f'Token {token} has no object.')
+            logger.info(f"Token {token} has no object.")
             continue
 
-        if rel.from_object and rel.to_object:
-            try:
-                results.append({
-                    'from_uuid': rel.from_object.UID(),
-                    'to_uuid': rel.to_object.UID(),
-                    'from_attribute': rel.from_attribute,
-                })
-                info[rel.from_attribute] += 1
-            except AttributeError as ex:
-                logger.info(f'Something went wrong while storing {rel}: \n {ex}')
-        else:
-            logger.info(f'Dropping relation {rel.from_attribute} from {rel.from_object} to {rel.to_object}')
-    msg = ''
+        rel_uid_info = {
+            "from_attribute": rel.from_attribute,
+        }
+        rel_uid_info["from_uuid"] = rel.from_object and rel.from_object.UID() or None
+        try:
+            rel_uid_info["to_uuid"] = rel.to_object and rel.to_object.UID() or None
+        except AttributeError as e:
+            rel_uid_info["to_uuid"] = None
+            logger.error(f"Broken relation: {rel} '{str(e)}'")
+        results.append(rel_uid_info)
+
+    msg = ""
     for key, value in info.items():
-        msg += f'{key}: {value}\n'
-    logger.info(f'\nFound the following relations:\n{msg}')
+        msg += f"{key}: {value}\n"
+    logger.info(f"\nFound the following relations:\n{msg}")
     return results
 
 
 def store_relations(context=None):
-    """Store all relations in a annotation on the portal.
-    """
+    """Store all relations in a annotation on the portal."""
     all_relations = get_all_relations()
     portal = getSite()
     IAnnotations(portal)[RELATIONS_KEY] = all_relations
-    logger.info(f'Stored {len(all_relations)} relations on the portal')
+    logger.info(f"Stored {len(all_relations)} relations on the portal")
 
 
 def purge_relations(context=None):
     """Removes all entries form zc.relation catalog.
+
     RelationValues that were set as attribute on content are still there!
     These are removed/overwritten when restoring the relations.
     """
-    rel_catalog = getUtility(ICatalog)
-    rel_catalog.clear()
-    logger.info('Purged zc.relation catalog')
+    relation_catalog = getUtility(ICatalog)
+    relation_catalog.clear()
+    logger.info("zc.relation catalog purged.")
 
 
 def restore_relations(context=None, all_relations=None):
-    """Restore relations from a annotation on the portal.
-    """
+    """Restore relations from an annotation on the portal."""
 
     portal = getSite()
     if all_relations is None:
         all_relations = IAnnotations(portal)[RELATIONS_KEY]
-    logger.info(f'Loaded {len(all_relations)} relations to restore')
+    logger.info(f"Loaded {len(all_relations)} relations to restore")
     update_linkintegrity = set()
     modified_items = set()
     modified_relation_lists = defaultdict(list)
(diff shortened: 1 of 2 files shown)


ksuess:

ksuess:

ksuess:

Products/CMFPlone/relationhelper.py
```diff
@@ -85,55 +90,53 @@ def get_all_relations():
...
-    msg = ''
+        rel_uid_info = {
+            "from_attribute": rel.from_attribute,
+        }
+        rel_uid_info["from_uuid"] = rel.from_object and rel.from_object.UID() or None
+        try:
+            rel_uid_info["to_uuid"] = rel.to_object and rel.to_object.UID() or None
+        except AttributeError as e:
+            rel_uid_info["to_uuid"] = None
+            logger.error(f"Broken relation: {rel} '{str(e)}'")
```
davisagli, line 104: Would it be easier to see useful information in the log if this logged rel_uid_info instead of rel?

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +151,138 @@ def restore_relations(context=None, all_relations=None):
...
-            target_obj = uuidToObject(item['to_uuid'])
+            target_obj = uuidToObject(item["to_uuid"])
         except KeyError:
             # brain exists but no object
             target_obj = None
+        finally:
+            if not target_obj:
+                logger.info(f'No target object found for UID {item["to_uuid"]}.')
+                # Do not continue. The source_obj needs to be updated.
+                # continue
```
davisagli, line 186:
```suggestion
                # The source_obj will be updated to remove the broken relation below.
```

(This is clearer to me. Using the word "continue" makes it hard to tell whether you mean continuing to the next iteration of the loop, as the `continue` statement does, or continuing to execute code below, which does happen now.)

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +151,138 @@ def restore_relations(context=None, all_relations=None):
...
-            logger.info(f'{source_obj} is no dexterity content')
+            logger.info(f"{source_obj} is no dexterity content")
             continue
 
         if not IDexterityContent.providedBy(target_obj):
-            logger.info(f'{target_obj} is no dexterity content')
-            continue
+            logger.info(f"{target_obj} is no dexterity content")
+            # Do not continue. The source_obj needs to be updated.
+            # continue
```
davisagli, line 196:
```suggestion
```

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +151,138 @@ def restore_relations(context=None, all_relations=None):
...
 
-        from_attribute = item['from_attribute']
+        # intid for target_obj exists.
```
davisagli, line 198:
```suggestion
        # Confirm intid for target_obj exists.
```

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +151,138 @@ def restore_relations(context=None, all_relations=None):
...
         try:
             to_id = intids.getId(target_obj)
         except KeyError as e:
-            logger.warning(f'No intid for {target_obj}')
-            continue
+            logger.warning(f"No intid for {target_obj}")
+            to_id = None
+            # continue
```
davisagli, line 204:
```suggestion
```

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +151,138 @@ def restore_relations(context=None, all_relations=None):
...
         if field_and_schema is None:
             # the from_attribute is no field
-            logger.info(f'No field. Setting relation: {item}')
+            logger.info(f"No field. Setting relation: {item}")
             event._setRelation(source_obj, from_attribute, RelationValue(to_id))
             continue
 
         field, schema = field_and_schema
-        relation = RelationValue(to_id)
+        relationvalue = to_id and RelationValue(to_id) or None
```
davisagli, line 230: In general I prefer to use `RESULT_IF_TRUE if CONDITION else RESULT_IF_FALSE` instead of `CONDITION and RESULT_IF_TRUE or RESULT_IF_FALSE`.

In most cases it does not make any difference, but it does if RESULT_IF_TRUE is itself a value which evaluates to False as a boolean.

davisagli (approved):
@ksuess Sorry for the delayed review. It basically looks good. I made a few minor suggestions.

Products/CMFPlone/relationhelper.py
ksuess: Definitly.

Products/CMFPlone/relationhelper.py
ksuess: Yes, _continue_ could mean both in this context. Updated with your suggestion.

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +152,133 @@ def restore_relations(context=None, all_relations=None):
...
         if from_attribute == referencedRelationship:
-            # Ignore linkintegrity for now. We'll rebuilt it at the end!
-            update_linkintegrity.add(item['from_uuid'])
+            update_linkintegrity.add(item["from_uuid"])
             continue
 
+        # Working copy relations
         if HAS_ITERATE and from_attribute == ITERATE_RELATION_NAME:
             # Iterate relations are not set as values of fields
             relation = StagingRelationValue(to_id)
```
mauritsvanrees, line 211: `to_id` could be `None` after your changes.  I think no staging relation should be created then.

mauritsvanrees (commented):
Thanks!  Mostly looks good from reading the code.  I added a few comments.

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +152,133 @@ def restore_relations(context=None, all_relations=None):
...
+        logger.info(f"Dropping {len(all_relations) - len(unique_relations)} duplicates")
         all_relations = unique_relations
 
     intids = getUtility(IIntIds)
     for index, item in enumerate(all_relations, start=1):
         if not index % 500:
-            logger.info(f'Restored {index} of {len(all_relations)} relations...')
+            logger.info(f"Restored {index} of {len(all_relations)} relations...")
 
+        # source object for UID exists.
```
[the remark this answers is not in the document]
ksuess: Yes, `item["from_uuid"] is None` should be checked.
In fact `KeyError` will never be raised, due to the new `get_all_relations`.
And as `uuidToObject` does return None if no object found for UID, the snippet can be  simplified.

Products/CMFPlone/relationhelper.py
```diff
@@ -148,101 +152,133 @@ def restore_relations(context=None, all_relations=None):
...
         try:
-            source_obj = uuidToObject(item['from_uuid'])
+            source_obj = uuidToObject(item["from_uuid"])
         except KeyError:
             # brain exists but no object
             source_obj = None
+        finally:
```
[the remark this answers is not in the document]
ksuess: Updated

Products/CMFPlone/relationhelper.py
```diff
@@ -132,71 +139,62 @@ def purge_relations(context=None):
...
+            continue
+        else:
             source_obj = uuidToObject(item["from_uuid"])
-        except KeyError:
-            # brain exists but no object
-            source_obj = None
-        finally:
-            if not source_obj:
-                logger.info(f'No source object found for UID {item["from_uuid"]}.')
-                continue
```
mauritsvanrees: Why are you removing the try/except KeyError here and further on?  If the portal catalog or relation catalog are out of sync with the object database or with each other, you may very well see KeyErrors.

I suppose if you call `rebuild_relations` it will work, because that calls `store_relations` and `restore_relations` after one another.  But if in custom code you call `store_relations`, then remove a content item with a relation, and then call `restore_relations`, you will get a KeyError.

mauritsvanrees (changes requested):
The changes seem okay, except for a removal of try/except KeyError.  See comment inline.

Products/CMFPlone/relationhelper.py
davisagli: @ksuess Does this case still need to be handled?

ksuess: > @ksuess Does this case still need to be handled?

@david I'll continue working in this today, writing tests and responding to change requests.

ksuess: @mauritsvanrees I could be wrong, but I do not see how a KeyError should be raised in https://github.com/plone/Products.CMFPlone/pull/3758/files/d3c50e2a60091b933aa8517c692c1233407ba8dd..dabed3cf67bf7e27bb361d62fbfca6b6a6ff9d59#diff-6d53467a1d2dd42d829f6bf8c87ed4a32aa04dea987f2d92ed34ed4a33c87f28L168 
If there is no object found, uuidToObject returns _None_.
I added a test. 
https://github.com/plone/Products.CMFPlone/pull/3758/files#diff-1bade6b3c76ae89fc7f34cf3acb70cefdfb738529c4ae263d69ad258750a8c6bR141-R178
Please have a look if I understood your objection correctly or tell me what you mean instead. Thank you.

ksuess:

Products/CMFPlone/relationhelper.py
mauritsvanrees: `uuidToObject` calls `uuidToPhysicalPath`.  This tries to get an "rid" ('reverse id' I think) from the UID index.  If this gives a wrong answer, then the [call to `catalog.getpath`](https://github.com/plone/plone.app.uuid/blob/2.2.2/plone/app/uuid/utils.py#L57) will give a KeyError:

```
>>> app.Plone.portal_catalog._catalog.getpath(12345)
  File "<console>", line 1, in <module>
  File "/Users/maurits/shared-eggs/cp311/Products.ZCatalog-7.0-py3.11.egg/Products/ZCatalog/ZCatalog.py", line 520, in getpath
    return self._catalog.paths[rid]
           ~~~~~~~~~~~~~~~~~~~^^^^^
KeyError: 12345
```

No, this should never happen.
But yes, I _have_ seen this happen in practice, and have written a script to [fix such a broken catalog](https://github.com/zestsoftware/plonescripts/blob/master/fix_uid_index.py). According to the readme in that repo: "This can at least happen when you import a zexp twice, in different folders."

So I would prefer to keep the `try/except KeyError`. :-)

ksuess:

Products/CMFPlone/relationhelper.py
ksuess: Done.
Catch KeyError of uuidToObject: uuid is indexed in UID, but catalog cannot resolve rid to path.
https://github.com/plone/Products.CMFPlone/pull/3758/commits/fe1ed81c9a7eb19b889ff319e51204ce040c3ff5

mauritsvanrees (approved):

mauritsvanrees:
This looks fine now.  Thank you Katja!

Do you want to create a PR for CMFPlone 6.0.x as well?  There should not be differences yet between the two branches.

PR #3774, opened 2023-05-13
Test, please ignore

gforcada:
Which jenkins jobs will be advertised? hopefully only 

- Plone 5.2 on python 2.7
- Plone 5.2 on python 3.8

🍀


gforcada:
@mauritsvanrees I had to remove the manual requirement on the branch protection so that the `Plone 5.2 on python 3.7` would not be required anymore.

PR #3775, opened 2023-05-17
Config with default template

gforcada:
Only formatting and QA so far 😅 

There are plenty of commits, but ignore the `black` one that is huuuuuuge.

PR #3776, opened 2023-05-17
Config with default template 4eba7476

mauritsvanrees:
This recreates PR #3775 which I had accidentally merged too soon.

BTW, some tests are failing, investigating now.


Products/CMFPlone/RegistrationTool.py
```diff
@@ -226,8 +226,8 @@ def pasValidation(self, property, password):
                     err = error
                 else:
                     msgid = _(
-                        "${sentances}. ${sentance}",
-                        mapping={"sentances": err, "sentance": error},
+                        "${sentences}. ${sentence}",
+                        mapping={"sentences": err, "sentence": error},
```
mauritsvanrees, line 229: @erral `codespell` was run, and this may mean that similar changes are needed in `plone.app.locales`.  Currently this PR is for CMFPlone master, which since today is only for Plone 6.1.  But this PR should be back ported soon to new branch 6.0.x.

mauritsvanrees:
@gforcada I see problems with various templates, at least `controlpanel/browser/types.pt` and `controlpanel/browser/usergroups_groupmembership.pt`.  I am inclined to remove all zpretty changes from this PR, and retry that part later (where later could easily be tomorrow, I don't want to drag this on).  Also remove zpretty from pre-commit for now, otherwise it will fail.  What do you think?

gforcada:
@mauritsvanrees all good 👍🏾 I applied `zpretty` in plenty of small chunks so we can disable `zpretty` but keep some of the changes already, to limit the number of files that still need to be parsed.

I did not have time yesterday to look at the test failures, I did see them, but I was busy talking with you and others during the sprint 😅

Products/CMFPlone/RegistrationTool.py
erral: OK. Some weeks ago I also modified manually the translations in such cases, so I would need to do it again when this is backported to 6.0.x.

Thanks for the heads up.

mauritsvanrees:
I close this old PR in favour of the new one: #4121.

PR #3777, opened 2023-05-17
Config with default template without zpretty

mauritsvanrees:
This is the same as PR #3776 but without the `zpretty` changes.


mauritsvanrees:

PR #3778, opened 2023-05-18
Remove plone.app.multilingual dependency

folix-01:


wesleybl:
@folix-01 can this not be done?

davisagli:
@wesleybl I think this was replaced by https://github.com/plone/Products.CMFPlone/pull/3780

Issue #3779, opened 2023-05-18
We need to remove the plone.app.multilingual dependency


folix-01:
The plone.app.multilingual dependency should be removed

mauritsvanrees:
It looks like three things are needed:
- [x] Load it conditionally in `Products.CMFPlone`. https://github.com/plone/Products.CMFPlone/pull/3780
- [ ] Add it to the `install_requires` of the `Plone` egg.  Only needed on Plone 6.1, but I see no problem in doing that on the current master branch used on both 6.0 and 6.1.  We could branch away though.
- [ ] `plone.restapi` has a hard dependency on plone.app.multilingual and does not yet declare it.  This was no problem until now.
    - [ ] Quickest solution is to add `plone.app.multilingual` to the `install_requires` of `plone.restapi`, that would be fine as a start.
    - [ ] Better solution would be to make `plone.restapi` use `plone.app.multilingual` only conditionally: only import it and load zcml when the package is available.

mauritsvanrees:
As far as I see, all is done.

PR #3780, opened 2023-05-18
Remove the plone.app.multilingual dependency

folix-01:
Related to https://github.com/plone/Products.CMFPlone/issues/3779

### DIFF
--- a/Products/CMFPlone/MigrationTool.py
+++ b/Products/CMFPlone/MigrationTool.py
@@ -107,7 +107,10 @@ ADDON_LIST = AddonList([
         profile_id='plone.app.iterate:default',
         check_module='plone.app.iterate'
     ),
-    Addon(profile_id='plone.app.multilingual:default'),
+    Addon(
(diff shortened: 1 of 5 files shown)


folix-01:

mauritsvanrees (approved):
LGTM, and could be merged when green.

One problem then though: this needs another change: the `Plone` egg should get `plone.app.multilingual` in its `install_requires`.  So if you install `Products.CMFPlone` you do not get this, but if you install full `Plone`, then you do.
Currently the package still gets pulled into the buildout because it is a test dependency of `plone.app.users` and `plone.i18n`.

I started a Jenkins job for 6.1.  There are three checks that say to run the jobs on 6.0, but we can ignore that.

mauritsvanrees:
Wait, if I use the coredev buildout and start the instance, it fails with an ImportError in plone.restapi.  So this cannot be merged yet.  I will add a comment to issue #3779.

PR #3781, opened 2023-05-18
Enable explicit acquisition in Plone 7+

jaroel:
Set PUBLISHING_EXPLICIT_ACQUISITION=true to opt-in.


jaroel:

jensens (approved):
LGTM. Merge only after we are using Products.CMFCore with the https://github.com/zopefoundation/Products.CMFCore/pull/129 published

jaroel:

gforcada:
Should we turn this into a draft PR, just to be sure it does not get merged accidentally? 😅

icemac:
I just released https://pypi.org/project/Products.CMFCore/3.1/ including the PR needed here.

jaroel:
Thanks @icemac!

gforcada:

mauritsvanrees (approved):
LGTM.  One suggestion about the changelog, because we are not opting in to explicit acquisition on Plone 7, but are opting out on Plone 6.

I think this makes sense to merge in 6.0 as well.  Earlier I tried using CMFCore 3.1 on 6.0, but [ran into a test failure](https://github.com/plone/buildout.coredev/pull/865#issuecomment-1590765809) so I reverted it.  I cannot reproduce this locally at the moment, but maybe Jenkins is somehow more picky.

With this PR merged into 6.0 as well (needs to be combined with a version update of CMFCore), I would be happy, because that means we can upgrade to CMFCore 3.2, which has a security fix.  If we would need to stay on CMFCore 3.0, we would need to add a (small) patch in CMFPlone.
Do others agree that this PR is fine for 6.0 as well?

mauritsvanrees:
Oops, I updated coredev 6.0 directly to use CMFCore 3.2, instead of on a branch for a PR.  See https://github.com/plone/buildout.coredev/commit/1c79d081c17a0710ce9a05395455861f6719560c.
Let's see what Jenkins says. Anyone should feel free to revert the commit if Jenkins breaks.

mauritsvanrees (approved):

mauritsvanrees:
Jenkins is green for Plone 6.0 with CMFCore 3.2, so it seems okay.  Still, having the above PR merged to CMFPlone 6.0 as well would be good.

mauritsvanrees:
@jaroel could you make a PR for CMFPlone 6.0?

PR #3796, opened 2023-06-06
Fix alerts to follow Bootstrap convention

petschki:
Since `plonetheme.barceloneta>=3.1.x` there's no more custom `alert-<xyz>` definition.

### DIFF
--- a/Products/CMFPlone/browser/templates/plone-overview.pt
+++ b/Products/CMFPlone/browser/templates/plone-overview.pt
@@ -44,7 +44,7 @@
              tal:condition="sites">
             <tal:loop tal:repeat="site sites">
                 <div tal:define="outdated python: view.outdated(site);"
-                     class="mb-3 ${python: 'p-3 alert-warning' if outdated else ''}">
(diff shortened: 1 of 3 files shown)


petschki:

petschki:
when we merge BS 5.3 into Plone 6.0.x we have to backport this too.

mauritsvanrees (approved):
Seems okay to me, but I do not understand what this has to do with Barceloneta 3.0 versus 3.1, or Bootstrap 5.2 versus 5.3.

In Plone 6.0 with Barceloneta 3.0, on the upgrade page with a small edit I see this:

Screenshot 2023-06-19 at 16 00 17

The last one is the original, with just `class="alert-success"`. The first one is what I get when I edit the html in the browser so it has `class="alert alert-success"`.
This looks prettier, and is what your PR would give us.  But it has the same effect on Barceloneta 3.0 and 3.1.
Am I missing something?

Anyway, it is prettier, so +1 from me.

mauritsvanrees:
Makes sense now, thanks for the explanation.

PR #3804, opened 2023-06-16
fix use the same default value for msgid

erral:
We are already using this msgid in plone.app.layout with a different default message, so with this PR we unify them both.


erral:

erral:
It looks like the wrong one is the other.

Issue #3811, opened 2023-06-27
Robot Scenario Reorder Folder Contents is unstable


mauritsvanrees:
This test often [fails](https://jenkins.plone.org/job/pull-request-6.0-3.11/715/testReport/junit/Products.CMFPlone.tests.test_robot/RobotTestCase/Scenario_Reorder_Folder_Contents/) with error message "Element with locator 'css=tr[data-id="doc1"]' not found."
Here is the [robot log](https://jenkins.plone.org/job/pull-request-6.0-3.11/715/robot/report/robot_log.html#s1-s13-t3).

It is [this test](https://github.com/plone/Products.CMFPlone/blob/6.0.6/Products/CMFPlone/tests/robot/test_folder_contents.robot#L35-L41).  At already fails at [line 39](https://github.com/plone/Products.CMFPlone/blob/6.0.6/Products/CMFPlone/tests/robot/test_folder_contents.robot#L39) before actually reordering anything.  I don't know if this is the line where it always goes wrong.

The [screen shot](https://jenkins.plone.org/job/pull-request-6.0-3.11/715/robot/report/robot_test_folder_contents_Scenario_Reorder_Folder_Contents_selenium-screenshot-1.png) shows that all four documents are there.
I guess before [checking the order](https://github.com/plone/Products.CMFPlone/blob/6.0.6/Products/CMFPlone/tests/robot/test_folder_contents.robot#L97-L101), we should wait until the elements are there.  That should be doable.  Let me try.

PR #3812, opened 2023-06-27
Fix unstable robot test scenario Reorder Folder Contents.

mauritsvanrees:
Fixes https://github.com/plone/Products.CMFPlone/issues/3811 by waiting until elements are visible.

### DIFF
--- a/Products/CMFPlone/tests/robot/test_folder_contents.robot
+++ b/Products/CMFPlone/tests/robot/test_folder_contents.robot
@@ -96,6 +96,10 @@ I reorder the elements
 
 The Order Should Be
     [Arguments]  ${first}  ${second}  ${third}  ${fourth}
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

davisagli (approved):
Does it need to be ported to the master branch as well?

mauritsvanrees:
> Does it need to be ported to the master branch as well?

Yes. But there are no differences yet between 6.0.x and master, so I just merged 6.0.x to master and pushed.  For now that is easiest and still seems safe, even without doing a PR.

gforcada:
playwright is coming soon, I ported the plone.app.discussion tests already, and fixed/enhanced them on the way. Next important part would be to enable Jenkins to run them as well 😄

PR #3830, opened 2023-07-30
Fix deprecated imports

petschki:
Note: the `Select2FieldWidget` import only works in `p.a.z3cform` master branch wich targets `CMFPlone` master branch (=6.1)


petschki:

PR #3834, opened 2023-08-18
Draft: Experiment with optimizing moves

davisagli:
I'm opening this to try out an optimization to make moving large folders faster by avoiding complete reindexing.

This is based on https://github.com/4teamwork/ftw.copymovepatches

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -290,6 +290,17 @@ class CatalogTool(PloneBaseTool, BaseTool):
         result.append("Anonymous")
         return result
 
+    def __url(self, ob):
+        try:
+            return aq_base(ob).UID()
+        except AttributeError:
+            return "/".join(ob.getPhysicalPath())
+
+    @security.protected(SearchZCatalog)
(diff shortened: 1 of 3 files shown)


davisagli:

davisagli:

davisagli:

mamico:
In my opinion, we should provide a way to configure the list of context-aware indexes (e.g., via ZCA, environment variables, or registry). This would allow developers who implement a new context-influenced index to easily register it alongside the existing ones (path, getId, etc.).

@davisagli Would you be okay with me integrating this directly into the PR? I'd be happy to handle it as soon as I can find some time.

davisagli:
@mamico In the long run I'd like to see this get refactored into a PR for CMFCore so that we don't need a monkeypatch.

mamico:
> @mamico In the long run I'd like to see this get refactored into a PR for CMFCore so that we don't need a monkeypatch.

@davisagli If you have some time and want to take a look, here is an initial implementation: https://github.com/zopefoundation/Products.CMFCore/pull/161 It's not fully complete or tested yet

PR #3844, opened 2023-09-21
Enable resource caching per context to fix context aware expressions

### LINKED ISSUE #3789: Resource Registry is missing resources in production mode when expressions are used
### What I did:

register a resource bundle with an expression like:

``python:object.portal_type == 'File'``

and make sure that the resource registry is in production mode.

### What I expect to happen:

The resource bundles should be loaded.

### What actually happened:

The resource bundles with the expression are missing.

Plone 6.0.4

petschki:
fixes #3789

### DIFF
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -62,9 +62,9 @@ class ResourceBase:
         hashtool.update(self._user_local_roles().encode("utf8"))
         if not getattr(self, "registry", None):
             self.registry = getUtility(IRegistry)
-            mtime = getattr(self.registry, _RESOURCE_REGISTRY_MTIME, None)
-            if mtime is not None:
-                hashtool.update(str(mtime).encode("utf8"))
+        mtime = getattr(self.registry, _RESOURCE_REGISTRY_MTIME, None)
+        if mtime is not None:
+            hashtool.update(str(mtime).encode("utf8"))
         return f"_v_rendered_cache_{hashtool.hexdigest()}"
 
     @property
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -50,7 +50,7 @@ class ResourceBase:
     def _user_local_roles(self):
         portal_membership = getToolByName(getSite(), "portal_membership")
         user = portal_membership.getAuthenticatedMember()
-        return "|".join(user.getRolesInContext(self.context))
+        return "|".join(sorted(set(user.getRolesInContext(self.context))))
 
     def _cache_attr_name(self):
         hashtool = hashlib.sha256()
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -20,7 +20,6 @@ import webresource
 
 logger = logging.getLogger(__name__)
 
-REQUEST_CACHE_KEY = "_WEBRESOURCE_CACHE_"
 _RESOURCE_REGISTRY_MTIME = "__RESOURCE_REGISTRY_MTIME"
 GRACEFUL_DEPENDENCY_REWRITE = {
     "plone-base": "plone",
@@ -84,7 +83,9 @@ class ResourceBase:
 
     def update(self):
         # cache on request
-        cached = getattr(self.request, self._cache_attr_name(), None)
+        REQUEST_CACHE_KEY = self._cache_attr_name()
+
+        cached = getattr(self.request, REQUEST_CACHE_KEY, None)
         if cached is not None:
             self.renderer = cached
             return
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -48,19 +48,19 @@ class ResourceBase:
             request_disabled_bundles.update(getattr(request, "disabled_bundles", []))
         return request_enabled_bundles, request_disabled_bundles
 
-    def _user_local_roles(self, site):
-        portal_membership = getToolByName(site, "portal_membership")
+    def _user_local_roles(self):
+        portal_membership = getToolByName(getSite(), "portal_membership")
         user = portal_membership.getAuthenticatedMember()
         return "|".join(user.getRolesInContext(self.context))
 
-    def _cache_attr_name(self, site):
+    def _cache_attr_name(self):
         hashtool = hashlib.sha256()
         hashtool.update(self.__class__.__name__.encode("utf8"))
-        hashtool.update(site.absolute_url().encode("utf8"))
+        hashtool.update(self.context.absolute_url().encode("utf8"))
         e_bundles, d_bundles = self._request_bundles()
         for bundle in e_bundles | d_bundles:
             hashtool.update(bundle.encode("utf8"))
-        hashtool.update(self._user_local_roles(site).encode("utf8"))
+        hashtool.update(self._user_local_roles().encode("utf8"))
         if not getattr(self, "registry", None):
             self.registry = getUtility(IRegistry)
             mtime = getattr(self.registry, _RESOURCE_REGISTRY_MTIME, None)
@@ -75,16 +75,16 @@ class ResourceBase:
         self.registry = getUtility(IRegistry)
         if not self.registry["plone.resources.development"]:
             site = getSite()
-            return getattr(site, self._cache_attr_name(site), None)
+            return getattr(site, self._cache_attr_name(), None)
 
     @_rendered_cache.setter
     def _rendered_cache(self, value):
         site = getSite()
-        setattr(site, self._cache_attr_name(site), value)
+        setattr(site, self._cache_attr_name(), value)
 
     def update(self):
         # cache on request
-        cached = getattr(self.request, REQUEST_CACHE_KEY, None)
+        cached = getattr(self.request, self._cache_attr_name(), None)
         if cached is not None:
             self.renderer = cached
             return
@@ -296,28 +296,24 @@ class ResourceBase:
 
 class ResourceView(ResourceBase, ViewletBase):
     """Viewlet Information for script rendering."""
+    resource_type = None
+
+    def index(self):
+        rendered = self._rendered_cache
+        if not rendered and self.resource_type is not None:
+            rendered = self.renderer[self.resource_type].render()
+            self._rendered_cache = rendered
+        return rendered
 
 
 class ScriptsView(ResourceView):
     """Script Viewlet."""
-
-    def index(self):
-        rendered = self._rendered_cache
-        if not rendered:
-            rendered = self.renderer["js"].render()
-            self._rendered_cache = rendered
-        return rendered
+    resource_type = "js"
 
 
 class StylesView(ResourceView):
     """Styles Viewlet."""
-
-    def index(self):
-        rendered = self._rendered_cache
-        if not rendered:
-            rendered = self.renderer["css"].render()
-            self._rendered_cache = rendered
-        return rendered
+    resource_type = "css"
 
 
 def update_resource_registry_mtime():
--- a/Products/CMFPlone/tests/testResourceRegistries.py
+++ b/Products/CMFPlone/tests/testResourceRegistries.py
@@ -10,6 +10,7 @@ from plone.registry.interfaces import IRegistry
 from plone.testing.zope import Browser
 from Products.CMFPlone.resources import add_bundle_on_request
 from Products.CMFPlone.resources import remove_bundle_on_request
+from Products.CMFPlone.resources.browser.resource import _RESOURCE_REGISTRY_MTIME
 from Products.CMFPlone.resources.browser.resource import ScriptsView
 from Products.CMFPlone.resources.browser.resource import StylesView
 from Products.CMFPlone.tests import PloneTestCase
@@ -396,7 +397,6 @@ class TestExpressions(PloneTestCase.PloneTestCase):
         self.assertIn("http://test4.foo/test.css", results)
         self.assertIn("http://test3.foo/test.css", results)
 
-
     def test_scripts_authenticated(self):
         scripts = ScriptsView(self.layer["portal"], self.layer["request"], None)
         scripts.update()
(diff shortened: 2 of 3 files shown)


petschki:

tlotze:
@petschki Thanks for looking into this so quickly!

Your change looks good to me as far as it concerns fixing the case that the expression depends on context identity, which is probably a very common case and is what the expression that gave rise to the ticket is about.

 I don't think, though, that it addresses the issue in the more fundamental way I tried to describe in my ticket comment. The expression could do more complex things than just look at basically immutable properties of the context (like portal type). Suppose, as an example, that the expression uses the context's workflow state which could change between requests; the new cache key wouldn't account for that.

What I'm trying to express is that the cache key shouldn't refer to parameters that might go into the expression but rather to the result of evaluating the expression. I'm not really sure whether evaluating the expression at this point might defeat caching to any notable extent, so working on this with someone who knows these parts of the code better would probably be a good idea. But to be correct, the code should indeed build the cache key from exactly those bundles that will be added to or removed from the theme and request after evaluating the expression, not just those that are statically enabled or disabled.

petschki:
> so working on this with someone who knows these parts of the code better would probably be a good idea.

👉🏼 @jensens

jensens:
Yay.
I will have a look, but it will take a while.

1letter (commented):
I have test this local. It looks good for me.If the condition is changed, a restart of the instance is required. The notes from @tlotze are a good points.
Side note: If the condition is changed, a restart of the instance is still required. The cache dont invalidate after the save action.

petschki:
Yes, there is and was no invalidation at all. We could take the `registry modification time` into the cachekey https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/resources/browser/resource.py#L323 ... and maybe the modification time of the context too, so that changes during the request would be updated too.

tlotze:
But this is still a lot of guesswork. The expression can depend on basically anything, all it takes is a custom method on context that looks up the moon phase or whatever. I suggest I'll try to come up with a PR that uses the expression result and we discuss the performance implications of evaluating the expression at this point with @jensens.

petschki:
I'm sorry, but I don't understand completely ... the expression result is either `True` or `False` depending on its logic.

I wonder how this was solved before the RR rewrite. Maybe there's some piece of code which we can reimplement here?

tlotze:
Of course a single expression will tell whether a particular bundle should be added. What we need is the set of bundles whose expression evaluates to `True`. (And we need to make sure not to mix bundles to be added with those to be removed as the set union in line 61 currently does.)

petschki:
Note: registry modification time is already in cachekey generator. https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/resources/browser/resource.py#L66 ... this needs obviously be tested if it works correctly.

tlotze:
Registry modification time looks strange as it is in the cache key only under a condition. Maybe that's not a problem, though. But in fact, I also consider including registry (or expression) modification time to be guesswork because neither registry nor context nor any other variable that goes in the evaluation context needs to be modified at all for the expression to evaluate in an unpredictable way for any given request.

yurj:
If the possible values are True or False, a couple of caches must be created. If there are 2 conditions, 4 caches must be saved. So just add the condition value to the cache key, it will get the cached value. The problem is that you also need 2 or 4 or whatever different bundles, depending on the result of the condition.

But I don't think the problem is cache related.

I find condition on bundles quite useless, it is application logic to load additional javascript or css on the fly. Bundle everything and live happy. Or add code in the page to load the additional css/js. Or tell webpack to use a different chunk for code with condition, and then webpack will load it when the condition is true (the result of the condition on the fly must be available to webpack).

What is the purpose to use caching when you've to calculate an expensive condition at every request? I don't get the picture, sorry.

yurj:
https://github.com/plone/Products.CMFPlone/blob/3129403f3524a31a15dae152edefb9380486f1f8/Products/CMFPlone/resources/browser/resource.py#L288C1-L290C10

```
        self.renderer["js"] = webresource.ResourceRenderer(
            resolver_js, base_url=self.portal_state.portal_url()
        )
```
this does not depend on the condition, is this the problem?

yurj:
> I'm sorry, but I don't understand completely ... the expression result is either `True` or `False` depending on its logic.
> 
> I wonder how this was solved before the RR rewrite. Maybe there's some piece of code which we can reimplement here?

Before a different bundle was returned if the condition was True. But with webpack, if you want code in different file to talk (and module federation?) you've to use different chucks to be loaded conditionally based on the condition.

petschki:
> https://github.com/plone/Products.CMFPlone/blob/3129403f3524a31a15dae152edefb9380486f1f8/Products/CMFPlone/resources/browser/resource.py#L288C1-L290C10
> 
> ```
>         self.renderer["js"] = webresource.ResourceRenderer(
>             resolver_js, base_url=self.portal_state.portal_url()
>         )
> ```
> 
> this does not depend on the condition, is this the problem?

The expression is inside the `PloneScriptResource`s which are registered to a `ResourceGroup` and then resolved with the `ResourceResolver` and rendered with the `ResourceRenderer` ... this is `webresource` logic. Not very explicit though IMHO.

yurj:
> > https://github.com/plone/Products.CMFPlone/blob/3129403f3524a31a15dae152edefb9380486f1f8/Products/CMFPlone/resources/browser/resource.py#L288C1-L290C10
> > ```
> >         self.renderer["js"] = webresource.ResourceRenderer(
> >             resolver_js, base_url=self.portal_state.portal_url()
> >         )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this does not depend on the condition, is this the problem?
> 
> The expression is inside the `PloneScriptResource`s which are registered to a `ResourceGroup` and then resolved with the `ResourceResolver` and rendered with the `ResourceRenderer` ... this is `webresource` logic. Not very explicit though IMHO.

And how you can evaluate the expression? This result is cached, so you need the result of the condition in the cache, thus having 2^n (n=number of conditions) bundles.

tlotze:
Yes, at least in theory. Maybe not all combinations actually occur, but if they do, caching would at least work correctly. But what's the problem with many combinations being cached? The only problem I see is that evaluating might be expensive. If that is really a reason not to cache with respect to the actual expression values, expressions don't work in all the generality they allow.

Based on that observation, we might as well merge this PR to solve the 80% case now and reconsider the whole thing afterwards.

petschki:
@yurj thats the problem @tlotze adresses. The cache is calculated once per context depending on registry mtime and its expression conditions at that time. There are many more expression examples which could be discussed in order to deliver a sane OOTB solution for this. If there are special cases (and I really mean moonphases c'mon ;) we should deliver a solution to hook/adapt into this cache mechanism to make it easy for integrators to customize.

tlotze:
Just for the record: The moon phases example was intentionally absurd to illustrate that you just cannot know what the expression might do.

MrTango:
I'm with @tlotze here. The right solution should eval the expression. But more important right now, would be the 80% fix, as this is biting us in many cases. So if we don't find a quick solution for the 100%, let's just merge what we have. And improve it later.
I'm even fine with, resources with expressions are being ignored by caching. Just not cache them at all.
On the other hand, the resource it self could be cached. The problem is merging them with other resources. That was avoided in the old registry. When ever you had a condition different to others the resource was delivered separate, but could be cached bu the browser.

MrTango (approved):
good for now, thx

petschki:
I'll have a look at the registry mtime though. There is a problem with the condition.

tlotze:
Given that we're going to merge the strategy of making the cache key depend on context, also adding context mtime would be easy and helpful IMO to really make this an 80% improvement.

petschki:
Updating mtime was a simple indentation error in the cachekey code. I've added a test and fixed it.
From my point of view good to merge now.

petschki:
@mauritsvanrees jenkins tests break because of the new `6019` version number https://jenkins.plone.org/job/pull-request-6.0-3.8/2908/ ... why?

petschki:

petschki:
Ah ... rebased the branch ... should be green now.

jensens:
I think this is technically fine. Just for really large sites, will RAM-usage explode?

MrTango:
@mauritsvanrees can we merge this and have it in a release soon?

Products/CMFPlone/resources/browser/resource.py
```diff
@@ -48,24 +47,24 @@ def _request_bundles(self):
             request_disabled_bundles.update(getattr(request, "disabled_bundles", []))
         return request_enabled_bundles, request_disabled_bundles
 
-    def _user_local_roles(self, site):
-        portal_membership = getToolByName(site, "portal_membership")
+    def _user_local_roles(self):
+        portal_membership = getToolByName(getSite(), "portal_membership")
         user = portal_membership.getAuthenticatedMember()
         return "|".join(user.getRolesInContext(self.context))
```
mauritsvanrees, line 53:
```suggestion
        return "|".join(sorted(set(user.getRolesInContext(self.context))))
```

We get a list here, and it might contain duplicates (I did not see this in a quick test, but I guess it might happen), and we don't want different cache keys when the order is different.

I know this line did not change in this PR, but seems good to me to include this, because we will get different cache keys per context now, instead of per site.

mauritsvanrees (changes requested):

mauritsvanrees:
I made two change requests.  But it does seem to work as intended, and fixes the problem.  So in general +1.

You do get multiple cache keys per context now, instead of per site: 2 for js and css, times the number of different combinations of roles.  The value that is cached is not too much, I saw 400 and 700 bytes locally.  Let's say that you get combinations that result in total in 10 KB per context.  If you then have a site with 10 thousand pages, and you visit them all, in all combinations, this gives a 100 MB increase in memory.  Times the number of threads and zeoclients.  So this change does have an impact on memory, but it seems reasonable to me.

I can imagine that in the future we want to be restrictive with the allowed expressions, allowing only anonymous versus authenticated, and allowing 1 (or maybe more) portal_types, and not anything else.  That would make this code simpler, and would need less cache keys.  For more advanced use cases, like indeed the phases of the moon, people can register an own viewlet apart from the resource registries, and do whatever they please.  But that is not for the near future.

yurj:
@mauritsvanrees or anyone involved

while you're on this patch, can you add
`**{"data-bundle": "diazo"},`

at line 245 and 267?

This is for back compatibility with backend,xml (https://github.com/plone/plonetheme.barceloneta/blob/030535a1e14a321cf46d2280a51338a9a8cd99b8/plonetheme/barceloneta/theme/backend.xml#L90) in Barceloneta theme. This would help in removing the main theme if we want to use Barceloneta to style the backend (content editing, control panel). 

This change just add the attribute `data-bundle="diazo"` like in the other scripts/css which already have the data-bundle attribute set to the bundle name and this can be used to eventually drop them when we want to use Barceloneta instead of the current theme.

See also https://github.com/plone/documentation/pull/1592#issuecomment-1855907291

jensens:
For very large sites, which already have some varnish or alike in front, it would be a must have to turn off caching of it at all even in prod. If it is 100MB for 10k pages, then some sites I now with >500k pages would explode in RAM usage.

petschki:
> @mauritsvanrees or anyone involved
> 
> while you're on this patch, can you add `**{"data-bundle": "diazo"},`
> 
> at line 245 and 267?
> 

@yurj I think this change would get merged faster if you open a separate PR. We're not ready here yet.

tlotze:
> For very large sites, which already have some varnish or alike in front, it would be a must have to turn off caching of it at all even in prod. If it is 100MB for 10k pages, then some sites I now with >500k pages would explode in RAM usage.

Sorry to reiterate, but RAM usage wouldn't be a function of the number of possible context objects if the cache key involved only the outcomes of evaluating conditions rather than a selection of possible inputs. It would depend exponentially on the number of conditions but even if they are many, that number only depends on code and configuration but is fixed with respect to the amount of content in the site.

jensens:
Currently the `hashtool.update(self.context.absolute_url().encode("utf8"))` add an cache item per context.

mauritsvanrees:
I think the current state is: this fixes an irritating bug, but there are doubts beause it increases memory usage, potentially a lot for large sites.

Idea: put the new code behind a switch. Enable it when an environment variable is set. Then it can more easily be tested in practice.

MrTango:
@mauritsvanrees if we just merge it, it only is a problem when you have expressions. If this really becomes a problem in huge sites, one way would be to remove the expressions in those sites.

mauritsvanrees:
> @mauritsvanrees if we just merge it, it only is a problem when you have expressions. If this really becomes a problem in huge sites, one way would be to remove the expressions in those sites.

In standard Classic UI Plone, we have the `eventedit` bundle from `plone.app.event`, which has a condition: `python: member is not None`.  So this will have an effect on all Plone sites.
The javascript from this bundle is [small](https://classic.demo.plone.org/++plone++plone.app.event/event.js), so if there are no side effects from including it for everyone, we could remove the expression in the `plone.app.event` code.

I checked a customer site, and it had the same expression for two other bundles: datagridfield-bundle and mosaic.

But still: even if there are no expressions, after merging this PR all Plone sites will use more memory and be a bit slower, right?  This is because the hash will differ per context, so the resource cache is far less effective: there will be at least two resource caches per context, instead of per site.

This could be improved by checking if there are any expressions other than `python: member is not None` (or similar).  If not, then the context does not need to be added to the hash.

Still, I think putting this behind an environment variable so we can opt in to try it out in practice, would be better.

petschki:
After discussing this in yesterdays Classic-UI Meeting and sleeping over it I'm against my current implementation of generating the cachekey because it adds way too much attributes with the same stored value to the portal.

After some investigations and timing logs I came to the conclusion, that there's nearly no difference in calculating the cachekey vs. rendering the webresources uncached. 

So I will come up with a new approach in a separate PR which removes the volatile cachekey attributes completely and only caches the rendered viewlet content on the request in order to not calculate it twice (because we have 2 viewlets with the same update() method ... could also be cleaned up in another PR to render only one `ResourcesViewlet`)

jensens:
I think you are right, calculating a complex cache key makes no sense. Lets close this one and go on with #3900

PR #3845, opened 2023-09-22
fix: avoid searching all users when many_users is flagged

mamico:
Despite the many_user flag checked, after an action for add or remove user in a group, a search is done without any filter. 
Where, for example Plone has a connected ldap with many users, this is a problem.


mamico:

jensens (approved):

mamico:

PR #3857, opened 2023-10-13
[cleanup] Remove mock dependency

jairhenrique:
Remove unnecessary dependency. 
Mock is builtin since python 3.3.


gforcada:
@jairhenrique thanks for your contribution to Products.CMFPlone! 🎉 

Seems that you have not signed the contributor agreement, would you mind doing that? On the comment above there is a link with all the information about it 🍀 

Let us know when you have done it! Otherwise, for this small patch we can accept it without the contributor agreement, but as we would love to have more contributions from you, the agreement would greatly help 😄

jairhenrique:
@gforcada I've already signed, but I don't know why it's not enabled here 🤔

polyester:
@jairhenrique sorry, I was late in processing contributor agreements, with travels from Plone conference, should be fixed in a minute.

davisagli:

davisagli:
@jairhenrique Please add a changelog entry in the `news` folder. (See https://6.docs.plone.org/contributing/index.html#change-log-entry)

jairhenrique:
@davisagli I missed some kind of extension like `removal` or `misc`, so I considered this change as `internal`, does that make sense?

news/3857.internal
```diff
@@ -0,0 +1 @@
+cleanup: remove mock dependencie from test dependencies.
```
davisagli, line 1:
```suggestion
cleanup: remove `mock` from test dependencies. @jairhenrique 
```

PR #3861, opened 2023-10-17
Add plone-61 as zcml feature, and define PLONE61MARKER as True.

mauritsvanrees:
Updated metadata version to 6100.


davisagli (approved):
LGTM if tests are green

Issue #3862, opened 2023-10-19
Problem with @@plone-addsite and INonInstallable without getNonInstallableProfiles method.


thet:
When adding a Plone site with `@@plone-addsite` and a custom add-on with a `Products.CMFPlone.interface.INonInstallable` utility which has a `getNonInstallableProfiles` method but not a `getNonInstallableProfiles`, you'll get an error like this:

```

2023-10-19 15:40:27,133 ERROR   [Zope.SiteErrorLog:252][waitress-2] 1697722827.132640.061328476540823074 http://localhost:6060/@@plone-addsite
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 176, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 385, in publish_module
  Module ZPublisher.WSGIPublisher, line 280, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 63, in call_object
  Module Products.CMFPlone.browser.admin, line 278, in __call__
  Module Products.Five.browser.pagetemplatefile, line 126, in __call__
  Module Products.Five.browser.pagetemplatefile, line 58, in __call__
  Module zope.pagetemplate.pagetemplate, line 133, in pt_render
  Module Products.PageTemplates.engine, line 378, in __call__
  Module z3c.pt.pagetemplate, line 176, in render
  Module chameleon.zpt.template, line 302, in render
  Module chameleon.template, line 215, in render
  Module chameleon.utils, line 53, in raise_with_traceback
  Module chameleon.template, line 192, in render
  Module c10ba7607eb6c62b516ecb500b8e3d0b, line 372, in render
  Module zope.tales.expressions, line 250, in __call__
  Module Products.PageTemplates.Expressions, line 225, in _eval
  Module Products.PageTemplates.Expressions, line 155, in render
  Module Products.CMFPlone.browser.admin, line 153, in profiles
AttributeError: 'HideEuphorieProducts' object has no attribute 'getNonInstallableProfiles'

 - Expression: "view/profiles"
 - Filename:   ... egg/Products/CMFPlone/browser/templates/plone-addsite.pt
 - Location:   (line 37: col 29)
 - Source:     tal:define="profiles view/profiles;
                                    ^^^^^^^^^^^^^
 - Arguments:  template: <Products.Five.browser.pagetemplatefile.ViewPageTemplateFile object at 0x7f32bede26d0>
               options: {}
[...]

PR #3863, opened 2023-10-19
Fix problem when adding a Plone site with a custom INonInstallable ut…

thet:
…ility without a getNonInstallableProfiles method.

Fixes: #3862.

This PR targets CMFPlone 5.2.x

### DIFF
--- a/Products/CMFPlone/browser/admin.py
+++ b/Products/CMFPlone/browser/admin.py
@@ -150,7 +150,11 @@ class AddPloneSite(BrowserView):
         ]
         utils = getAllUtilitiesRegisteredFor(INonInstallable)
         for util in utils:
-            not_installable.extend(util.getNonInstallableProfiles())
+            not_installable.extend(
+                util.getNonInstallableProfiles()
+                if hasattr(util, "getNonInstallableProfiles")
+                else []
+            )
 
         for info in profile_registry.listProfileInfo():
             if info.get('type') == EXTENSION and \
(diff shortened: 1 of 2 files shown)


thet:

gforcada (approved):
LGTM! 👍🏾

davisagli (commented):
I have a doubt about this. Technically if the utility doesn't have a getNonInstallableProfiles method, then it didn't properly implement the INonInstallable interface. If I am trying to add an INonInstallable utility and I make a typo in the method name, this change means that it will silently fail instead of giving a helpful error so that I notice the typo.

mauritsvanrees:
> I have a doubt about this. Technically if the utility doesn't have a getNonInstallableProfiles method, then it didn't properly implement the INonInstallable interface. If I am trying to add an INonInstallable utility and I make a typo in the method name, this change means that it will silently fail instead of giving a helpful error so that I notice the typo.

That is true.  But in the add-ons control panel we already [do the same](https://github.com/plone/Products.CMFPlone/blob/6a8eb557dc8007668dc4a837d41987ab8daf2896/Products/CMFPlone/controlpanel/browser/quickinstaller.py#L83-L89): we check if `getNonInstallableProfiles` is there.
Also, if you do make a typo in the method name, it is usually not so bad: you will get more profiles or products in a list, which probably confuses users, but the package should still work.

It is not clear in the interface, but I intended the two methods to be optional: it is fine to only give a list of profiles or only give a list of products, without needing to provide a dummy method for the other, or inheriting from a base class with dummy methods.

So: LGTM.

Issue #3871, opened 2023-11-16
Explicit acquisition is enabled on Plone 6.


gotcha:
### What I did:
- create new plone site
- add a folder (folder-one) at the root
- add a page (page-one) at the root
- access acquired URL /Plone/folder-one/page-one

### What I expect to happen:

I should see the rendering of page-one even if it is not in folder-one.

### What actually happened:

I got a 404

### What version of Plone/ Addons I am using:

Tested with 6.0.8

(Does work in 6.0.6)

### What I already debugged:
This is due to https://github.com/plone/Products.CMFPlone/pull/3781 being broken:
`Products.CMFCore` is imported by `OFS.Application.import_products` before `Products.CMFPlone`.
IOW, it is too late to change `os.environ`.

jensens:
Closing as resolved. Fixed: __init__.py now sets SKIP_PTA directly on the module-level variable, bypassing the import-order issue.
Thank you for the original report!

PR #3872, opened 2023-11-16
Actually disable explicit acquisition in Plone 6

gotcha:
Fix for https://github.com/plone/Products.CMFPlone/issues/3871

This still needs tests in order to avoid regression again.


Products/CMFPlone/__init__.py
```diff
@@ -10,8 +10,10 @@
 __version__ = pkg_resources.require("Products.CMFPlone")[0].version
 
 if __version__ < '7':
-    from Products.CMFCore.explicitacquisition import PTA_ENV_KEY
+    from Products.CMFCore import explicitacquisition
+    PTA_ENV_KEY = explicitacquisition.PTA_ENV_KEY
     os.environ[PTA_ENV_KEY] = os.environ.get(PTA_ENV_KEY, 'false')
```
davisagli, line 15: I'd say we can remove this line; it's less confusing if we don't mutate the environment.

```suggestion
```

davisagli (commented):
@gotcha Thanks for this fix. Would you be willing to add the test?

Products/CMFPlone/__init__.py
mauritsvanrees: I have rebased and force-pushed the branch, and copied over the code that is already in use on master.  It has some comments that explain a bit more what we are actually doing here.

davisagli:
Please do add a test. I was working with this setting recently to enable the check for a project, and I found the naming is quite confusing. The test can probably be something like this

```
    def setUp(self):
        self.app = self.layer["app"]
        self.portal = self.layer["portal"]
        self.portal_url = self.portal.absolute_url()
        setRoles(self.portal, TEST_USER_ID, ["Manager"])

        api.content.create(
            container=self.portal,
            type="Document",
            id="page",
        )

        transaction.commit()
        self.api_session = RelativeSession(self.portal_url)
        self.api_session.auth = (SITE_OWNER_NAME, SITE_OWNER_PASSWORD)

    def test_acquired_object_returns_200(self):
        response = self.api_session.get("/page/page")
        self.assertEqual(response.status_code, 200)
```

PR #3881, opened 2023-11-29
handle parenthesis inside quotes

### LINKED ISSUE #3879: Catalog search does not handle correctly parenthesis inside quotes
### What I did:

- Add a page with parenthesis enclosed text as a title
- Try to search that page using quoted query

### What I expect to happen:

- The search searches the page

### What actually happened:

- The search does not find the page
- When accessing through REST API (because plone.app.querystring uses the `munge_search_term` function here), we get an error because the query is transformed to an unknown string for ZCatalog

- Plone 6

erral:
Fixes #3879 for master


erral:

mauritsvanrees (approved):

PR #3883, opened 2023-11-30
1letter fix#3873

1letter:
Update wording


stevepiercy (changes requested):
Thanks for hunting down these files and creating a pull request. I have a couple of corrections, and then I approve.

PR #3884, opened 2023-12-01
Backport Fix #3873 to 6.0.x

### LINKED ISSUE #3873: "Create Classic Plone Site" should include "UI"
When creating a site, we are presented with three buttons, one of which is `Create Classic Plone Site`. It and the description text following it should be changed to `Create Classic UI Plone Site`

plone1

1letter:
https://github.com/plone/Products.CMFPlone/pull/3883


stevepiercy (approved):
Thank you!

PR #3887, opened 2023-12-11
mark get_production_resource_directory deprectated

jensens:
the function is used nowhere in Plone core.
It is a leftover from before `webresource` was introduced.


jensens:

mauritsvanrees (approved):
Thanks.

Could be back ported to 6.0.x as well, but it is no big deal.

PR #3900, opened 2024-02-15
Fix resource expression calculation by removing volatile cache

petschki:
continued from here https://github.com/plone/Products.CMFPlone/pull/3844#issuecomment-1945928927

### DIFF
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -1,19 +1,15 @@
 from ..webresource import PloneScriptResource
 from ..webresource import PloneStyleResource
-from App.config import getConfiguration
 from plone.app.layout.viewlets.common import ViewletBase
 from plone.app.theming.interfaces import IThemeSettings
 from plone.app.theming.utils import theming_policy
 from plone.base.interfaces import IBundleRegistry
 from plone.registry.interfaces import IRegistry
-from Products.CMFCore.utils import getToolByName
 from time import time
 from zope.component import getMultiAdapter
 from zope.component import getUtility
 from zope.component import queryUtility
-from zope.component.hooks import getSite
 
-import hashlib
 import logging
 import webresource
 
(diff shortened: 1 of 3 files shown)


petschki:

petschki:

mauritsvanrees (approved):
I agree with this approach.  While you were working on this, I did some timings as well:

* With the current (6.0.9) way, rendering both resource viewlets takes about 0.002 seconds total.  Expressions do not work.
* With the previous PR it is a bit more due to a [not effective request cache key](https://github.com/plone/Products.CMFPlone/pull/3844#discussion_r1428207568), but with a fix for that it is about the same.  Expressions work, but memory usage increases.
* With the current PR it becomes about 0.005, so it takes longer.  Expressions work.  Memory usage decreases (very) slightly because the volatile cache is gone.  The code is easier.

All in all, I prefer this PR.  Thanks!
Let's wait a bit till others have reviewed it as well.

Side note: most likely, on a production site, the resource viewlets are calculated once on a context, and then the html of the page is cached in Varnish or similar, at least for anonymous visitors.  So it does not matter too much if calculating these viewlets takes a bit longer.

yurj (approved):
LGTM

This really does what is expected, so +1

Optimizations:
Render time could be improved just caching the single webresource.render() method, using the registry modification time (_RESOURCE_REGISTRY_MTIME), if there's no condition.
99% of the resources has no condition, so we can cache the render() method. This method take most of the time on creating the path. The path does not depends on anything, it is always the same.

Maybe we've to check if the RAM cache here:
https://github.com/plone/Products.CMFPlone/blob/875b899de4671c1870bd4c1fe37bf057952396f1/Products/CMFPlone/resources/webresource.py#L51
for cooked_expressions but I think they are not used at all.

jensens (approved):

jensens:
@petschki Please do the same for 6.1

petschki:
will do 😉

PR #3902, opened 2024-02-17
Fix robot test "When page is linked show warning"

wesleybl:
I'm noticing that the `When page is linked show warning` test is failing sometimes. This test consists of creating content 1, selecting it as a link in content 2, trying to delete content 1, showing the message that it is related to content 2.

When the error occurs, it does not show the message that content 1 is related to content 2. It is as if content 1 was not related to any content. I suspect that in the test, another content was selected as a link, instead of content 1. This is because when the test tries to click on content 1 to select as a link, the list of objects is being loaded and moving on the screen. Then we wait some time for the content to remain still on the screen before clicking.


wesleybl:

jensens (approved):

PR #3904, opened 2024-02-19
Fix robot test "When page is linked show warning"

wesleybl:
Backport of #3902 to 6.0

### DIFF
--- a/Products/CMFPlone/tests/robot/test_linkintegrity.robot
+++ b/Products/CMFPlone/tests/robot/test_linkintegrity.robot
@@ -77,15 +77,14 @@ a link in rich text
   Click Button  css=button[aria-label="Insert/edit link"]
 
   Given patterns are loaded
-  # Somehow this does not work:
-  # Wait For Then Click Element  css=.pat-relateditems .select2-input.select2-default
   Wait until element is visible  css=.pat-relateditems .select2-input.select2-default
   Click Element  css=.pat-relateditems .select2-input.select2-default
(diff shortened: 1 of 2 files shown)


wesleybl:
Is Plone 6.0 not tested with Python 3.12?

jensens:
> Is Plone 6.0 not tested with Python 3.12?

No, this is a 6.1 feature and tested only on the master branch, your PR is against 6.0.x branch.

wesleybl:
I found it strange because [support for Python 3.12 is declared](https://github.com/plone/Products.CMFPlone/blob/ad24f0f413a496aa5a769365d6e9ab0a77916069/setup.py#L28) in Plone 6.0, but it is not tested in that version. Maybe the tests should be in Python 3.12 instead of 3.11.

stevepiercy:
@wesleybl see https://plone.org/download/release-schedule:

> ## Supported versions
> ### Plone 6.0
> ...
> - Supports Python 3.8, 3.9, 3.10, 3.11. Plone 6.0.9 should work fine on 3.12, official support is expected to land in 6.0.10.

mauritsvanrees:
PRs for Plone 6.0 are tested on Python 3.8 and 3.11, just like the 6.0 coredev branch when something is committed there.

There are jobs for 3.9, 3.10 and 3.12, but these are only run once a week.  This avoids running lots of jobs where there is very little chance of something going wrong.  See the [Core schedules tab](https://jenkins.plone.org/view/Core%20schedules/) on Jenkins.

We might start testing on 3.12 instead of 3.11 by default at some point.

Issue #3911, opened 2024-02-29
viewlets.xml references no longer existing viewlets


mauritsvanrees:
The [`viewlets.xml`](https://github.com/plone/Products.CMFPlone/blob/97f00be8bdef1726c3e335d0fa189accfec6748e/Products/CMFPlone/profiles/default/viewlets.xml) in CMFPlone (here from 6.0.x), references several viewlets that no longer exist or have been renamed.  If a non-existing viewlet name is added in an `order` or `hidden` manager, nothing seems to go wrong.  But it can be confusing.

I compared the viewlets in 5.2 and 6.0 and came up with the following differences.  I could add this in an upgrade guide.

# Plone viewlets from 5.2 to 6.0

- Plone 6.0 removes the `plone.documentactions (IDocumentActions)`  viewlet manager. In Plone 5.2 it was already empty.
- Plone 6.0 adds the `plone.belowcontentdescription (IBelowContentDescription)` viewlet manager. By default this has no viewlets.
- Plone 6.0 removes `plone.header` viewlet from `plone.portaltop` manager, making it empty.
- Plone 6.0 renames `plone.abovecontenttitle.documentactions` viewlet to `plone.documentactions` and moves it from manager `plone.belowcontentbody` to `plone.belowcontent`.
- Plone 6.0 renames `plone.abovecontenttitle.socialtags` viewlet to `plone.socialtags`. It remains in manager `plone.abovecontenttitle`.
- Plone 6.0 renames `plone.belowcontentbody.relateditems` viewlet to `plone.relateditems`. It remains in manager `plone.belowcontentbody`.
- Plone 6.0 removes the `plone.manage_portlets_fallback` viewlet from the `plone.belowcontent` manager.
- Plone 6.0 renames `plone.belowcontenttitle.documentbyline` viewlet to `plone.documentbyline` . It remains in manager `plone.belowcontenttitle`.
- Plone 6.0 renames `plone.belowcontenttitle.keywords` viewlet to `plone.keywords` and moves it from manager `plone.belowcontent` to `plone.belowcontentbody`.
- Plone 6.0 adds `plone.rights` viewlet in manager `plone.belowcontentbody`.
- Plone 6.0 moves the `plone.footer` viewlet from `plone.app.layout/viewlets` to `plone.app.portlets`. The viewlet remains in manager `plone.portalfooter`. It renders the portlets from the `plone.footerportlets` portlet manager.

Perhaps a table works better:

[...]

PR #3913, opened 2024-02-29
Cleanup viewlets.xml to not mention viewlets that no longer exist.

mauritsvanrees:
Nothing really goes wrong with non-existing viewlets, but it is confusing. Fixes https://github.com/plone/Products.CMFPlone/issues/3911

Forward port of the already merged PR #3912.


mauritsvanrees:

Issue #3916, opened 2024-03-05
getIcon indexer returns True for a folderish item with any children with id "image"


frapell:
The issue is happening here https://github.com/plone/Products.CMFPlone/blob/72ec2cad51e3096b043f79b1777f6e452f83afdf/Products/CMFPlone/CatalogTool.py#L226-L237

In order to see the problem:
1. Create a folder with any name (e.g. Test Folder), this results in /Plone/test-folder
2. Inside this new folder, add another item, can be a page or another folder, and name it specifically "Image" so its id ends up being "image". this results in /Plone/test-folder/image

At this point, if you go check the indexed folder from step 1 (/Plone/test-folder) you will notice that the `getIcon` metadata is True.

This causes issues, such as https://github.com/plone/plone.namedfile/issues/156

@davisagli provided a bit more context in this other issue, however I believe this must be fixed here, and have `getIcon` to properly return True only when the indexed item does have an image.

frapell:
I have included a test that triggers the bug. I am not sure how to properly fix this... I don't think we want to import behaviors such as https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/behaviors/leadimage.py in here?

davisagli:
@frapell How about, get the image attribute like it does currently, but then add a check that the value is an instance of NamedImage?

frapell:
@davisagli Yeah, it is either that or maybe adding `image` to the list of reserved words... what about checking if it provides `plone.namedfile.interfaces.IImage` ?

davisagli:
@frapell the IImage interface check should work

yurj:
So to implement the "icon" for a content, a NamedImage or an object implementing the `plone.namedfile.interfaces.IImage` interface should be returned when calling 'image' (attribute or function) on it?

frapell:
@yurj I am not sure why this indexer is called `getIcon` while it should really be called `hasImage` or something like that... it will only return True or False if the item has an image. You can see the referenced PR in the comment (https://github.com/plone/Products.CMFPlone/issues/1226) it is almost 10 years old, so you need to dig some deeper to find the origin of it, maybe at some point it was used for icons.
Content icons these days are managed through css AFAIK...

PR #3917, opened 2024-03-05
Bugfix for getIcon indexer and conflicting "image" id

frapell:
This PR fixes https://github.com/plone/Products.CMFPlone/issues/3916

### DIFF
--- a/Products/CMFPlone/tests/testCatalogTool.py
+++ b/Products/CMFPlone/tests/testCatalogTool.py
@@ -10,6 +10,7 @@ from plone.app.textfield import RichTextValue
 from plone.indexer.wrapper import IndexableObjectWrapper
 from plone.uuid.interfaces import IAttributeUUID
 from plone.uuid.interfaces import IUUID
+from plone.namedfile.file import NamedImage
 from Products.CMFCore.indexing import processQueue
 from Products.CMFCore.permissions import AccessInactivePortalContent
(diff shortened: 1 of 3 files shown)


frapell:

Products/CMFPlone/tests/testCatalogTool.py
```diff
@@ -1414,6 +1415,21 @@ def test_uuid(self):
         self.assertTrue(wrapped.UID)
         self.assertTrue(uuid == wrapped.UID)
 
+    def test_getIcon(self):
+        from Products.CMFPlone.CatalogTool import getIcon
```
davisagli, line 1419: getIcon here vs get_icon below?

davisagli (approved):
Looks okay to me, except for one small question.

Could you open a PR to the master branch too, so the change doesn't get stuck in 6.0 only?

Products/CMFPlone/tests/testCatalogTool.py
frapell: yeah, getIcon is a `plone.indexer.delegate.DelegatingIndexerFactory` and by doing `getIcon.callable` I get the actual function to test.
This is the same way it is done above, with `getObjSize`

davisagli: @frapell yep sorry, I overlooked that line

PR #3920, opened 2024-03-05
move condition to parent element

### LINKED ISSUE #3919: Empty <a> tag in @@search results page for results that do NOT have an icon/image.
Search results (@@search) that do NOT contain an icon/image have an empty <a> tag

image

There is a condition on the  tag inside that <a> `tal:condition="python: show_images and hasIcon"`. We should just move this condition up to the parent <a> instead. This probably only matters if using a screen reader.

Plone 6.0.10

ewohnlich:
This fixes #3919 where the search results page would sometimes have an 'empty' anchor tag.

### DIFF
--- a/news/3919.bugfix
+++ b/news/3919.bugfix
@@ -1,1 +1,1 @@
-Fixes case where empty links would be present in the search results for content that does not have an icon.
+Fixes case where empty links would be present in the search results for content that does not have an icon. @ewohnlich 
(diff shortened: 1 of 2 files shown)


news/3919.bugfix
```diff
@@ -0,0 +1 @@
+Fixes case where empty links would be present in the search results for content that does not have an icon.
```
davisagli, line 1:
```suggestion
Fixes case where empty links would be present in the search results for content that does not have an icon. @ewohnlich 
```

davisagli:

PR #3929, opened 2024-03-21
Fix test Scenario: Select All items

wesleybl:
Fix intermittent failure in test robot. Sometimes waiting for the element with the `itemRow` class fails despite it being on the screen. So we wait 1 second instead of waiting for the element.

Fix: https://jenkins.plone.org/job/pull-request-6.1-3.10/2/robot/report/robot_log.html#s1-s9-t1

### DIFF
--- a/Products/CMFPlone/tests/robot/test_folder_contents.robot
+++ b/Products/CMFPlone/tests/robot/test_folder_contents.robot
@@ -60,7 +60,7 @@ I click the '${link_name}' link
 
 I select all the elements
     Wait until page contains element  css=.pat-structure .select-all
-    Wait until page contains element  css=.itemRow
+    Sleep  1s
     ${select_all_selector}  Set Variable  .pat-structure .select-all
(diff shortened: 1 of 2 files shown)


wesleybl:

davisagli (approved):
It always makes me a little sad when we have to add a sleep. But I also accept we have to be pragmatic sometimes.

petschki:
Just a note: @mauritsvanrees added the keywords `Wait for Element` and `Wait for then click element` in https://github.com/plone/plone.app.robotframework/blob/master/src/plone/app/robotframework/selenium.robot#L99 which fixed several flaky tests in CMFPlone robot tests ... there'S also a`Wait for Elements` which does the same for multiple elements.

wesleybl:
@petschki the error here was not a lack of waiting. The wait command did not detect the element, even with the element on the screen. So I think the commands you indicated would also fail.

Anyway, thanks for pointing out the commands!

PR #3941, opened 2024-04-23
feat: extends sendmail wrapper

mamico:
This change enables the use of new `MailHost` features, e.g. `implicit_tls`.

https://github.com/zopefoundation/Products.MailHost/pull/47
https://github.com/zopefoundation/zope.sendmail/pull/54

### DIFF
--- a/Products/CMFPlone/patches/sendmail.py
+++ b/Products/CMFPlone/patches/sendmail.py
@@ -38,24 +38,16 @@ def applyPatches():
         SMTPMailer.send = catchAllExceptions(SMTPMailer.send)
 
 
-def new_init(
-    self,
-    hostname="localhost",
-    port=25,
-    username=None,
-    password=None,
-    no_tls=False,
-    force_tls=False,
-):
-    registry = getUtility(IRegistry)
-    mail_settings = registry.forInterface(IMailSchema, prefix="plone")
-    self.hostname = mail_settings.smtp_host
-    self.port = mail_settings.smtp_port
-    self.username = mail_settings.smtp_userid
-    self.password = mail_settings.smtp_pass
-    self.force_tls = force_tls
-    self.no_tls = no_tls
-    self._smtp = _SMTPState()
+def mail_settings_wrapper(func):
+    def wrapper(self, **kwargs):
+        registry = getUtility(IRegistry)
+        mail_settings = registry.forInterface(IMailSchema, prefix="plone")
+        kwargs["hostname"] = mail_settings.smtp_host
+        kwargs["port"] = mail_settings.smtp_port
+        kwargs["username"] = mail_settings.smtp_userid
+        kwargs["password"] = mail_settings.smtp_pass
+        return func(self, **kwargs)
 
+    return wrapper
 
-SMTPMailer.__init__ = new_init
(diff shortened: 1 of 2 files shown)


mamico:
more info here: https://community.plone.org/t/configuring-plone-email-smtp-when-ssl-is-required/17075/7

manually tested against different smtp server configuration with

docker-compose.yml
```
version: '3'

services:
  smtp4dev:
    image: rnwood/smtp4dev:v3
    ports:
      - '1080:80'
      - '1025:25'
      - '1143:143'
    volumes:
      - smtp4dev-data:/smtp4dev
    environment:
      # Specifies the TLS mode to use. None=Off. StartTls=On demand if client supports STARTTLS.
      # ImplicitTls=TLS as soon as connection is established.
      - ServerOptions__TlsMode=None
      # - ServerOptions__TlsMode=StartTls
      # - ServerOptions__TlsMode=ImplicitTls
volumes:
  smtp4dev-data:
```

zope.sendmail >= 6.2
Products.MailHost branch https://github.com/zopefoundation/Products.MailHost/tree/implicit_tls

mamico:

mauritsvanrees (approved):
LGTM.

One failure on one of the two Jenkins jobs, but it is unrelated.
This would need a backport to 6.0.x.

Just checking if I understand the code correctly: this is *not* dependent on your open PR in `Products.MailHost`, right?  It can be used with the latest release, and then once your PR is merged, the wrapper will automatically support the new `implicit_tls` option.

Just some further thoughts, which need no action in this PR:

* It would be nice to have all `Products.MailHost` options that are in the ZMI also available in the Plone control panel (Classic and Volto).  But the current control panel is already easily misconfigured if you don't know what you are doing.
* Alternatively, I wonder if on submit of the control panel we could automatically set the various TLS options in the ZMI depending on the port number.
* There is another patch in this file which has a comment "BBB remove when zope.sendmail 3.8.0 is released."  So that patch can likely safely be removed.

mamico:
> Just checking if I understand the code correctly: this is *not* dependent on your open PR in `Products.MailHost`, right?  It can be used with the latest release, and then once your PR is merged, the wrapper will automatically support the new `implicit_tls` option.

Correct.

> * It would be nice to have all `Products.MailHost` options that are in the ZMI also available in the Plone control panel (Classic and Volto).  But the current control panel is already easily misconfigured if you don't know what you are doing.
> * Alternatively, I wonder if on submit of the control panel we could automatically set the various TLS options in the ZMI depending on the port number.

I personally don't like the current behavior of having the same configuration in the mailhost and plone registry, however that's how it is. In any case we can implement your second option and let people tweaking the mailhost tool for non-standard use cases. 
And anyway, as you said, not in this pr

mauritsvanrees:
Alternative idea: in `Products/MailHost/MailHost.py` patch [`_makeMailer`](https://github.com/zopefoundation/Products.MailHost/blob/41cd753391f22e9a03910effc2770c2a24b7f46f/src/Products/MailHost/MailHost.py#L235-L242).  Do more or less the same patch as we do now, but only do it if `self.id == 'MailHost'`.  Then the patch should not affect alternative MailHost instances that you setup.

PR #3946, opened 2024-05-02
use context instead of here and prefix main_template with @@

jensens:
`here` is deprecated or at least discouraged to use AFAIK. Using `@@` speeds up lookup in the publisher.


jensens:

PR #3965, opened 2024-05-30
do not test for a property that is not part of the underlying vocabulary

jensens:
see https://github.com/plone/plone.base/pull/65


jensens:

PR #3976, opened 2024-06-15
Use details element for collapsibles in the resource registry

### LINKED ISSUE #3942: Broken or missing JavaScript leads to unusable resource registry.

The resource registry has all it's bundles collpapsed in an bootstrap accordion element.

When JavaScript cannot be loaded (e.g. by providing it via webpack dev server or if it is broken), then the resource registry has all it's resource tabs are closed and cannot be opened and thus are not interactable.

See:

grafik

### What I did:

- Remove the plone bundle by deleting the entry or disabling it.

### What I expect to happen:

In http://localhost:8080/Plone/@@resourceregistry-controlpanel I'd expect all the resource bundle entries to be un-collapsed and the form fields being shown.

### What actually happened:

All the resource entries are collapsed and the resource registry is not interactable.

    Plone 6.1.0a3.dev0 (6102)
    CMF 3.3
    Zope 5.9
    Python 3.12.2 (main, Mar 17 2024, 12:06:58) [GCC 13.2.0]
    PIL 10.2.0 (Pillow)
    WSGI: On
    Server: waitress 2.1.2

## Proposal to fix this

The HTML element `details` is widely available and provides content collapsing without JavaScript being necessary.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

reinhardt:
Makes it possible to toggle elements even with broken or missing javascript.

However, this changes the look & feel. The CSS is not applied any more because some classes were removed so that the accordion JS does not interfere. Maybe there's a better way to do this, or maybe we just need some additional CSS?

2024-06-15-12-58-42_2552x1368

A note on accessibility: A details/summary combination is already accessible and aparently does not need any more aria attributes. However, the `<summary>` should be descriptive enough, we could argue if it is in our case. But I believe, thati's fine already. More info see: http://kb.daisy.org/publishing/docs/html/details.html

Fixes #3942

PRs:
https://github.com/plone/Products.CMFPlone/pull/3976
https://github.com/plone/plonetheme.barceloneta/pull/374

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/resourceregistry.pt
+++ b/Products/CMFPlone/controlpanel/browser/resourceregistry.pt
@@ -102,39 +102,27 @@
              "
         >
           <tal:block tal:repeat="bundle python:view.bundles_data">
-            <div class="accordion-item ${python:'active' if bundle['name'] == request.form.get('name', None) else ''}">
-              <h2 class="accordion-header"
-                  id="heading-${python:bundle['safe_name']}"
-              >
-                <button class="accordion-button"
-                        aria-controls="collapse-${python:bundle['safe_name']}"
-                        aria-expanded="false"
-                        type="button"
-                        data-bs-target="#collapse-${python:bundle['safe_name']}"
-                        data-bs-toggle="collapse"
-                >
-                  <span class="me-2"
-                        tal:condition="python:bundle['name'] and bundle['enabled']"
+            <details class="accordion-item ${python:'active' if bundle['name'] == request.form.get('name', None) else ''}">
+              <summary class="accordion-button">
+                <span class="me-2"
+                      tal:condition="python:bundle['name'] and bundle['enabled']"
+                      i18n:translate=""
+                ><tal:icon tal:replace="structure python:icons.tag('file-check', tag_alt='Enabled bundle')" /></span>
+                <span class="me-2"
+                      tal:condition="python:bundle['name'] and not bundle['enabled']"
+                      i18n:translate=""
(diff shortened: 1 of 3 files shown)


rohnsha0:

mauritsvanrees:
+1 from me.  A few tests are failing though.

thet:
I've added some classes and styles. It now looks almost like before:

image

PR: https://github.com/plone/plonetheme.barceloneta/pull/374

thet:
Now with label wrapping their inputs and no id's being duplicated (tnx @ale-rt  for pointing this out).

Together with the changes in plonetheme.barceloneta, this looks like that:

image

reinhardt:
Awesome, thanks @thet! LGTM (Though I can't give an approving review because I created the PR :stuck_out_tongue:)

petschki (changes requested):
Input fields aren't 100% width because of nested label/input markup. BS 5 default isn't nested ... see screenshot: first without nesting, second input inside label.

Bildschirmfoto 2024-06-18 um 08 03 33

thet:
@petschki did you also check out the plonetheme.barceloneta branch from https://github.com/plone/plonetheme.barceloneta/pull/374 ?

For me, it looks like this:

image

MrTango (changes requested):
looks good, but i would keep the h2 instead of the strong tags, as it is semantically better.

thet:

MrTango (approved):

petschki (approved):
@thet sorry ... my already checked out barceloneta was on the wrong branch ... looks good now here.

Issue #3975, opened 2024-06-15
Without plone.app.upgrade, the plone-upgrade page does not work.


mauritsvanrees:
### What I did:

* Start a 6.1 site with only Products.CMFPlone.
* Use a database copy from Plone 5.2 (or 6.0).
* Go to `@@plone-upgrade`.  It shows you need to upgrade the site.
* Click to upgrade and you see an error that the site did not reach the required version.

This is logical, because the page needs `plone.app.upgrade`: the upgrade steps are define there.

### What I expect to happen:

See a message on the page, before you even try anything: "The plone.app.upgrade package is missing.  You need it for running upgrades."

### What actually happened:

I got an uninformative error, and no hints on the console.

mauritsvanrees:
PR was merged a while ago.

PR #3977, opened 2024-06-15
Plone upgrade page: show missing packages

mauritsvanrees:
Fixes issue #3975.

This does the following:

1. It shows which packages were previously installed (with GenericSetup) but are currently not available.  Example is `plone.app.discussion` which will be missing in Plone 6.1 if you only use `Products.CMFPlone`.
2. It shows an error when the site needs upgrading but there are no upgrades available.  Most likely cause is that `plone.app.upgrade` is missing, so we explicitly check this.
3. If no upgrades are there, we disable the Upgrade button.

It looks like this:

Screenshot 2024-06-15 at 14 00 21

### DIFF
--- a/Products/CMFPlone/browser/admin.py
+++ b/Products/CMFPlone/browser/admin.py
@@ -1,6 +1,8 @@
 from AccessControl import getSecurityManager
 from AccessControl.Permissions import view as View
 from collections import OrderedDict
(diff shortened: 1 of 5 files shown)


rohnsha0:
LGTM

ale-rt:

ale-rt (approved):

PR #3980, opened 2024-06-18
Fix robot tests for new `pat-contentbrowser` pattern

petschki:
Update selectors in TinyMCE `linkModal`

Test in combination with https://github.com/plone/plone.staticresources/pull/336


petschki:
closed in favor of #4008

Issue #3961, opened 2024-05-17
Refactor site creation in Products.CMFPlone to use plone.distribution


ericof:
Part of #3854

# Tasks
- [x] `Products.CMFPlone.browser.admin.AddPloneSite` only registered if `plone.distribution` is not available
- [ ] Refactor site creation to be like the function from plone.distribution (supporting a distribution)
- [x] Remove creation of example content (make sure those are still created in the test layer)

mauritsvanrees:
Without `plone.distribution` available, do we want to remove the Advanced options?  I can imagine that we want to point users to *either* distributions *or* just let them install the extra profiles via the Add-ons control panel.

<img width="851" alt="Screenshot 2024-06-19 at 19 31 45" src="https://github.com/plone/Products.CMFPlone/assets/210587/ac0160ec-4788-4156-ba50-baf2527004fe">

Actually, you can't install "Zope Root Cookie Login" (`Products.PlonePAS:root-cookie` profile) in the control panel.  You would need to do that via `portal_setup`.  But that is really an advanced corner case.   Since it works on the Zope root, you don't even want it available for install from within Plone.  So I think that is fine.  Anyway, you can still do that in a Distribution. :-)

mauritsvanrees:
Progress during Buschenschankensprint (in the train back actually)

* https://github.com/plone/Products.CMFPlone/pull/3981
* https://github.com/plone/plone.app.testing/pull/97

mauritsvanrees:
The CMFPlone PR does not "Refactor site creation to be like the function from plone.distribution (supporting a distribution)".  I don't really know what that would like like if `plone.distribution` is not available, which is the only way that the `plone-addsite` view is even visible.

Ah, or do you mean that the `factory.addPloneSite` function should be able to install a distribution if `plone.distribution` *is* available?  Then this function would accept an extra `distribution` keyword argument, and call a function in `plone.distribution`.  Something like that?

mauritsvanrees:
@ericof @davisagli Can this issue be closed?  Or do we want to do something about my previous comment about the open task "Refactor site creation to be like the function from plone.distribution (supporting a distribution)"?

mauritsvanrees:
Ah, right. In `addPloneSite` you can pass a [`distribution_name`](https://github.com/plone/Products.CMFPlone/pull/3981/files#diff-364fa892973e67343f3a2c29d75de4aa9885dfb4a2735ca5ea88ca7e95fb0b89R141).

PR #3981, opened 2024-06-19
Refactor site creation

mauritsvanrees:
This is part of https://github.com/plone/Products.CMFPlone/issues/3961

This affects the site creation process when `plone.distribution` is not available.  See the individual commits for details.  In this case, the plone-overview page on the Zope root looks like this:

Screenshot 2024-06-19 at 20 38 50

The Advanced version of the Add Plone Site form is gone.  For advanced options you should use a Plone Distribution.

This needs a change in `plone.app.testing`, otherwise loads of tests fail.  Only a one line change though.  I will link the PR.

I will keep this as Draft PR for now.


Products/CMFPlone/browser/templates/plone-overview.pt
```diff
@@ -117,6 +102,11 @@
                 these frontends and possible upgrade paths from older Plone versions
                 to Plone 6.
             </p>
+            <p i18n:translate="help_create_plone_site_buttons_3">
+                Starting with Plone 6.1, you should add the <code>plone.classicui</code>
+                package if you want to create the default content.
+                If you see this text, that means you do not have this package yet.
```
stevepiercy, line 108: Suggest the following wording.

```
Starting with Plone 6.1, default content is not loaded into the site.
If you want to load the default content, you should install the <code>plone.classicui</code> package.
If you see this text, that means you have not installed that package yet.
```

news/3961.breaking.3
```diff
@@ -0,0 +1,4 @@
+`factory.addPloneSite`: remove `setup_content` and `content_profile_id` keyword arguments.
+We never setup content anymore.  Use a Plone Distribution if you need this.
```
stevepiercy, line 2:
```suggestion
We no longer load default content. Use a Plone Distribution if you need this.
```

stevepiercy (changes requested):
Polish up the messages.

Products/CMFPlone/browser/templates/plone-overview.pt
mauritsvanrees: Suggestion applied, thanks.

stevepiercy (approved):
Thank you! Looks good so far.

stevepiercy:
Upon merge and release, we need to update the documentation. See https://github.com/plone/documentation/issues/1721

PR #3998, opened 2024-08-12
Do not use deprecated base_hasattr in utils.py. [6.0]

mauritsvanrees:
This was moved to plone.base.utils, so utils.py should not be using its own deprecated import.

In fact, this gives a strange error when running the `Products.validation` tests in the coredev 6.0 buildout:

```
$ bin/test -s Products.validation
...
Error in test test_isValidId_plone (Products.validation.tests.test_validation.TestValidation.test_isValidId_plone)
Traceback (most recent call last):
  File "/Users/maurits/.pyenv/versions/3.11.9/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/Users/maurits/.pyenv/versions/3.11.9/lib/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/Users/maurits/.pyenv/versions/3.11.9/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
  File "/Users/maurits/community/plone-coredev/6.0/src/Products.validation/Products/validation/tests/test_validation.py", line 154, in test_isValidId_plone
    self.assertEqual(v('good', obj), 1)
  File "/Users/maurits/community/plone-coredev/6.0/src/Products.validation/Products/validation/validators/IdValidator.py", line 69, in __call__
    result = check_id(instance, id, required=kwargs.get('required'))
  File "/Users/maurits/community/plone-coredev/6.0/src/Products.CMFPlone/Products/CMFPlone/utils.py", line 641, in check_id
    result = _check_for_collision(contained_by, id, **kwargs)
  File "/Users/maurits/community/plone-coredev/6.0/src/Products.CMFPlone/Products/CMFPlone/utils.py", line 684, in _check_for_collision
    if base_hasattr(contained_by, "checkIdAvailable"):
NameError: name 'base_hasattr' is not defined
```

One strange thing here is that the error only shows up when you run these packae tests in isolation.  A full `bin/test` in coredev 6.0 works.

PR #3999, opened 2024-08-12
Do not use deprecated base_hasattr in utils.py. [6.1]

mauritsvanrees:
This was moved to plone.base.utils, so utils.py should not be using its own deprecated import.

This is a forward port of PR #3998.


mauritsvanrees:

davisagli (approved):

PR #4000, opened 2024-08-14
Use Products.isurlinportal directly

mauritsvanrees:
Do not rely on it patching our URLTool.  This solves a cyclic dependency.
See https://github.com/plone/Products.isurlinportal/issues/12

Test together with https://github.com/plone/Products.isurlinportal/pull/30

### DIFF
--- a/Products/CMFPlone/URLTool.py
+++ b/Products/CMFPlone/URLTool.py
@@ -6,6 +6,7 @@ from Products.CMFCore.interfaces import ISiteRoot
 from Products.CMFCore.URLTool import URLTool as BaseTool
 from Products.CMFPlone.patches.gtbn import rewrap_in_request_container
 from Products.CMFPlone.PloneBaseTool import PloneBaseTool
+from Products.isurlinportal import isURLInPortal
 from zope.component import getUtility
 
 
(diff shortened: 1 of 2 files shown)


Products/CMFPlone/URLTool.py
```diff
@@ -14,26 +15,10 @@ class URLTool(PloneBaseTool, BaseTool):
     security = ClassSecurityInfo()
     toolicon = "skins/plone_images/link_icon.png"
 
-    @security.public
```
davisagli, line 17: Do we still need a security declaration?

mauritsvanrees:
Pull request 4000!  Cake for everyone! :-D

Products/CMFPlone/URLTool.py
mauritsvanrees: I wondered that too, and wanted to add `security.declarePublic("isURLInPortal")`, because I think `@security.public` can only work on a method.  But without any security declaration I added this Script Python in the ZMI and I could call it just fine, even as anonymous user:

```
urls = ["http://localhost:8080/Plone", "https://maurits.vanrees.org"]
for url in urls:
    ip = context.portal_url.isURLInPortal(url)
    print(f"{ip} {url}")

return printed
```

davisagli (approved):

PR #4007, opened 2024-09-04
Update help text for redirector target path

davisagli:
plone.app.redirector supports redirecting to external URLs, so we shouldn't say the target has to start with a slash.


davisagli:

PR #4016, opened 2024-09-24
Fix false positive `missing_packages` in `@@plone-upgrade`

### LINKED ISSUE #4015: `@@plone-upgrade` page claims false missing packages
There are packages which defines multiple installable profiles in subfolders. See `collective.collectionfilter` for example.

The upgrade page tries to find a distribution for a profile name but it fails for those "sub" profiles (though they are installed) and shows the following warning:

Bildschirmfoto 2024-09-24 um 08 59 25

petschki:
fixes #4015

### DIFF
--- a/Products/CMFPlone/browser/admin.py
+++ b/Products/CMFPlone/browser/admin.py
@@ -2,6 +2,7 @@ from AccessControl import getSecurityManager
 from AccessControl.Permissions import view as View
 from collections import OrderedDict
 from functools import cached_property
+from importlib import import_module
 from importlib.metadata import distribution
 from importlib.metadata import PackageNotFoundError
 from OFS.interfaces import IApplication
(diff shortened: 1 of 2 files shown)


petschki:

davisagli (approved):
First thought: this might be dangerous, since importing modules can have side effects.

Second thought: If the profile was installed, then it's very likely the module was already imported by ZCML anyway.

Third thought: this is probably more correct even without submodules, since the name of a Python distribution does not necessary match the name of the packages inside.

mauritsvanrees (commented):
This seems fine.  But how about trying both ways?

```
            try:
                distribution(package)
            except PackageNotFoundError:
                # profiles can live in submodules of packages.
                # check if we can import the module namespace
                try:
                    import_module(package)
                except ModuleNotFoundError:
                    missing.append(package)
```

This might catch some other corner cases, although I can't really say which ones. :-)

petschki:

mauritsvanrees (approved):

PR #4018, opened 2024-10-02
Fix robottests for latest contentbrowser updates

petschki:
see https://github.com/plone/plone.staticresources/pull/354

### DIFF
--- a/Products/CMFPlone/tests/robot/test_contentbrowser.robot
+++ b/Products/CMFPlone/tests/robot/test_contentbrowser.robot
@@ -197,7 +197,7 @@ I upload an image via contentbrowser
     Wait For Then Click Element  css=.linkModal .content-browser-selected-items-wrapper button.btn-primary
     Click item in column    1    3
     Click item in column    2    1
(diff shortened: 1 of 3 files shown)


1letter (approved):
LGTM, Thanks! Sidenote: If i run `bin/test -s Products.CMFPlone -t test_contentbrowser.robot --all` all is fine, but if i run `ROBOT_BROWSER=chrome bin/test -s Products.CMFPlone -t test_contentbrowser.robot --all` then some tests fails. I think it is  a viewport misconfiguration, and has nothing todo with the tests.

1letter:
> OK, on my Mac both commands run without failures. I mark this PR as draft thought because it needs merged staticresources first.

That was the reason why I recently changed the the tests with the JS Hack in the tests. But this happens only on my small Notebook at Home. On my PC at work with a larger Screen all tests run without problems. I think it's a Selenium thing. I hope that all the Problems go away with port of tests to playwright. But this takes a while. The CI is sometimes broken, depending on which node the tests are running. that's my impression.

petschki:
There was a version update recently on coredev which looks like it made the tests flaky. I've created a PR which reverts this.

https://github.com/plone/buildout.coredev/pull/948

petschki:
this is outdated ... I'll update to playwright asap

PR #4021, opened 2024-10-08
Use "non-strict" parsing for `email.utils.getaddresses`

### LINKED ISSUE #4020: `email.utils.getaddresses` was changed to "strict-parsing" in Python 3.11.10 and 3.12.6
when running the tests we've encountered the following error in `testvalidateSingleEmailAddress`

```
AssertionError: True is not false : user@example.org
Bcc: user@example.org should fail
```

this is because `email.utils.getaddresses` has changed its strategy to "strict parsing" in the newest Python 3.11 and 3.12 versions. (see https://docs.python.org/3.12/whatsnew/3.12.html#id11 and https://docs.python.org/3.11/whatsnew/3.11.html#email)

### proposed solution:

~~add the parameter `strict=False` to get the old behavior back and test with `getattr(email.utils, 'supports_strict_parsing', False)` if this parameter is available~~

new approach: respect the new output from strict parsing.

petschki:
fixes #4020

### DIFF
--- a/Products/CMFPlone/PloneTool.py
+++ b/Products/CMFPlone/PloneTool.py
@@ -8,7 +8,6 @@ from Acquisition import aq_inner
 from Acquisition import aq_parent
 from ComputedAttribute import ComputedAttribute
 from DateTime import DateTime
-from email.utils import getaddresses
 from OFS.ObjectManager import bad_id
 from OFS.SimpleItem import SimpleItem
 from plone.base.defaultpage import check_default_page_via_view
@@ -55,6 +54,7 @@ from zope.event import notify
 from zope.interface import implementer
 from zope.lifecycleevent import ObjectModifiedEvent
 
+import email.utils
 import re
 import sys
 import transaction
@@ -93,6 +93,15 @@ METADATA_DCNAME = {
     "Rights": "DC.rights",
 }
 METADATA_DC_AUTHORFIELDS = ("Creator", "Contributors", "Publisher")
+
+
+# use old non-strict parsing behavior
+# see https://docs.python.org/3.11/whatsnew/3.11.html#email
+# and https://docs.python.org/3.12/whatsnew/3.12.html#id11
+def getaddresses(fieldvalues):
+    if getattr(email.utils, "supports_strict_parsing", False):
+        return email.utils.getaddresses(fieldvalues, strict=False)
+    return email.utils.getaddresses(fieldvalues)
 
 
 @implementer(IPloneTool)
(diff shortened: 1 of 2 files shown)


petschki:

mauritsvanrees (approved):
It sounds a bit weird: the test expects that "email\nemail" fails, which it does not, and you fix it by being *less* strict.  I don't quite follow the logic, but I see locally that this fixes a test failure, also on Python 3.10.

mauritsvanrees:
Can be merged when green.

petschki:
in our case `strict=False` means "old parsing behavior" ... what I don't really understand is the new parsing behavior that when you do `getaddresses(["mail@mail.com\nmail@mail.com"])` you get `(["", ""])` back:

```
>>> email.utils.getaddresses(["mail@mail.com\nmail@mail.com"])
[('', '')]
>>> email.utils.getaddresses(["mail@mail.com\nmail@mail.com"], strict=False)
[('', 'mail@mail.com'), ('', 'mail@mail.com')]
>>>
```

yurj:
https://github.com/python/cpython/issues/102988 maybe this?

petschki:
Update: I think we should update the email validation to follow the actual strict behavior. So inputs like `email@mail.com\n email@mail.com` aren't valid anymore. But I found out, that if you separate mails only with a comma it returns the expected result:

### python 3.11.10
```
>>> email.utils.getaddresses(["user@example.org,\n user2@example.org"])
[('', 'user@example.org'), ('', 'user2@example.org')]
>>> email.utils.getaddresses(["user@example.org\n user2@example.org"])
[('', '')]
```

### python 3.11.5
```
>>> email.utils.getaddresses(["user@example.org,\n user2@example.org"])
[('', 'user@example.org'), ('', 'user2@example.org')]
>>> email.utils.getaddresses(["user@example.org\n user2@example.org"])
[('', 'user@example.org'), ('', 'user2@example.org')]
```


So in addition to the `len()` I'd check for the fallback return value `[('', '')]` here https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/PloneTool.py#L163

What do you think?

mauritsvanrees:
What I still don't get is why these tests pass fine on Jenkins with the CMFPlone master branch, but locally they fail.
Ah, if you already get a difference between Python 3.11.10 and 3.11.15, then that may be the explanation.  Locally I have all the latest bugfix releases installed, and Jenkins probably lags a bit behind.

Also, https://docs.python.org/3/library/email.utils.html says the "strict" parameter was introduced in 3.13.  But the links you pointed to in issue #4020 already mention this in the release notes of 3.11 and 3.12.  Ah, those are also notes for bug fix releases, they were not yet in the first final releases.

Meanwhile locally I see that even the latest Python 3.8 has the "strict" parameter.

So indeed the way to go is not to rely on `getaddresses` working properly, but to have our own extra checks.

mauritsvanrees:
Closing this PR in favour of #4023

PR #4019, opened 2024-10-07
Port robot tests

1letter:
This is part of https://github.com/plone/Products.CMFPlone/issues/3813

port robot tests

- [x] [test_actionmenu.robot](https://github.com/plone/Products.CMFPlone/commit/e0718d9d659b866c4402035b869e562ba4980609)
- [x] [test_contentbrowser.robot](https://github.com/plone/Products.CMFPlone/commit/24c042bfc0a93209ec8667f2a5da819fed1a6c61)
- [x] [test_controlpanel_actionmenu.robot](https://github.com/plone/Products.CMFPlone/commit/522f942b5e4c65d531cbd76f3dc9536aca18303f)
- [x] [test_controlpanel_editing.robot](https://github.com/plone/Products.CMFPlone/commit/936cc36d6be4cc49500452b3efc1907a273e5d1f)
- [x] [test_controlpanel_filter.robot](https://github.com/plone/Products.CMFPlone/commit/8a26de1d04f05dbdf3427c21312c541546919519)
- [x] [test_controlpanel_language.robot](https://github.com/plone/Products.CMFPlone/commit/12746f26ed1cd16c5a64574e4b7ce3e4e36ec3ee)
- [x] [test_controlpanel_markup.robot](https://github.com/plone/Products.CMFPlone/commit/53f5956882dcda9be412f61d78332649deb78d48)
- [x] [test_controlpanel_navigation.robot](https://github.com/plone/Products.CMFPlone/commit/f0a70d1aa636fa448c14a799a1a26493a7c99aa8)
- [x] [test_controlpanel_redirection.robot](https://github.com/plone/Products.CMFPlone/commit/73677a5bf5e98e20afcac19389ae01479ecd6326)
- [x] [test_controlpanel_search.robot](https://github.com/plone/Products.CMFPlone/commit/9826bf0b73e9b022ff029c6f5c2714473ae5fecc)
- [x] [test_controlpanel_security.robot](https://github.com/plone/Products.CMFPlone/commit/caba91ed383980a42e186c626c89f42edb8656d9)
- [x] [test_controlpanel_site.robot](https://github.com/plone/Products.CMFPlone/commit/a6b2e810145aaee377dcdafb853db35bee4a9adf)
- [x] [test_controlpanel_social.robot](https://github.com/plone/Products.CMFPlone/commit/ceaacf758498eab8361141850bf78cef8bd290c4)
[...]

PR #4023, opened 2024-10-09
Respect new strict parsing of `email.utils.getaddresses`

petschki:
preferred solution for #4020 ... successor of #4021

### DIFF
--- a/Products/CMFPlone/PloneTool.py
+++ b/Products/CMFPlone/PloneTool.py
@@ -137,13 +137,15 @@ class PloneTool(PloneBaseTool, UniqueObject, SimpleItem):
         if not isinstance(address, str):
             return False
 
-        sub = EMAIL_CUTOFF_RE.match(address)
-        if sub is not None:
-            # Address contains two newlines (possible spammer relay attack)
-            return False
+        address = address.strip()
 
         # address can be empty if getaddresses has parsing errors (returns [("", "")])
         if address == "":
+            return False
+
+        sub = EMAIL_CUTOFF_RE.match(address)
+        if sub is not None:
+            # Address contains two newlines (possible spammer relay attack)
             return False
 
         # sub is an empty string if the address is valid
--- a/Products/CMFPlone/PloneTool.py
+++ b/Products/CMFPlone/PloneTool.py
@@ -142,6 +142,10 @@ class PloneTool(PloneBaseTool, UniqueObject, SimpleItem):
             # Address contains two newlines (possible spammer relay attack)
             return False
 
+        # address can be empty if getaddresses has parsing errors (returns [("", "")])
+        if address == "":
+            return False
+
         # sub is an empty string if the address is valid
         sub = EMAIL_RE.sub("", address)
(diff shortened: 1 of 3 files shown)


petschki:

petschki:

Products/CMFPlone/PloneTool.py
```diff
@@ -142,6 +142,10 @@ def validateSingleNormalizedEmailAddress(self, address):
             # Address contains two newlines (possible spammer relay attack)
             return False
 
+        # address can be empty if getaddresses has parsing errors (returns [("", "")])
```
mauritsvanrees, line 145: Can we do `address = address.strip()` first, for good measure?  Actually I would do that a few lines above, after the `isinstance` check.

When we are being called by our own `validateSingleEmailAddress` there is probably nothing to strip away, but someone else may call this method.

mauritsvanrees (commented):
Thanks for working on this.  I have a few suggestions.
`PloneTool.py` has these methods:

* `validateEmailAddresses`: validate a string with possibly multiple email addresses.  For each address this calls `validateSingleNormalizedEmailAddress`.
* `validateSingleEmailAddress`: validate a string with at most ONE email address.  For this address this calls `validateSingleNormalizedEmailAddress`.
* `validateSingleNormalizedEmailAddress` only needs to handle one address, obviously.

- [x] In all three methods: after checking if it is a string, can we do a `.strip()`?
- [x] In the two methods handling only a single address, can we do `address.splitlines()` and exit when this returns more than one line?  In both cases this can replace the use of `EMAIL_CUTOFF_RE`.

BTW, I agree with your removing this line from the valid inputs:

```
 user@example.org\n user2@example.org",  # omitting comma is ok
```

Products/CMFPlone/PloneTool.py
petschki: yes good point ... I'll change that.

petschki:
> BTW, I agree with your removing this line from the valid inputs:
> 
> ```
>  user@example.org\n user2@example.org",  # omitting comma is ok
> ```

I first moved this line to the invalid inputs but then it fails on current py3.10 jenkins with no strict parsing.

petschki:

petschki:
Currently its hard to get a green jenkins ...

petschki:

rohnsha0:

rohnsha0:
Is Node4 experiencing flakiness again, or could the issues be related to code changes?

petschki:
@rohnsha0 please see progress here #4019

rohnsha0:
Ah, I understood... thanks for providing more info!

petschki:
Currently the build is broken because of this strict parsing which got backported to 3.10 too (see https://jenkins.plone.org/job/plone-6.1-python-3.10/371/) ... anyone willing to approve here?

davisagli (approved):
LGTM, thanks for working on this

rohnsha0 (approved):
LGTM

PR #4026, opened 2024-10-09
Avoid POSKeyError when commit occurs and we have savepoint that involves Plone Site

wesleybl:
When ZODB handles savepoint and we have changes in Plone Site at that savepoint, it changes the `_p_estimated_size` attribute of Plone Site. This is an assignment to one of the special persistency attributes (identified by an `_p_` name prefix); it should happen without access to any other attributes of obj. But obj._tree is accessed in `__setattr__` of PloneSite class and results in a ZODB load which apparently fails.

See: https://github.com/plone/plone.restapi/pull/1823#issuecomment-2402855105

### DIFF
--- a/Products/CMFPlone/Portal.py
+++ b/Products/CMFPlone/Portal.py
@@ -61,7 +61,7 @@ class PloneSite(Container, SkinnableObjectManager):
 
     def __setattr__(self, name, obj):
         # handle re setting an item as an attribute
-        if self._tree is not None and name in self:
+        if not name.startswith("_") and self._tree is not None and name in self:
             del self[name]
             self[name] = obj
         else:
--- a/news/4026.bugfix
+++ b/news/4026.bugfix
@@ -1,0 +1,1 @@
+Avoid POSKeyError when commit occurs and we have savepoint that involves Plone Site. @wesleybl


wesleybl:

wesleybl:
@d-maurer FYI

Products/CMFPlone/Portal.py
```diff
@@ -61,7 +61,7 @@ def __getattr__(self, name):
 
     def __setattr__(self, name, obj):
         # handle re setting an item as an attribute
-        if self._tree is not None and name in self:
+        if not name.startswith("_") and self._tree is not None and name in self:
```
ale-rt, line 64: Shouldn't we be more strict with this change and check for `name.startswith("_p_") or `name.startswith("_v_")``?

d-maurer:
Alessandro Pisa wrote at 2024-10-14 22:28 -0700:
>@ale-rt commented on this pull request.
>
>> @@ -61,7 +61,7 @@ def __getattr__(self, name):
>
>     def __setattr__(self, name, obj):
>         # handle re setting an item as an attribute
>-        if self._tree is not None and name in self:
>+        if not name.startswith("_") and self._tree is not None and name in self:
>
>Shouldn't we be more strict with this change and check for `name.startswith("_p_") or `name.startswith("_v_")``?

Obviously, it depends on what is put into `_tree` in the first place.

When I looked at the `__setattr__` implementation, I got the impression
that the `_tree` is there to let the `PloneSite` emulate the behavior
of a "Large Folder", i.e. to accomodate a large number of content items.
In the "Large Folder" case, only content items are put into `_tree`,
"normal" attributes are stored directly on the object.
Should `PloneSite` do the same, then `startswith("_")` is sufficient
because the id of content items must not start with `_` (they would not
be publishable otherwise).

Should `PloneSite` store arbitrary attributes in `_tree`,
then `startswith("_p_")` might not be correct:
while all special persistency attributes have this name prefix,
the prefix can be used for other attributes as well and then
might be stored in `_tree` (event though it would not arrive there
via normal attribute assignment).

rohnsha0:

davisagli (approved):
@d-maurer @wesleybl Thanks for the analysis and fix. I've seen this error before but hadn't figured out the scenario when it happens.

PR #4027, opened 2024-10-10
add new permission for aliases controlpanel

jackahl:
as requested in: https://github.com/plone/plone.restapi/issues/1820#issuecomment-2383449468


jackahl:

davisagli:
I see we have existing permissions "Manage portal aliases" and "Manage context aliases" which you used in https://github.com/plone/plone.restapi/pull/1825. I wasn't aware of those, it makes sense to use them rather than adding a new one.

PR #4038, opened 2024-10-26
Fix possible `Pdata` objects in `OFS.File` resources

petschki:
If you upload a theme in the theming controlpanel and the CSS or JS files are big, you get a `OFS.Pdata` object from `resource.data` ... this can be resolved by using `bytes(resource.data)`


petschki:

davisagli (approved):

PR #4042, opened 2024-11-04
Allow Contributor and Editor to Access inactive portal content

MrTango:


MrTango:

MrTango:
@mauritsvanrees this needs an upgrade step to load the rolemap,  where in plone.app.upgrade should this be added?

MrTango:
I would argue that the new behavior is the better one. Right now only administrators can really work with content which have effective date in the future. For me that's a bug, which needs to be corrected. We also should back port this at least to Plone 6.0.

davisagli:
> I would argue that the new behavior is the better one.

I agree it's a better default. If an admin customized the permissions to give this permission to non-default roles, I'm not sure we should clobber that.

petschki:

petschki (approved):

petschki:
The test failures seem to be related to this change: https://jenkins.plone.org/job/pull-request-6.1-3.10/508/testReport/plone.restapi.tests.test_search/TestSearchFunctional/test_respect_access_inactive_permission/

MrTango:

PR #4048, opened 2024-11-07
Fixes proper pat-plone-modal config to show user addition info

### LINKED ISSUE #4044: Adding a user in classic does not yield any feedback
Adding a user using the modal-forms in Plone Classic (6.0.x and 6.1.x) does not yield the expected portal-message "User added" (that is indeed triggered by the AddUserForm). Especially when the option "Many Users?" is active it may be very confusing. 

The pat-plone-modal config in `Products/CMFPlone/controlpanel/browser/usergroups_usersoverview.pt` needs `disableAjaxFormSubmit`.

Currently it is:
```data-pat-plone-modal='{
  "actionOptions": {
    "redirectOnResponse": true,
    "redirectToUrl": "${portal_url}/@@usergroup-userprefs"
  }
}'
```

I think it should be:
```data-pat-plone-modal='{
  "actionOptions": {
    "disableAjaxFormSubmit": true
  }
}'
```

The same is true for groups and `Products/CMFPlone/controlpanel/browser/usergroups_groupsoverview.pt`.

rohnsha0:
fixes #4044

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/usergroups_usersoverview.pt
+++ b/Products/CMFPlone/controlpanel/browser/usergroups_usersoverview.pt
@@ -64,7 +64,8 @@
              data-pat-plone-modal='{
                 "actionOptions": {
                   "redirectOnResponse": true,
-                  "redirectToUrl": "${portal_url}/@@usergroup-userprefs"
+                  "redirectToUrl": "${portal_url}/@@usergroup-userprefs",
+                  "disableAjaxFormSubmit": true
                 }
               }'
              tal:attributes="href string:${portal_url}/@@new-user"


petschki (changes requested):
As @pbauer mentioned in the issue, please use only `disableAjaxFormSubmit: true` and remove `redirect*` options ... otherwise these interfere.

Notes: 

What I didn't know is, that the addform redirects with the `searchstring=<newusername>` URL query parameter, so the list shows **only** the added user. while this is nice if you have activated "many users" its maybe confusing when you have only 10 users and after adding a new one the list shows only one user ... we could think of adding `searchtring` only if `many_users` is activated?

And please do the same for `usergroups_groupsoverview.pt` with the additional `content` options, so that the tabs are not shown:

```
data-pat-plone-modal='{
  "content": "#content-core",
  "actionOptions": {
    "disableAjaxFormSubmit": true
  }
}'
```

petschki:
PS: and don't forget a `news` entry for the changelog

rohnsha0:
> As @pbauer mentioned in the issue, please use only `disableAjaxFormSubmit: true` and remove `redirect*` options ... otherwise these interfere.
> 
> Notes:
> 
> What I didn't know is, that the addform redirects with the `searchstring=<newusername>` URL query parameter, so the list shows **only** the added user. while this is nice if you have activated "many users" its maybe confusing when you have only 10 users and after adding a new one the list shows only one user ... we could think of adding `searchtring` only if `many_users` is activated?
> 
> And please do the same for `usergroups_groupsoverview.pt` with the additional `content` options, so that the tabs are not shown:
> 
> ```
> data-pat-plone-modal='{
>   "content": "#content-core",
>   "actionOptions": {
>     "disableAjaxFormSubmit": true
>   }
> }'
> ```

disabling/ remove redirect lines effect the message,

when redirect it shows `Info: User Added`
when not redirecting it shows `Info: Changes Applied`

rohnsha0:
`Elements states '['detached']' (list) should contain '['visible']' (list)`

what does the above error signify? when running locally it works!

petschki:
This test checks, that if the form data isn't valid the modal should remain open with the error message and the invalid form. This is indeed broken, because `disableAjaxFormSubmit` always closes the modal and redirects to the invalid form in the main browser.

You have to make sure, that you run this robottest scenario with:

```
$ bin/rfbrowser init chrome
$ ROBOT_BROWSER=headlesschrome bin/test --all -t "user overlay remains on wrong data"
```

Now this is odd because (at the time of writing this) we cannot use an ajax modal form and show the success message in the main window but stay for errors in the modal at the same time. The only solution would be to remain in the modal on success and on error and the admin has to close the modal manually. (Side effect: the userlist shows up in the modal ... that's not nice 🤨 )

`pat-plone-modal` has the option `displayInModal` which defaults to `true`, that means after you have successfully submitted a form, you remain inside the modal. If you set this to false, the pattern should "keep" the success message and show it in the main window after reloading it.

rohnsha0:
Sorry, but I am unsure how to run robottest using that mentioned scenarious...

stevepiercy:
It looks like we need to enhance the [Test locally](https://6.docs.plone.org/contributing/core/index.html#test-locally) section of the docs with when and how to run [`rfbrowser`](https://6.docs.plone.org/search.html?q=rfbrowser).

Anyway, @rohnsha0 if you take a look at how Jenkins runs its CI, you might find something useful in its [console](https://jenkins.plone.org/job/pull-request-6.1-3.12/855/console). It's only a machine that runs the same tests that us mere mortals manually run.

stevepiercy:
@rohnsha0, @petschki just created a PR with instructions, and I merged it. Instructions should appear at https://6.docs.plone.org/contributing/core/#test-locally

rohnsha0:
> @rohnsha0, @petschki just created a PR with instructions, and I merged it. Instructions should appear at https://6.docs.plone.org/contributing/core/#test-locally

thanks for the efforts @petschki @stevepiercy

rohnsha0:

rohnsha0:

rohnsha0:

PR #4054, opened 2024-11-08
Allow bundles to be rendered after "theme" and "custom" bundles.

thet:
JS and CSS resources can now be rendered after all other resources in their
resource group including the theme (e.g. the Barceloneta theme CSS).

There is a exception for custom CSS which can be defined in the theming
controlpanel. This one is always rendered as last style resource.

To render a resource after all other give it the "depends" value of "all". This
indicates that the resource depends on all other being rendered before. The
resource is then rendered as last resource of it's resource group.

This allows to override a theme with custom CSS from a bundle instead of having
to add the CSS customizations to the registry via the "custom_css" settings.
As a consequence, theme customization can now be done in the filesystem in
ordinary CSS files instead of being bound to a time consuming workflow which
involces upgrading the custom_css registry after every change.

## Docs
https://github.com/plone/documentation/pull/1802
https://github.com/plone/documentation/pull/1803

### DIFF
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -107,7 +107,9 @@ class ResourceBase:
             valid_dependencies = []
 
             for name in depend_names:
-                if name in bundles:
+                if name in bundles or name == "all":
+                    # A valid dependency or special dependency "all", which
+                    # specifies that the bundle should be rendered last.
                     valid_dependencies.append(name)
                     continue
                 if name in all_names:
@@ -133,7 +135,13 @@ class ResourceBase:
 
             return valid_dependencies
 
-        # register
+        # Helper vars for the `all` dependency.
+        last_js_bundle = None
+        last_css_bundle = None
+        depends_all_js = []
+        depends_all_css = []
+
+        # Register resources
         for name, record in records.items():
             include = record.enabled
             include = include or name in theme_enabled_bundles
@@ -146,7 +154,7 @@ class ResourceBase:
                 if depends == "__broken__":
                     continue
                 external = self.is_external_url(record.jscompilation)
-                PloneScriptResource(
+                resource = PloneScriptResource(
                     context=self.context,
                     name=name,
                     depends=depends,
@@ -163,12 +171,19 @@ class ResourceBase:
                     integrity=not external,
                     **{"data-bundle": name},
                 )
+                if "all" in depends:
+                    # Remember the JS bundles that should be rendered last
+                    depends_all_js.append(resource)
+                else:
+                    # Remember the last JS bundle that was added and not be
+                    # rendered last.
+                    last_js_bundle = name
             if record.csscompilation:
                 depends = check_dependencies(name, record.depends, css_names)
                 if depends == "__broken__":
                     continue
                 external = self.is_external_url(record.csscompilation)
-                PloneStyleResource(
+                resource = PloneStyleResource(
                     context=self.context,
                     name=name,
                     depends=depends,
@@ -183,6 +198,13 @@ class ResourceBase:
                     rel="stylesheet",
                     **{"data-bundle": name},
                 )
+                if "all" in depends:
+                    # Remember the JS bundles that should be rendered last
+                    depends_all_css.append(resource)
+                else:
+                    # Remember the last JS bundle that was added and not be
+                    # rendered last.
+                    last_css_bundle = name
 
         # Collect theme data
         themedata = {}
@@ -213,6 +235,7 @@ class ResourceBase:
                 integrity=not external,
                 **{"data-bundle": "diazo"},
             )
+            last_js_bundle = "theme"
 
         # add Theme CSS
         if themedata["production_css"]:
@@ -236,8 +259,29 @@ class ResourceBase:
                 rel="stylesheet",
                 **{"data-bundle": "diazo"},
             )
+            last_css_bundle = "theme"
 
-        # add Custom CSS
+        # Fix "all" dependencies to be rendered after the last known non-all resource.
+        for resource in depends_all_js:
+            # Change all "all" to last known JS resource
+            dependencies = map(
+                lambda dep: dep if dep != "all" else last_js_bundle, resource.depends
+            )
+            # Remove None values, e.g. if no known JS resource was found
+            dependencies = filter(lambda dep: dep is not None, dependencies)
+            resource.depends = list(dependencies)
+
+        for resource in depends_all_css:
+            # Change all "all" to last known JS resource
+            dependencies = map(
+                lambda dep: dep if dep != "all" else last_css_bundle, resource.depends
+            )
+            # Remove None values, e.g. if no known JS resource was found
+            dependencies = filter(lambda dep: dep is not None, dependencies)
+            resource.depends = list(dependencies)
+
+        # Add Custom CSS
+        # This one is always rendered last.
         registry = getUtility(IRegistry)
         theme_settings = registry.forInterface(IThemeSettings, False)
         if theme_settings.custom_css:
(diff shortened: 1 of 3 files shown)


Products/CMFPlone/resources/browser/resource.py
```diff
@@ -87,6 +87,11 @@ def update(self):
         # collect names
         js_names = [name for name, rec in records.items() if rec.jscompilation]
         css_names = [name for name, rec in records.items() if rec.csscompilation]
+        # Allow bundles to have "theme" resp. "theme" and "custom" as dependency
+        # so that bundles can be rendered after theme and custom resources.
+        js_names += ["theme"]
+        css_names += ["theme", "custom"]
```
thet, line 93: See below in https://github.com/plone/Products.CMFPlone/pull/4054/files#diff-9e095767a08db89a248c84c46d31c75898eed2d7dc40b0757a835bb2af0ed95bR143 and https://github.com/plone/Products.CMFPlone/pull/4054/files#diff-9e095767a08db89a248c84c46d31c75898eed2d7dc40b0757a835bb2af0ed95bR165

If "theme" or "custom" does not exist (both added in line 192+) the bundle dependency cannot be resolved and the bundle is not included in the rendered output.

thet:

stevepiercy:
This needs to be documented, but I'm not sure where. Perhaps in [Theming of Classic UI](https://6.docs.plone.org/classic-ui/theming/index.html)? That section is so sad.

jensens (approved):
I think this is a good enhancement. Anyway, this needs some documentation, since it is not obvious.

jensens:
> This needs to be documented, but I'm not sure where. Perhaps in [Theming of Classic UI](https://6.docs.plone.org/classic-ui/theming/index.html)? That section is so sad.

I think this should go into https://6.docs.plone.org/classic-ui/static-resources.html

petschki (changes requested):
Somehow I could not get this to work correctly. I checked out your branch and tested in with the `plone-fullscreen` bundle:

1. I added `custom` to depends, but the CSS was not added... also the `.js` is missing (because custom is only for CSS defined, but the bundle has JS also)
2. I added `theme` to depends and get the following traceback:
```
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 181, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 391, in publish_module
  Module ZPublisher.WSGIPublisher, line 285, in publish
  Module ZPublisher.mapply, line 98, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module zope.browserpage.simpleviewclass, line 44, in __call__
  Module Products.Five.browser.pagetemplatefile, line 127, in __call__
  Module Products.Five.browser.pagetemplatefile, line 59, in __call__
  Module zope.pagetemplate.pagetemplate, line 134, in pt_render
  Module Products.PageTemplates.engine, line 368, in __call__
  Module z3c.pt.pagetemplate, line 198, in render
  Module chameleon.zpt.template, line 328, in render
  Module chameleon.template, line 229, in render
  Module chameleon.utils, line 20, in raise_with_traceback
  Module chameleon.template, line 200, in render
  Module 1ac8f88237956af17a352f1fc406ccdd, line 1880, in render
  Module 985090d925cc459bf0e69fde89f3e030, line 903, in render_master
  Module 615bdf9739ecd0ffd4d4eee8b7a6ebd5, line 498, in render_master
  Module zope.contentprovider.tales, line 76, in __call__
  Module zope.viewlet.manager, line 157, in update
  Module zope.viewlet.manager, line 163, in _updateViewlets
  Module Products.CMFPlone.resources.browser.resource, line 260, in update
  Module webresource._api, line 719, in render
  Module webresource._api, line 678, in resolve
webresource._api.ResourceMissingDependencyError: webresource._api.ResourceMissingDependencyError: Resource defines missing dependency: <PloneScriptResource name="plone-fullscreen", depends="['theme']">
```

what am I missing here?

jensens:
> > IMO custom should stay the last one.
> 
> +1 I think that would be the expected behavior. If I change something TTW it should override everything else (without using `!important` all the time)

I too agree

thet:

thet:

stevepiercy:
Should I use the news item as the documentation for this new feature?

https://github.com/plone/Products.CMFPlone/pull/4054/files#diff-b79fc648ba65de1bd4da45d2bd3181b51ec6ff5b95fcb0cc91125609d52ad0fc

stevepiercy (changes requested):
While working on the documentation, I found a bug in the control panel that should be corrected either in this PR or a related one in the correct repository, if not this one.

- Visit https://classic.demo.plone.org/@@theming-controlpanel#autotoc-item-autotoc-2 (Site Setup > Theming > Advanced settings > Custom Styles)
- Click the link [Theming documentation](https://docs.plone.org/adapt-and-extend/theming), which goes to a 404.

The correct link and label are [Theming of Classic UI](https://6.docs.plone.org/classic-ui/theming/index.html).

Screenshot 2024-12-03 at 7 35 42 PM

stevepiercy:
PR created to address the bug in the control panel UI at https://github.com/plone/plone.app.theming/pull/248.

While I'm in there, should I update any other control panel message?

Products/CMFPlone/tests/testResourceRegistries.py
```diff
@@ -19,30 +20,32 @@
...
+):
+    registry = getUtility(IRegistry)
+
+    bundles = registry.collectionOfInterface(IBundleRegistry, prefix="plone.bundles")
+    bundle = bundles.add(name)
+    bundle.name = name
+    bundle.jscompilation = f"http://foo.bar/{name}.js"
+    bundle.csscompilation = f"http://foo.bar/{name}.css"
+    bundle.depends = depends
+    return bundle
```
ale-rt, line 35: What is the reason to move this code outside of the class?

Products/CMFPlone/resources/browser/resource.py
```diff
@@ -236,8 +259,29 @@ def check_dependencies(bundle_name, depends, bundles):
...
+
+        # Fix "all" dependencies to be rendered after the last known non-all resource.
+        for resource in depends_all_js:
+            # Change all "all" to last known JS resource
+            dependencies = map(
+                lambda dep: dep if dep != "all" else last_js_bundle, resource.depends
+            )
+            # Remove None values, e.g. if no known JS resource was found
+            dependencies = filter(lambda dep: dep is not None, dependencies)
+            resource.depends = list(dependencies)
```
ale-rt, line 272: I think that use `map` and `filter` with two custom lambdas and then casting to a list can be replaced with a much more clear for cycle and probably performant for cycle.

petschki (changes requested):
@thet I think we should rethink this and use the grouping api of `webresource`. If we move the resources with `depends="all"` to a new "deferred" ResourceGroup which gets resolved after the "theming" bundles then `webresource` does our work. See my experimental commit here: https://github.com/plone/Products.CMFPlone/commit/92a82ec016b40e63ec215f8cf12e3f18988e7209 ...

Products/CMFPlone/tests/testResourceRegistries.py
ale-rt: But in the other class it is defined as a method... https://github.com/plone/Products.CMFPlone/pull/4054/files#diff-8ef40cff64bf27605c753443d198309a61a480a09ec62d66fed8152c607cb600R264

IMO leave it like it was or make the test classes inherit from a common class.

Products/CMFPlone/resources/browser/resource.py
```diff
@@ -236,8 +259,29 @@ def check_dependencies(bundle_name, depends, bundles):
...
+
+        for resource in depends_all_css:
+            # Change all "all" to last known JS resource
+            dependencies = map(
+                lambda dep: dep if dep != "all" else last_css_bundle, resource.depends
+            )
+            # Remove None values, e.g. if no known JS resource was found
```
ale-rt, line 279: > Remove None values, e.g. if no known JS resource was found

This is for CSS right?
The code in this cycle and in the one above looks the same.
If the treatment for the resources is equal maybe it is better to move it to a method?

petschki:
Closed this in favor of #4076

PR #4060, opened 2024-11-13
Update content filtering options for TinyMCE

### LINKED ISSUE #3204: Default valid_elements field in pattern options adds 40kb for every RichtTextField on an edit page
Plone 5.2.3 coredev.  While checking for something else I noticed a huge list in the pattern-options of the pat-mimetypeselector field.  This is a small drop down to switch the mimetype of our RichText fields and it contains the default pattern options for TinyMCE based on settings in the control panel registry. 

The vield valid_elements is a huge string which seems to be a matrix for all tags and all theoretically possible attributes on any tag:

```"valid_elements": "a[datetime|char|coords|width|cols|shape|rules|rowspan|border|lang|prompt|maxlength|href|hreflang|compact|tabindex|longdesc|size|dir|hspace|media|start|disabled|axis|align|nowrap|rows|accept|usemap|for|type|valign|span|clear|method|target|src|name|accept-charset|id|ismap|headers|nohref|frame|noshade|alt|height|readonly|action|selected|cellspacing|charset|cite|scope|summary|accesskey|label|rev|checked|class|colspan|abbr|color|multiple|enctype|title|value|rel|vspace|cellpadding|charoff|align|style],abbr[datetime|char|coords|width|cols|shape|rules|rowspan|border|lang|prompt|maxlength|href|hreflang|compact|tabindex|longdesc|size|dir|hspace|media|start|disabled|axis|align|nowrap|rows|accept|usemap|for|type|valign|span|clear|method|target|src|name|accept-charset|id|ismap|headers|nohref|frame .... ```

It adds about 40Kb to the edit page of the default core Plone frontpage, going from 63,98 Kb to 108,48Kb . This edit page only contains 1 rich text area. This string will be added for every RichTextField(widget). :-O  

I tested this by escaping the last line from this part of the TinyMCESettingsGenerator::

https://github.com/plone/Products.CMFPlone/blob/cba026e32b3bde5a00f2301b938b409e96cf7bed/Products/CMFPlone/patterns/tinymce.py#L209-L226

It seems to me the generated valid_elements is completely incorrect, probably a left over or wrong conversion from the Plone 4 filter control panel where you could list valid attributes per html tag/container.  

But do we really need to pass TinyMCE any valid_elements when all html is filtered on save anyway server side? What's this still doing here?

### PR DESCRIPTION
**Pull request by petschki**
fixes #3204 

This reduces the loaded page size for forms with tinymce
(tested with `curl --user admin:admin -so /dev/null http://localhost:8080/Plone/edit -w "%{size_download}"`)

Before: 128.620 Bytes
After: 75.674 Bytes

### DIFF
--- a/Products/CMFPlone/patterns/tinymce.py
+++ b/Products/CMFPlone/patterns/tinymce.py
@@ -1,3 +1,4 @@
+from lxml import html
 from plone.app.theming.utils import theming_policy
 from plone.base.interfaces import IFilterSchema
 from plone.base.interfaces import ITinyMCESchema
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by petschki** [comment]

**Comment by petschki** [comment]
To demonstrate the impact: `collective.easyform` has 4 TinyMCE editors in the schema:

Before: 343.143 Bytes
After: 127.947 Bytes

😉

File: Products/CMFPlone/patterns/tinymce.py [hunk 1]
```diff
@@ -177,22 +174,13 @@ def get_tiny_config(self):
...
-            tiny_valid_elements = []
-            for tag in valid_tags:
-                tag_str = "{}[{}]".format(tag, "|".join(valid_attributes))
-                tiny_valid_elements.append(tag_str)
-            # We want to remove the nasty tag including the content in the
-            # backend, so TinyMCE should allow them here.
-            for tag in nasty_tags:
-                tag_str = "{}[{}]".format(tag, "|".join(valid_attributes))
-                tiny_valid_elements.append(tag_str)
-            tiny_config["valid_elements"] = ",".join(tiny_valid_elements)
```
MrTango, line 195: Are you sure we don't need this anymore?
We need to make sure, that what ever is set in the HTML filter control panal is not in conflict with what TinyMCE filtering does. Is that's the case I'm fine with the simplification.

petschki: You're right. Only relying on the "html5" schema doesn't fit our `safe_html` transformation. I've updated this to fix our test scenarios again.

petschki:

petschki:
I'll close this... need to think about it again.

PR #4076, opened 2024-12-05
Implement resource bundle depends on `all` to be rendered at last

petschki:
This is a successor of #4054 with the simplified `webresource.ResourceGroup` approach.
Tests are copied 1:1 from the previous PR

PR #4077, opened 2024-12-06
Implement resource bundle depends on `all` to be rendered at last (6.0 branch)

thet:
Plone 6.0 backport of https://github.com/plone/Products.CMFPlone/pull/4076

### DIFF
--- a/Products/CMFPlone/resources/browser/resource.py
+++ b/Products/CMFPlone/resources/browser/resource.py
@@ -79,8 +79,14 @@ class ResourceBase:
         registry_group_js = webresource.ResourceGroup(
             name="registry_js", group=root_group_js
         )
+        registry_group_js_deferred = webresource.ResourceGroup(
+            name="registry_js_deferred",
+        )
         registry_group_css = webresource.ResourceGroup(
             name="registry_css", group=root_group_css
+        )
+        registry_group_css_deferred = webresource.ResourceGroup(
+            name="registry_css_deferred",
         )
         records = self.registry.collectionOfInterface(
             IBundleRegistry, prefix="plone.bundles", check=False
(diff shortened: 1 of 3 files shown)


thet:

petschki (approved):

stevepiercy:
Just to confirm, for `all`, we documented that they will appear in the order declared, but here they will appear in alphabetic order. Should I change the docs accordingly?

https://github.com/plone/documentation/pull/1802/files#diff-eab566b0dcb958909e31d49af95b85b45c6197064b0afb537726cd61e467e2e7R106

petschki:
> Just to confirm, for `all`, we documented that they will appear in the order declared, but here they will appear in alphabetic order. Should I change the docs accordingly?
> 
> https://github.com/plone/documentation/pull/1802/files#diff-eab566b0dcb958909e31d49af95b85b45c6197064b0afb537726cd61e467e2e7R106

ResourceRegistry always sorts the bundles alphabetically, so they will appear in the alphabetically order no matter when you declare them ...

thet:
> > Just to confirm, for `all`, we documented that they will appear in the order declared, but here they will appear in alphabetic order. Should I change the docs accordingly?
> > https://github.com/plone/documentation/pull/1802/files#diff-eab566b0dcb958909e31d49af95b85b45c6197064b0afb537726cd61e467e2e7R106
> 
> ResourceRegistry always sorts the bundles alphabetically, so they will appear in alphabetically order no matter when you declare them ...

@stevepiercy my assumption was that the resource registry doesn't do alphabetical sorting, therefore several references in comments and docs that it does sort as added. But that was apparently wrong. I fixed it in this PR.
Thanks for fixing it in the docs!

PR #4066, opened 2024-11-27
Python 3.13.x: `unittest.makeSuite` got removed.

petschki:
I use this PR as POC for the change: https://github.com/zopefoundation/zope.security/pull/84/files and to summarize all occurrences in the core (addon) packages (and its PRs):

- [x] collective.monkeypatcher (https://github.com/plone/collective.monkeypatcher/pull/14)
- [x] plone.app.upgrades (https://github.com/plone/plone.app.upgrade/pull/331)
- [x] plone.keyring (https://github.com/plone/plone.keyring/pull/33)
- [x] plone.portlet.static (https://github.com/plone/plone.portlet.static/pull/45)
- [x] plone.protect (https://github.com/plone/plone.protect/pull/121)
- [x] plone.testing (https://github.com/plone/plone.testing/pull/100 README only)
- [x] plone.theme (https://github.com/plone/plone.theme/pull/27)
- [ ] Products.PortalTransforms (https://github.com/plone/Products.PortalTransforms/pull/70)
- [ ] z3c.relationfield (https://github.com/zopefoundation/z3c.relationfield/pull/28)

PR #4081, opened 2024-12-17
Remove unittest.makeSuite for Py 3.13 compatibility [6.0]

mauritsvanrees:
This back ports PR #4066 from @petschki to Plone 6.0.
Here is needed an extra fix because 6.0 still has a `testPropertiesTool.py` file which used `makeSuite`.


mauritsvanrees:

davisagli (approved):

PR #4082, opened 2024-12-18
Fix robottests for `pat-contentbrowser` with new component registry

petschki:
Use together with https://github.com/plone/plone.staticresources/pull/362


1letter (approved):
LGTM, thanks.

PR #4089, opened 2025-01-14
Update declared Python support

davisagli:
Plone 6.0.14 dropped support for Python 3.8 and added support for 3.13, as documented in https://github.com/plone/documentation/pull/1823 and https://plone.org/download/release-schedule


davisagli:

stevepiercy (approved):
LGTM

mauritsvanrees (changes requested):
I would prefer to only update the classifiers, and not touch `python_requires`.

Until now, our policy, or maybe just my own preference, was to not explicitly break support for Python 3.8.  I think I have written or said this on occasion.  If for some reason you still want to run Plone 6.0 on Python 3.8, you should be able to do so, without getting an error on installation because your Python version does not match with the `python_requires` setting.

You would be on your own for getting the correct versions of other packages, as our official `versions.cfg` may include versions that do not work on Python 3.8.  And subtle things could go wrong, maybe someone starts using syntax that only works on Python 3.9+.  I would actually consider that an error, which we could fix in a new release, if the effort is reasonable.

In short: unless we somehow know that Plone 6.0 cannot possibly run on Python 3.8, I think we should leave `python_requires` as it is.

Would you agree?

BTW, I do think we should start showing a warning in the Site Setup when using an outdated Python version, just like we do in Plone 5.2.  Plus warn when Plone 6.0 is out of security support, but that will still be a few years.  I would probably skip the warnings about being out of maintenance.  But this can be something for another PR.

davisagli:
@mauritsvanrees I have some misgivings about it (it seems like it will make it too easy for someone to install a broken site with Python 3.8 but not realize it until they have a really obscure error) but I'll let you make the decision.

mauritsvanrees (approved):

mauritsvanrees:
Thanks!

mauritsvanrees:
A few days ago I updated `python_requires` to [require 3.9 minimum](https://github.com/plone/Products.CMFPlone/pull/4142) after all.  Will be in 6.0.15rc1.

PR #4099, opened 2025-02-03
Filter inactive items also for admins.

### LINKED ISSUE #4098: Filtering expired/scheduled items should be possible even for managers
## Logic problem in searchResults

This follows up on earlier work done in #1952 and #3571.

IMO the code there has a logic bug, see below. ~Fixing that may break other code relying on the buggy behavior, so let's discuss.~ The fix I made retains the old default behavior and should be backward compatible.

I want to show a listing of news items, shown to somebody with sidewide Administrator rights, that shows only the news items as visible to 'normal' users. So my editors, who sometimes have admin rights, don't panic about something showing up when it's actually expired/scheduled and nobody except themselves will get to see it.

### What I did:

Call `searchResults(show_inactive=False)` 

### What I expect to happen:

 I expect to not have any inactive content in the returned results.

### What actually happened:

The returned search results contain inactive content, if the user performing the search has permission `AccessInactivePortalContent`.

Products.CMFPlone current master HEAD 
https://github.com/plone/Products.CMFPlone/blob/c8694d8ef8d1383afd8a744bb10122e7415c2704/Products/CMFPlone/CatalogTool.py#L392

### Details

Logically, we have four scenarios. Let's explore them in pseudocode that closely mimics the logic involved here.

#### 1. `show_inactive==False` (the default) and the user does NOT have `AccessInactivePortalContent`

Results should be filtered.

```
>>> show_inactive = False
>>> allow_inactive = False
>>> filter_effective = not show_inactive and not allow_inactive
>>> filter_effective
True
```

OK.

#### 2. `show_inactive==True`  and the user does NOT have `AccessInactivePortalContent`

Results should not be filtered. 

_I'm surprised that you'd want to bypass security that way, but this is clearly the intent of the code as documented in the comment-as-docstring of the searchResults method, so let's stick with that._

```
>>> show_inactive = True
>>> allow_inactive = False
>>> filter_effective = not show_inactive and not allow_inactive
>>> filter_effective
False
```

OK

#### 3. `show_inactive==True`  and the user DOES have `AccessInactivePortalContent`

Results should not be filtered.

```
>>> show_inactive = True
>>> allow_inactive = True
>>> filter_effective = not show_inactive and not allow_inactive
>>> filter_effective
False
```

OK.

#### 4. `show_inactive==False`  and the user DOES have `AccessInactivePortalContent`

Results should be filtered.

```>>> show_inactive = False
>>> allow_inactive = True
>>> filter_effective = not show_inactive and not allow_inactive
>>> filter_effective
False
```

WRONG! And this is the bug I'm reporting.

### Fix

I'll issue a PR to fix scenario (4) without causing regressions to the other 3 scenarios.

gyst:
Fixes #4098.

This keeps the existing default behavior, while allowing suppressing inactive content for admins, if `show_inactive=False` is explicitly passed as a keyword or query.

### DIFF
--- a/Products/CMFPlone/CatalogTool.py
+++ b/Products/CMFPlone/CatalogTool.py
@@ -398,19 +398,29 @@ class CatalogTool(PloneBaseTool, BaseTool):
         # It also accepts a keyword argument show_inactive to disable
         # effectiveRange checking entirely even for those without portal
         # wide AccessInactivePortalContent permission.
+        #
+        # By default, this shows inactive content to users with the required
+        # permission, while hiding it from users without AccessInactivePortalContent.
+        # - To show inactive content to users without the portal wide
+        #   AccessInactivePortalContent permission, pass show_inactive=True
+        #   as either a keyword or query parameter.
+        # - To suppress inactive content for admins, pass show_inactive=False
+        #   as either a keyword or query parameter
 
         # Make sure any pending index tasks have been processed
         processQueue()
 
         kw = kw.copy()
-        show_inactive = kw.get("show_inactive", False)
-        if isinstance(query, dict) and not show_inactive:
-            show_inactive = "show_inactive" in query
+        show_inactive = kw.get("show_inactive", None)
+        if show_inactive is None and isinstance(query, dict):
+            show_inactive = query.get("show_inactive", None)
+        if show_inactive is None:
+            show_inactive = self.allow_inactive(kw)
 
         user = _getAuthenticatedUser(self)
         kw["allowedRolesAndUsers"] = self._listAllowedRolesAndUsers(user)
 
-        if not show_inactive and not self.allow_inactive(kw):
+        if not show_inactive:
             kw["effectiveRange"] = DateTime()
 
         # filter out invalid sort_on indexes
--- a/news/4098.bugfix
+++ b/news/4098.bugfix
@@ -1,0 +1,1 @@
+Enable searchResults filtering of inactive items also for admins, when requested.
--- a/news/4098.bugfix
+++ b/news/4098.bugfix
@@ -1,1 +1,1 @@
-Enable searchResults filtering of inactive items also for admins, when requested.
+In catalog queries, make sure `show_inactive=False` hides inactive items even for admins who have the `Access inactive portal content` permission. @gyst
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,14 @@
 
 <!-- towncrier release notes start -->
 
+## 6.1.0 (2025-02-07)
+
+
+### Bug fixes:
+
+- fixes modal behaviour after form submission in 404 page
+  [rohnsha0] #1636
+
 ## 6.1.0rc1 (2025-01-31)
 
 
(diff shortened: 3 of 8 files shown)


gyst:
Hang on, I think I can fix the issue while keeping the current default behavior.

gyst:
I improved this PR to retain the current default behavior, while still fixing the bug.

Products/CMFPlone/CatalogTool.py
```diff
@@ -398,19 +398,29 @@ def searchResults(self, query=None, **kw):
...
+        #   as either a keyword or query parameter
 
         # Make sure any pending index tasks have been processed
         processQueue()
 
         kw = kw.copy()
-        show_inactive = kw.get("show_inactive", False)
-        if isinstance(query, dict) and not show_inactive:
-            show_inactive = "show_inactive" in query
+        show_inactive = kw.get("show_inactive", None)
```
gyst, line 414: I know I could've just used `kw.get("show_inactive")` here and below, but since this logic depends so crucially on `None` being different from `False`, I preferred the explicit variant.

yurj:
Seems fine to me, show inactive = False is ok. If it is True, that means "let me show inactive" even if the user should not? show_inactive is used here:

https://github.com/plone/Products.CMFPlone/blob/c8694d8ef8d1383afd8a744bb10122e7415c2704/Products/CMFPlone/WorkflowTool.py#L261
"                    # Include inactive content in result list. This is
                    # especially important for content scheduled to go public
                    # in the future, but needs to be reviewed before this."

and here:

https://github.com/plone/Products.CMFPlone/blob/c8694d8ef8d1383afd8a744bb10122e7415c2704/Products/CMFPlone/browser/search.py#L148
# respect effective/expiration date

and some other places:

https://github.com/search?q=org%3Aplone+show_inactive&type=code

gyst:
> If it is True, that means "let me show inactive" even if the user should not? 

Yes, that's unchanged from the current implementation. Also the default behavior remains unchanged for both admins and non-admins. The only change is, that if you're an admin, it was previously hardcoded that you would always see inactive items; and this is still the default; however if you as an admin request `show_inactive=False` it will now honor that setting.

Since `show_inactive=False` before did not work for admins, any code that explicitly asks for that and then still expects inactive items was broken by design and needs to be fixed.

gyst:
> We went to the `/manage_access` and set the "Access inactive portal content" permission for "Authenticated" users, and this way they can see these content.

This will still be required, this PR does not change that.

What it does change, that if you _have_ that permission and you then query `searchResults(show_inactive=False)`, you will actually _not_ see that content.

news/4098.bugfix
```diff
@@ -0,0 +1 @@
+Enable searchResults filtering of inactive items also for admins, when requested.
```
davisagli, line 1:
```suggestion
In catalog queries, make sure `show_inactive=False` hides inactive items even for admins who have the `Access inactive portal content` permission. @gyst
```

davisagli (approved):
Looks okay to me, and thanks for adding tests. Let's run the tests for all Plone packages and make sure there are no regressions:

gyst:
The upgrade test failures are false negatives.

davisagli:
Try again now that I merged the change in plone.restapi:

PR #4104, opened 2025-02-12
Fixes Unvalidated Email Succeeding to group addition

### LINKED ISSUE #933: Users -> Groups -> New group: new email validation
The form for "new group" does not validate a given string for being an email

rohnsha0:
fixes #933

### DIFF
--- a/Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
+++ b/Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
@@ -2,13 +2,28 @@ from Acquisition import aq_inner
 from plone.base import PloneMessageFactory as _
 from plone.protect import CheckAuthenticator
 from Products.CMFCore.utils import getToolByName
-from Products.CMFPlone.controlpanel.browser.usergroups import (
-    UsersGroupsControlPanelView,
-)
+from Products.CMFPlone.controlpanel.browser.usergroups import UsersGroupsControlPanelView
 from Products.statusmessages.interfaces import IStatusMessage
-
+import re
 
 class GroupDetailsControlPanel(UsersGroupsControlPanelView):
+    def validate_email(self, email):
+        """Validate email format using regex pattern."""
+        pattern = r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'
+        return re.match(pattern, email) is not None
+
+    def validate_group_emails(self, emails):
+        """Validate a list of email addresses."""
+        if isinstance(emails, str):
+            emails = [e.strip() for e in emails.split('\n') if e.strip()]
+        
+        invalid_emails = []
+        for email in emails:
+            if not self.validate_email(email):
+                invalid_emails.append(email)
+        
+        return invalid_emails
+
     def get_group_property(self, prop_id):
         try:
             return self.group.getProperty(prop_id, None)
@@ -40,6 +55,19 @@ class GroupDetailsControlPanel(UsersGroupsControlPanelView):
             title = self.request.form.get("title", None)
             description = self.request.form.get("description", None)
             addname = self.request.form.get("addname", None)
+            
+            # Get email field(s) from the form
+            emails = self.request.form.get("email", "")
+            
+            # Validate emails before proceeding
+            invalid_emails = self.validate_group_emails(emails)
+            if invalid_emails:
+                msg = _(
+                    "Invalid email address(es): ${emails}. Please correct them before proceeding.",
+                    mapping={"emails": ", ".join(invalid_emails)},
+                )
+                IStatusMessage(self.request).add(msg, "error")
+                return self.index()
 
             if addname:
                 if not self.regtool.isMemberIdAllowed(addname):
@@ -84,11 +112,20 @@ class GroupDetailsControlPanel(UsersGroupsControlPanelView):
 
             processed = {}
             for id, property in self.gdtool.propertyItems():
-                processed[id] = self.request.get(id, None)
+                value = self.request.get(id, None)
+                # Additional validation for email properties
+                if id.lower().endswith('email') and value:
+                    invalid_emails = self.validate_group_emails(value)
+                    if invalid_emails:
+                        msg = _(
+                            "Invalid email address(es) in ${field}: ${emails}",
+                            mapping={"field": id, "emails": ", ".join(invalid_emails)},
+                        )
+                        IStatusMessage(self.request).add(msg, "error")
+                        return self.index()
+                processed[id] = value
 
             if self.group:
-                # for what reason ever, the very first group created does not
-                # exist
                 self.group.setGroupProperties(processed)
 
             IStatusMessage(self.request).add(msg, type=self.group and "info" or "error")
--- a/news/933.bugfix
+++ b/news/933.bugfix
@@ -1,0 +1,1 @@
+fixes unvalidated email field succeeding in group addition @rohnsha0


rohnsha0:

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,13 +2,28 @@
...
+from Products.CMFPlone.controlpanel.browser.usergroups import UsersGroupsControlPanelView
 from Products.statusmessages.interfaces import IStatusMessage
-
+import re
 
 class GroupDetailsControlPanel(UsersGroupsControlPanelView):
+    def validate_email(self, email):
+        """Validate email format using regex pattern."""
+        pattern = r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'
+        return re.match(pattern, email) is not None
```
ale-rt, line 13: Thanks a lot for your contribution :)

There is for sure some email validator already in our code base. 
Please try to find and reuse the existing code!

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,13 +2,28 @@
...
+
+    def validate_group_emails(self, emails):
```
ale-rt, line 15: What about:
```suggestion
    def extract_invalid_emails(self, emails):
```

ale-rt (changes requested):
Please check my comments. The only thing that I consider a blocker is to not introduce new code for email validation.

rohnsha0:
> Please check my comments. The only thing that I con sider a blocker is to not introduce new code for email validation.

I understand, it was my oversee... However pushed an commit with validation using `PloneTool`

rohnsha0:

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,13 +2,23 @@
 from plone.base import PloneMessageFactory as _
 from plone.protect import CheckAuthenticator
 from Products.CMFCore.utils import getToolByName
-from Products.CMFPlone.controlpanel.browser.usergroups import (
-    UsersGroupsControlPanelView,
-)
+from Products.CMFPlone.controlpanel.browser.usergroups import UsersGroupsControlPanelView
+from Products.CMFPlone.PloneTool import PloneTool
```
ale-rt, line 6:
```suggestion
```

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,13 +2,23 @@
...
 from Products.statusmessages.interfaces import IStatusMessage
 
-
 class GroupDetailsControlPanel(UsersGroupsControlPanelView):
+    def extract_invalid_emails(self, emails):
+        """Validate a list of email addresses."""
+        if isinstance(emails, str):
+            emails = [e.strip() for e in emails.split('\n') if e.strip()]
+
```
ale-rt, line 14:
```suggestion
        
        plone_utils = getToolByName(self.context, "plone_utils")
```
You should use this rather than getting the class.

mauritsvanrees:
@rohnsha0 Did you push everything? I now get an AttributeError for `validate_group_emails`.

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,13 +2,23 @@
...
-
 class GroupDetailsControlPanel(UsersGroupsControlPanelView):
+    def extract_invalid_emails(self, emails):
+        """Validate a list of email addresses."""
+        if isinstance(emails, str):
+            emails = [e.strip() for e in emails.split('\n') if e.strip()]
+        
+        invalid_emails = []
+        for email in emails:
+            if not PloneTool.validateEmailAddresses(email):
```
ale-rt, line 17: This code will break because you are trying to call the method without initializing the class:
```suggestion
            if not plone_utils.validateEmailAddresses(email):
```

But anyway this will not be working because it expects a list:
```
    @security.public
    def validateEmailAddresses(self, addresses):
        # Validate a list of possibly several email addresses, see also
        # validateSingleEmailAddress.
        if not isinstance(addresses, str):
            return False
```

Writing a test would have helped you to debug this issues earlier. I suggest you to do that!

rohnsha0:
> @rohnsha0 Did you push everything? I now get an AttributeError for `validate_group_emails`.

Ah, I renamed  `validate_group_emails` -> `extract_invalid_emails`... missed renaming usage of the same.. but its fixed now! please recheck now!

ale-rt (changes requested):

mauritsvanrees:
Thanks for working on this.

There is a problem when the new email validation fails: the form switches to a form for creating a new group.  This is caused by the line `self.group = None` which is already in the code on master.
You should move this line after your validation.  Actually, it looks like you could actually remove the line.

Also, even with that fixed, any changes that the user has filled in, are gone when there is a validation error: the form shows the current group data.  It should show the data from the request.  One way to do this, would be to change the `get_group_property` method to take the `prop_id` from the request, but only if it is a POST request.

rohnsha0:
> Thanks for working on this.
> 
> There is a problem when the new email validation fails: the form switches to a form for creating a new group. This is caused by the line `self.group = None` which is already in the code on master. You should move this line after your validation. Actually, it looks like you could actually remove the line.

I've removed `self.group=None`

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,13 +2,25 @@
...
+        """Validate a list of email addresses."""
+        if isinstance(emails, str):
+            emails = [e.strip() for e in emails.split('\n') if e.strip()]
+        
+        invalid_emails = []
+
+        plone_utils = getToolByName(self.context, "plone_utils")
+
+        for email in emails:
+            if not plone_utils.validateEmailAddresses(email):
```
ale-rt, line 19: This is still not valid... see my previous comment

rohnsha0:
> Also, even with that fixed, any changes that the user has filled in, are gone when there is a validation error: the form shows the current group data. It should show the data from the request. One way to do this, would be to change the get_group_property method to take the prop_id from the request, but only if it is a POST request.

I've pushed a PR that retains the form data after an unsuccessful attempt. However, the name field is not getting retained!

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
ale-rt: > I'm sorry, I was not sure what to do next.

No need to be sorry!

I think you want to use `validateSingleEmailAddress`. Check this out: https://github.com/plone/Products.CMFPlone/blob/4a765ec624f5b93e6388c2057a8caced954ae958/Products/CMFPlone/PloneTool.py#L158
Also the tests might give you more insights on what you can do: https://github.com/plone/Products.CMFPlone/blob/4a765ec624f5b93e6388c2057a8caced954ae958/Products/CMFPlone/tests/testPloneTool.py#L34

If you want to add a test yourself please do it here: https://github.com/plone/Products.CMFPlone/blob/4a765ec624f5b93e6388c2057a8caced954ae958/Products/CMFPlone/controlpanel/tests/test_controlpanel_usergroups.py

Thanks again for being patient!

rohnsha0: verify again @ale-rt

ale-rt:

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,18 +2,35 @@
 from plone.base import PloneMessageFactory as _
 from plone.protect import CheckAuthenticator
 from Products.CMFCore.utils import getToolByName
-from Products.CMFPlone.controlpanel.browser.usergroups import (
-    UsersGroupsControlPanelView,
-)
+from Products.CMFPlone.controlpanel.browser.usergroups import UsersGroupsControlPanelView
```
ale-rt, line 5: No need to change this line

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,18 +2,35 @@
...
+        """Retrieve group property, prioritizing request data if it's a POST request."""
+
+        if self.request.method == "POST":
+            return self.request.form.get(prop_id, None)
+        
         try:
             return self.group.getProperty(prop_id, None)
         except AttributeError:
-            pass
+            return None
```
ale-rt, line 33: No need to change this line

Products/CMFPlone/controlpanel/browser/usergroups_groupdetails.py
```diff
@@ -2,18 +2,35 @@
...
-)
+from Products.CMFPlone.controlpanel.browser.usergroups import UsersGroupsControlPanelView
 from Products.statusmessages.interfaces import IStatusMessage
 
-
 class GroupDetailsControlPanel(UsersGroupsControlPanelView):
+    def extract_invalid_emails(self, emails):
+        """Validate a list of email addresses."""
+        if isinstance(emails, str):
+            emails = [e.strip() for e in emails.split('\n') if e.strip()]
```
ale-rt, line 12:
```suggestion
            emails = [email.strip() for email in emails.split('\n') if email.strip()]
```
Also does it make sense to split by newline? I do not think so...

ale-rt (changes requested):
This PR prevents:

- setting multiple mail addresses separated by commas+space, see `COMMASPACE` in https://python.readthedocs.io/fr/hack-in-language/library/email-examples.html.
- setting an email + display name/

This might be a breaking change for somebody.
I am not sure, I want that.

CC @plone/framework-team

mauritsvanrees:
Since when do we want to support multiple emails here?  That would surprise me.

ale-rt:
> Since when do we want to support multiple emails here? That would surprise me.

I am just saying that it is actually possible with the current code base.

rohnsha0:
the form now validates multiple mail id(s) as well along with displayName!  @ale-rt @mauritsvanrees

rohnsha0:

rohnsha0:

rohnsha0:
@ale-rt can u reverify with updated implementations

PR #4107, opened 2025-02-13
Start the Plone 6.2 development cycle

ale-rt:
- add `plone-62` as zcml feature
- define `PLONE62MARKER` as `True`
- bump the profile version to 6200

Based on https://github.com/plone/Products.CMFPlone/pull/3861

### DIFF
--- a/Products/CMFPlone/factory.py
+++ b/Products/CMFPlone/factory.py
@@ -27,6 +27,7 @@ _IMREALLYPLONE5 = True
 PLONE52MARKER = True
 PLONE60MARKER = True
 PLONE61MARKER = True
+PLONE62MARKER = True
 
 logger = getLogger("Plone")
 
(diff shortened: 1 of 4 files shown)


ale-rt:

petschki:
Should we create Jenkins jobs for Plone 6.2?

gforcada:
Good point, I would say so, @mauritsvanrees any reason not to? 🤔

mauritsvanrees:
Yes please!
Probably needs some mr.roboto config work on the server as well.

gforcada:
- `mr.roboto` is already aware of it: https://jenkins.plone.org/roboto/branches
  - it will ask to run jobs on python 3.13 and 3.10 (like 6.1) do we want to already restrict to to 3.11? 🤔

gforcada:
- jenkins jobs are there 🎉 see https://github.com/plone/jenkins.plone.org/pull/380

PR #4120, opened 2025-02-19
RegistrationTool: add method principal_id_or_login_name_exists. [6.0]

mauritsvanrees:
This is factored out from the `isMemberIdAllowed` method, which now calls this after checking the allowed member id pattern.

Main use case: let `plone.app.users` call the new method when saving personal information. This is done when email-as-login is used.  In that case, we currently check if the email address would be valid as id, and if there is not yet a user or group with this id or login name. Problem is that some valid emails are not valid as user id, for example an email address with a quote, and this makes the check fail.

I need this for a 6.0 site, so I first make a PR targeted at 6.0.x.  I will make PRs for 6.1.x and master if the idea is approved.

This PR also changes the logic slightly: after checking if a group or user with the given _id_ exists, we now *always* check if there is a user with this _login name_, even if email-as-login is not used.  This should be superfluous in that case, but you never know what people may have done to their site, and it should be fast.

I should look at adding tests, so I open as a draft PR at first.


mauritsvanrees:
Ready for review.

I am testing this on Jenkins for 6.0 in combination with https://github.com/plone/plone.app.users/pull/136

ale-rt (approved):

Issue #4126, opened 2025-02-26
PLIP Replace pkg_resources in normal code


mauritsvanrees:
## PLIP (Plone Improvement Proposal)

We already want to get rid of `pkg_resources` style namespaces, see issue #3928.

But `pkg_resources` in its entirety is basically deprecated, so we should stop calling it "normal" Python code as well.  In fact, some code already does not work anymore, but it depends on how you have installed Plone.  See several comments by me [starting here](https://github.com/plone/Products.CMFPlone/pull/4121#issuecomment-2683407005) in a PR on CMFPlone.

Technically, this does not need to be a PLIP, and we do not need explicit improvement from any team, and this can all be done in a bugfix release of Plone.  But I wanted to make it more visible and more easily findable.  And this affects lots of packages.

## Abstract

Go through all code used in Plone 6, including in Zope code.  Main target is branches that are used in Plone 6.1 and 6.2.  If we have time, fixing it in 6.0 would be welcome as well.

## Motivation

Several uses of `pkg_resources` actually do not work anymore in Python 3.13.  This was kind-of known, but I did not really notice any problems until now.  But here it is.

In a checkout of `buildout.coredev` 6.2, start a Zope instance based on pip, so using the Makefile.

Very important: **use Python 3.13**.  For example with `export PRIMARY_PYTHON=python3.13`.  Or otherwise make sure that `python3` resolves to python3.13.

Maybe first call `make install` twice, just to be sure, there is some stuff still to fix there.  But technically, just one `make` command should be enough:

```
$ make zope-start
...
Traceback (most recent call last):
  File "/Users/maurits/community/plone-coredev/6.2/.venv/lib/python3.13/site-packages/zope/configuration/xmlconfig.py", line 448, in endElementNS
    self.context.end()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/maurits/community/plone-coredev/6.2/.venv/lib/python3.13/site-packages/zope/configuration/config.py", line 748, in end
    self.stack.pop().finish()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/maurits/community/plone-coredev/6.2/.venv/lib/python3.13/site-packages/zope/configuration/config.py", line 918, in finish
    actions = self.handler(context, **args)
  File "/Users/maurits/community/plone-coredev/6.2/.venv/lib/python3.13/site-packages/plone/autoinclude/zcml.py", line 37, in includePluginsDirective
[...]

PR #4129, opened 2025-03-01
Replace pkg_resources usage

gforcada:
See https://github.com/plone/Products.CMFPlone/issues/4126


news/4126.bugfix
```diff
@@ -0,0 +1 @@
+Replace `pkg_resources` with `importlib.metadata`/`packaging` @gforcada
```
stevepiercy, line 1:
```suggestion
Replace `pkg_resources` with `importlib.metadata` and `packaging`. @gforcada
```

stevepiercy (changes requested):
Minor tweak to change log to make it clear that it's two packages replacing `pkg_resources`, and which may make it easier for search engines to index.

stevepiercy (approved):

PR #4142, opened 2025-03-13
Require 3.9 as minimum Python version on 6.0

mauritsvanrees:
We already officially dropped 3.8 support in 6.0.14, but now Zope requires 3.9 as minimum, so we stop pretending you can get Plone running on this no longer security supported Python version. See also [my comment on the release checklist](https://github.com/plone/Products.CMFPlone/issues/4141#issuecomment-2722710581).


mauritsvanrees:

gforcada (approved):

PR #4144, opened 2025-03-19
[6.1.x] Drop pkg_resources

gforcada:
Backport of #4129


mauritsvanrees (approved):
Thanks.
I have adapted this to use `importlib.metadata.version` instead of first getting the distribution and then getting its version.

I have imported it in this way:

```
from importlib.metadata import version as dist_version
```

Otherwise there may be clashes where we already use `version` as variable name somewhere.

mauritsvanrees:

jensens (approved):

PR #4157, opened 2025-04-11
Fix `redirect_after_login` in VHM hosted sites

### LINKED ISSUE #4151: Redirect after Login is broken
After login, the redirect goes to the portal homepage, not to the subpage.

### What I did:

On https://classic.demo.plone.org

- login as manager
- add a page in a folder
- don't publish the page
- logout
- call the added page in your browser
- login
- i see the portal homepage

### What I expect to happen:
- i see the private page that is added before

### What actually happened:
- i see the portal homepage

Plone 6.1.1 (6108)

petschki:
fixes #4151

### DIFF
--- a/Products/CMFPlone/browser/login/login.py
+++ b/Products/CMFPlone/browser/login/login.py
@@ -16,6 +16,7 @@ from z3c.form import button
 from z3c.form import field
 from z3c.form import form
 from z3c.form.interfaces import HIDDEN_MODE
+from zExceptions import NotFound
 from zope.component import getMultiAdapter
 from zope.component import queryMultiAdapter
 from zope.component import queryUtility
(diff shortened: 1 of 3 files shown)


petschki:

1letter:
i can do the review at monday, currentl, i have no proper internet connection in my new home

Products/CMFPlone/browser/login/login.py
1letter, line 1: this fails by me, because i think a wrong usage of `lstrp`

```
# cpath is your plone site id
/nct

# your came_from parameter
/testseite/
```
result: `estseite/`

because lstrip remove all chars in the given paramter lstrip("/nct"), means remove all "/","n","c" and "t" from the left side

1letter (changes requested):
@petschki see my comment of wrong usage "lstrip"

ale-rt:
See https://github.com/plone/Products.CMFPlone/pull/4159, this works for me

szakitibi:
This one seems to still have a bug here:

https://github.com/plone/Products.CMFPlone/blob/d6cd459d8112da8400bdeba99455341015cbdaeb/Products/CMFPlone/browser/login/login.py#L219

I have local Plone 6.1.2 test site, where  I have implemented the fix in a policy layer login view from the PR. When I open the `https://my.site/test/en/document/edit` url which requires loing I get the following at the `came_from` redirect:

* The `came_from` has value `'/test/en/docuemnt/edit'`.
* The `cpath` is `/site`.

Due to how `lstrip` works in python the result is:
```python
>>> '/test/en/docuemnt/edit'.lstrip('/site')
'n/docuemnt/edit'
```
The problem here is that `.lstrip()` doesn’t remove a prefix, it removes any characters from the start of the string that appear in the argument. This is definitely not what the goal of this PR.

Issue #2966, opened 2019-10-25
implement deletion to a recycle bin that allows users to restore items


tkimnguyen:
## Responsible Persons

Kim Nguyen

## Abstract

Allow users to delete and then undelete items themselves.

## Motivation

Because Plone needs to have this basic, expected content editing feature, and the old Undo is not available or reliable.

## Proposal & Implementation

TBD

## Deliverables

New feature in code. New docs materials. New training materials.

idgserpro:
Two years ago a discussion was set in https://community.plone.org/t/any-plans-for-an-official-trashcan-in-plone/5440 analyzing all available solutions. Matbe it's useful since it summarizes the implementation details of each solution.

tisto:
This would be a great feature to have in Plone. Though we implemented something like this in different forms (workflow, flags) in the past and I can not say that I was happy with any of the implementations we did.

Therefore, I would highly recommend implementing this as an add-on product first to gain an understanding of the pros and cons of the different approaches before getting this into core.

ebrehault:
@tkimnguyen the FWT approved this PLIP

sneridagh:
Just a reminder that the implementation should cover the endpoints to manage this as well.

tisto:
@tkimnguyen @pbauer not sure if you are aware but @rohnsha0 started to work on an implementation that looks very promising. @davisagli is already actively reviewing the PR and we plan to ship this with a future version of Plone. Are you ok with passing this PLIP along to @rohnsha0 and @davisagli?

tkimnguyen:
Indeed – thank you!

tisto:
@rohnsha0 I added you as the responsible person and proposer for this PLIP. Can you please add a link to your current work-in-progress (PRs)?

rohnsha0:
the PRs include:-

- [ ] https://github.com/plone/Products.CMFPlone/pull/4168
- [ ] https://github.com/plone/plone.base/pull/81
- [ ] https://github.com/plone/plone.app.discussion/pull/264
- [ ] https://github.com/plone/Products.CMFPlone/pull/4176
- [ ] https://github.com/plone/plone.restapi/pull/1920

@tisto thanks for adding me in. BTW above are the list of all the PRs related to recycle bin. the `restapi` PR is not yet ready, else other have been implemented with awaiting review changes (that I will do in coming days)

jensens:
Since I am curious about the state of this implementation I did an agent-supported review and summary of the current state.

[...]

PR #4166, opened 2025-04-27
Implement Recycle Bin

rohnsha0:
ref #2966 

image


rohnsha0:

rohnsha0:

stevepiercy (changes requested):
This is a great addition! Minor tweaks to messages and news. Thank you!

stevepiercy (approved):
Changes look good. Let's get a technical review from a senior Plonista.

petschki (changes requested):
This is nice, but I'd say this should go into master and 6.2 only as it is a new feature.

EDIT: maybe it could be refactored as a "core-addon" which could be used (optionally) in 6.1 too ?

pbauer:
This is an interesting solution to the problem that was not covered in https://community.plone.org/t/any-plans-for-an-official-trashcan-in-plone/5440 
I have not tested it yet but would like know where you found inspiration or discussed the implementation. I also wonder what stashing the obj in an annotation means for relations and annotations. I would expect that relations are removed and not restored upon restore. Are annotations (e.g. comments) kept in an annotation on the obj stashed in an annotation?

ale-rt:
All the presentation logic for classic UI (in particular the templates), should go in `plone.app.layout`.
The code should be structured in such a way that it is convenient to build some Rest API endpoint on it.

I still did not have the time to deeply checked the code.

About the approach taken, from my previous experiences, I like the approach of applying a trashed interface because it is way faster.
This does matter a lot when you have big sites.

Products/CMFPlone/browser/recyclebin.py
```diff
@@ -0,0 +1,206 @@
...
+
+class RecycleBinView(BrowserView):
+    """Browser view for recycle bin management"""
+
+    template = ViewPageTemplateFile("templates/recyclebin.pt")
+
+    def __call__(self):
+        form = self.request.form
+
+        if form.get("form.submitted", False):
```
ale-rt, line 21: I would use a z3c.form form here

PR #4175, opened 2025-05-14
[RECYCLE BIN] Fixes broken redirection after restoration and handle value-error gracefully if parent is deleted

### LINKED ISSUE #4171: Recycle bin: Can't restore item if it's original container no longer exists
This is a bug found while testing #4168 

Steps to reproduce:
1. Add a folder.
2. Add a page inside the folder.
3. Delete the page.
4. Delete the folder.
5. In the main recycle bin view, try to restore the page.

Expected: There should be a status message saying the item can't be restored because its original container no longer exists, and that you have to go to the individual item in the recycle bin to specify where to restore. (OR there should be a special form to immediately choose the new location.)
Actual: There's an error with a traceback:

```
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 181, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 390, in publish_module
  Module ZPublisher.WSGIPublisher, line 284, in publish
  Module ZPublisher.mapply, line 98, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module z3c.form.form, line 238, in __call__
  Module Products.CMFPlone.browser.recyclebin, line 39, in update
  Module z3c.form.form, line 233, in update
  Module plone.app.z3cform.csrf, line 21, in execute
  Module z3c.form.action, line 99, in execute
  Module z3c.form.button, line 312, in __call__
  Module z3c.form.button, line 167, in __call__
  Module Products.CMFPlone.browser.recyclebin, line 60, in handle_restore
  Module Products.CMFPlone.recyclebin, line 471, in restore_item
  Module Products.CMFPlone.recyclebin, line 394, in _find_target_container
ValueError: Original parent container at /Plone/folder no longer exists. You must specify a target_container to restore this item.
```

Next I thought I'll try to use the option to specify a target container on the recycle bin item view.

Steps:
1. Click to view the details for the page I deleted
2. Enter /Plone as the target container (try to restore with the top level of the Plone site as the container)
3. Click the "Restore item" button

Expected: The page is restored inside the Plone site, and the browser navigates to the restored page
Actual: There is a message 'Item 'page inside folder' successfully restored, but it wasn't restored. The browser navigated to the original path (/Plone/folder/page-inside-folder) which is a 404. The page is no longer in the recycle bin and has been lost.

rohnsha0:
fixes #4171

### DIFF
--- a/Products/CMFPlone/browser/recyclebin.py
+++ b/Products/CMFPlone/browser/recyclebin.py
@@ -567,8 +567,39 @@ class RecycleBinItemView(form.Form):
             )
             IStatusMessage(self.request).addStatusMessage(message, type="info")
 
-            redirect_url = f"{restored_obj.absolute_url()}/view"
-
+            # First try to use the restored object's URL
+            redirect_url = None
+            try:
+                object_url = f"{restored_obj.absolute_url()}/view"
+                
+                # Verify the URL is valid by checking if the path exists
+                site = getSite()
+                path = object_url.replace(site.absolute_url(), "").lstrip("/")
+                
+                try:
+                    # Try to traverse to the path to verify it exists
+                    site.unrestrictedTraverse(path)
+                    redirect_url = object_url
+                except (KeyError, AttributeError):
+                    # Path doesn't exist, don't set redirect_url yet
+                    pass
+            except (AttributeError, TypeError):
+                # If absolute_url fails, don't set redirect_url yet
+                pass
+                
+            # If the object URL doesn't work, try target container URL
+            if not redirect_url and target_container:
+                try:
+                    target_url = f"{target_container.absolute_url()}"
+                    redirect_url = target_url
+                except (AttributeError, TypeError):
+                    # If target container URL fails, don't set redirect_url yet
+                    pass
+                    
+            # If neither worked, fall back to recycle bin view
+            if not redirect_url:
+                redirect_url = f"{self.context.absolute_url()}/@@recyclebin"
+            
             self.request.response.redirect(redirect_url)
         else:
 
--- a/Products/CMFPlone/browser/recyclebin.py
+++ b/Products/CMFPlone/browser/recyclebin.py
@@ -14,12 +14,25 @@ from zope.component import getUtility
 from zope.i18n import translate
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
+from zope.component.hooks import getSite
 
 import logging
 import uuid
 
 
 logger = logging.getLogger(__name__)
+
+
+def _is_error_result(result):
+    """Helper method to check if a result is an error dictionary
+    
+    Args:
+        result: The result to check
+        
+    Returns:
+        Boolean indicating if the result is an error dictionary
+    """
+    return isinstance(result, dict) and not result.get("success", True)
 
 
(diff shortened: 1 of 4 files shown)


news/4171.bugfix
davisagli, line 1: Don't add release notes for a PR that is fixing issues in a new feature that hasn't been released yet.

Users only care about the feature that is added in the end, not the steps we took along the way to finish implementing it.

Products/CMFPlone/browser/recyclebin.py
```diff
@@ -338,6 +422,13 @@ def get_items(self):
                 if filter_type and item.get("type") != filter_type:
                     continue
 
+                # Check if parent container exists and add flag to the item
+                # Don't check for comments and comment trees which have special handling
+                if item.get("type") not in ("CommentTree", "Discussion Item"):
+                    item["parent_exists"] = self._check_parent_exists(item)
+                else:
+                    item["parent_exists"] = True  # Comments have special handling
```
davisagli, line 430: I would move this check for the type into the _check_parent_exists function

Products/CMFPlone/browser/recyclebin.py
```diff
@@ -669,6 +780,57 @@ def format_size(self, size_bytes):
         """Format size in bytes to human-readable format"""
         return human_readable_size(size_bytes)
 
+    def _get_redirect_url_after_restoration(self, restored_obj, target_container=None):
+        """Determine the appropriate URL to redirect to after item restoration.
+
+        This method implements a fallback strategy:
+        1. Try restored object's view URL first
+        2. If that fails, try target container URL
+        3. If that fails, use the recycle bin URL
```
davisagli, line 789: I don't understand the strategy. What is the situation where an item was restored but we can't redirect to its view URL?

davisagli (changes requested):

Products/CMFPlone/browser/recyclebin.py
rohnsha0: Umm, the implementation was thought inna way:-
- try to get the `{absolute_url()}/view` but if failing (maybe it is not available for the object)
- then get the target container url but if the worst case scenario - object is broken
- then target recyclebin url...

davisagli: > but if failing (maybe it is not available for the object)

This is the part I don't understand. Getting the object's absolute_url should always work. You can use restored_obj.restrictedTraverse("@@plone_context_state").view_url() to get it with the `/view` suffix for the content types that require that. (This comes from https://github.com/plone/plone.app.layout/blob/master/plone/app/layout/globals/context.py#L64)

davisagli (changes requested):
I'm still not seeing the restored object after I restore it to a different path on the item detail page. The restored object needs to be reindexed.

Products/CMFPlone/browser/recyclebin.py
rohnsha0: Maybe, I overcomplicated it too much... However, I have implemented a simplified based on ur suggestions in https://github.com/plone/Products.CMFPlone/pull/4175/commits/56428b176408451e30b9038268012c031311d48b

PR #4181, opened 2025-05-19
add login to the list of button names that show the primary button

### LINKED ISSUE #4180: Login form button has both btn-primary and btn-secondary classes
It looks like an error, because btn-primary is injected when building the form, when this should be placed in plone.app.z3cform.

Pr is coming

erral:
Fixes #4180 

Test together: https://github.com/plone/plone.app.z3cform/pull/234


petschki (approved):
LGTM. Could you do a second one for `master` too?

PR #4186, opened 2025-05-29
Make resource registry more robust against broken resources.

thet:
Don't break the resource registry when a resource error happens (missing dependency, circular dependency, file not found, etc). Admins will see a warning badge and can fix the problem in the resource registry user interface. Previously such errors broke the rendering of the whole site, making fixes very fiddly.


Products/CMFPlone/resources/browser/resource.py
```diff
@@ -116,25 +117,32 @@ def check_dependencies(bundle_name, depends, bundles):
...
+                )
 
                 if name in GRACEFUL_DEPENDENCY_REWRITE:
                     # gracefully rewrite old bundle names
                     graceful_depends = GRACEFUL_DEPENDENCY_REWRITE[name]
                     logger.error(
-                        msg
-                        + f"Bundle dependency graceful rewritten to '{graceful_depends}' "
-                        + "Fallback will be removed in Plone 7."
+                        "%r Bundle dependency graceful rewritten to %r."
```
stevepiercy, line 132:
```suggestion
                        "%r Bundle dependency gracefully rewritten to %r."
```

stevepiercy (changes requested):
I caught one more possible grammar issue. I'm not really sure because the original sentence doesn't quite make sense to me.

Products/CMFPlone/resources/browser/resource.py
thet: I even removed the graceful/ly.

stevepiercy (approved):

thet:

PR #4188, opened 2025-06-03
Automatically set the ajax_load request parameter.

thet:
Related:
https://github.com/plone/Products.CMFPlone/pull/4188
https://github.com/plone/plone.base/pull/85
https://github.com/plone/plonetheme.barceloneta/pull/404
https://github.com/plone/plone.app.portlets/pull/203

Plone 6.1: https://github.com/plone/Products.CMFPlone/pull/4188  (this one)
Plone 6.2: https://github.com/plone/Products.CMFPlone/pull/4169

Zope maintains the "HTTP_X_REQUESTED_WITH" request header. If this is set to "XMLHttpRequest", we have an AJAX request. In this case the ajax_load parameter is set to 1 directly on the request object.

This should make use for the ajax_main_template for any AJAX request, regardless if the ajax_load parameter was manually set or not and potentially speed up Plone Classic UI rendering.

### DIFF
--- a/Products/CMFPlone/browser/templates/ajax_main_template.pt
+++ b/Products/CMFPlone/browser/templates/ajax_main_template.pt
@@ -16,7 +16,7 @@
           portal_url python:portal_state.portal_url();
           checkPermission python:context.restrictedTraverse('portal_membership').checkPermission;
           ajax_include_head python:request.get('ajax_include_head', False);
-          ajax_load python:False;"
+          ajax_load python:True;"
       i18n:domain="plone"
       tal:attributes="lang lang;">
 
--- a/Products/CMFPlone/traversal.py
+++ b/Products/CMFPlone/traversal.py
@@ -10,9 +10,6 @@ from zope.interface import implementer
 from zope.pagetemplate import engine as zpt_engine
 
 
-MARKER = object()
-
-
 class PloneBundlesTraverser(ResourceTraverser):
     # the name is misleading - it is used not only for bundles.
     # in fact in Plone 6 bundles are no longer used, despite that the traverser
@@ -85,7 +82,7 @@ def set_ajax(obj, event):
     # it again.
     if (
         request.getHeader("HTTP_X_REQUESTED_WITH") == "XMLHttpRequest"
-        and not request.get("ajax_load", MARKER) is not MARKER
+        and "ajax_load" not in request
     ):
         # Directly set on the request object
         request.set("ajax_load", True)
--- a/Products/CMFPlone/traversal.py
+++ b/Products/CMFPlone/traversal.py
@@ -10,6 +10,9 @@ from zope.interface import implementer
 from zope.pagetemplate import engine as zpt_engine
 
 
+MARKER = object()
+
+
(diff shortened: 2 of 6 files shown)


thet:

mauritsvanrees:
The related PRs seem fine, and could be merged and used in 6.1, even without the current PR, in my opinion.

But I have a doubt about this one for Plone 6.1.  You already saw some unexpected problem in `plone.app.portlets` caused by this.  I wonder which other problems lurk in custom code where someone may do an XMLHTTPRequest from Javascript and expect this to contain the full page.  If this assumption breaks in 6.2, I would say it is not so bad, and can be fixed by appending `?ajax_load=0` to the url.  But for a 6.1 bug fix release I am not so sure.

What do others think?

Products/CMFPlone/browser/templates/ajax_main_template.pt
```diff
@@ -16,7 +16,7 @@
           portal_url python:portal_state.portal_url();
           checkPermission python:context.restrictedTraverse('portal_membership').checkPermission;
           ajax_include_head python:request.get('ajax_include_head', False);
-          ajax_load python:False;"
+          ajax_load python:True;"
```
ale-rt, line 19: I have no idea why this `ajax_main_template` has `ajax_load=False` in the first place. This variable seems really obscure defined like this.

Products/CMFPlone/traversal.py
```diff
@@ -72,3 +75,17 @@ def get_zope_page_template_engine(engine):
...
+    request = event.request
+
+    # If ajax_load was already set to a true-ish or false-ish value, don't set
+    # it again.
+    if (
+        request.getHeader("HTTP_X_REQUESTED_WITH") == "XMLHttpRequest"
+        and not request.get("ajax_load", MARKER) is not MARKER
+    ):
+        # Directly set on the request object
+        request.set("ajax_load", True)
```
ale-rt, line 91: I do not like this approach.
You are checking the request, fiddling with the request parameters, and then checking the request again.

IMO this is too complicated.
There should be some come somewhere that has a method `is_ajax_load` (or whatever other name) that is the unique source of truth about the fact that we are loading stuff through AJAX.

Ideally, this should be a method in a view (`plone_context_state`?) so that it is easily configurable (e.g. you might want to always load some portal type as if it was called as AJAX, add a check for the referrer or whatever).

BTW even if you want to have this traverser, the second condition has to be simplified because it contains a double negation.

Something like:
```python
request.get("ajax_load", MARKER) is MARKER
```
or even:
```python
"ajax_load" not in request
```
is preferable.

thet: LOL, `and not request.get("ajax_load", MARKER) is not MARKER` .... yes, i'll take your last suggestion - this is much better :dagger:

thet:
@ale-rt I have addressed your comments in https://github.com/plone/Products.CMFPlone/pull/4169

I agree with @mauritsvanrees this should probably not be merged into Plone 6.1 but 6.2 instead.

Issue #2666, opened 2018-12-14
A simple PT can kill the Zope/WSGI server


zopyx:
Plone 5.2a1, Python 3.6.5.

By accident I had a condition `tal:condition="view"` in one of my templates.

This causes the complete Plone instance to shut down and exit (instead of raising a RecursionError exception).

Haven't verified if this is an issue with Zope 4 and/or Python 2.7 vs. 3.x.

```
2018-12-14 11:31:41,825 INFO  [Zope:44][MainThread] Ready to handle requests
Starting server in PID 14831.
Serving on http://0.0.0.0:8080
2018-12-14 11:32:24,117 DEBUG [plone.app.theming.policy:149][waitress] local cache invalidated on thread 140065861908224
2018-12-14 11:32:24,117 DEBUG [plone.app.theming.policy:112][waitress] initializing local cache on thread 140065861908224 for http://dev.zopyx.de:8080/Plone2::barceloneta
2018-12-14 11:32:35,100 ERROR [plone.app.viewletmanager:118][waitress] Error while rendering viewlet-manager=plone.toolbar, viewlet=plone.contentviews
Traceback (most recent call last):
  File "/home/ajung/src/xmldirector.connector/eggs/Chameleon-3.4-py3.7.egg/chameleon/template.py", line 183, in render
    self._render(stream, econtext, rcontext)
  File "6c8ffcea73d20e781fc49bce59d6f5e4.py", line 109, in render
  File "/home/ajung/src/xmldirector.connector/eggs/Zope-4.0b7-py3.7.egg/Products/PageTemplates/expression.py", line 105, in __call__
    return base()
  File "/home/ajung/src/xmldirector.connector/eggs/plone.app.contentmenu-2.2.4-py3.7.egg/plone/app/contentmenu/view.py", line 27, in menu
    items = menu.getMenuItems(self.context, self.request)
  File "/home/ajung/src/xmldirector.connector/eggs/zope.browsermenu-4.4-py3.7.egg/zope/browsermenu/menu.py", line 52, in getMenuItems
    if item.available():
  File "/home/ajung/src/xmldirector.connector/eggs/plone.memoize-1.2.3-py3.7.egg/plone/memoize/instance.py", line 53, in memogetter
    val = func(*args, **kwargs)
  File "/home/ajung/src/xmldirector.connector/eggs/plone.app.contentmenu-2.2.4-py3.7.egg/plone/app/contentmenu/menu.py", line 82, in available
[...]

PR #4195, opened 2025-06-20
Drop workaround for #2666

pbauer:
Drop workaround for #2666 - a bug in Python (bugs.python.org/issue46110) which was fixed in 3.9.10.


pbauer:

mauritsvanrees:
I wonder how safe this is to do in a 6.1 bug fix release, if people have overridden this.  I know I have overridden this view I several projects, because it is the only way to override the main template.  

In my case I have probably always copied the entire class, so that will still work.  But if someone has made a monkey patch to set `main_template_name` to something else, that will no longer have an effect.

mauritsvanrees:
The failing Jenkins test should be fixed if you do a rebase.

pbauer:
I don't really mind leaving that for 6.2 if you prefer that, I've only ever overwritten the whole class.

davisagli:
I'll close the PR for Plone 6.1 then.

PR #4225, opened 2025-09-24
Make referencing IDs in inline rendered SVGs unique

### LINKED ISSUE #4224: Broken SVG display
The `@@iconresolver` view removes the `id` attribute from `<svg>` tags, but this attribute is needed for displaying repeatable elements with `<use>` ('href' and (deprecated) 'xlink:href').

The error is visible, when you're activating the brazilian flag in the language selector.

Compared here: https://jsfiddle.net/petschki/d0egph7y/11/

petschki:
fixes #4224

### DIFF
--- a/Products/CMFPlone/browser/icons.py
+++ b/Products/CMFPlone/browser/icons.py
@@ -1,4 +1,3 @@
-from collections import defaultdict
 from lxml import etree
 from OFS.Image import File
 from plone.registry.interfaces import IRegistry
@@ -9,8 +8,8 @@ from zope.component.hooks import getSite
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
 
-import hashlib
 import logging
+import uuid
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


petschki:

jensens:
IIRC correctly, IDs were stripped to avoid duplicated IDs in DOM, which lead to failures in W3C site validation.

yurj:
Can't you just can replace/prepend them with a random string?

petschki:
Yeah that also came into my mind. Let me try something this weekend.

petschki:
I've refactored this to make the referencing IDs unique. I use a hash of `time.time()` so that when the same SVG is placed multiple times on a page every instance has its unique ID ... I've tested this with the countryflags, because they have many references inside and are placed inline in the languageselector. 

I've also fixed the `aria-labelledby` attribute, when you provide a `tag_alt` parameter. This needs to be an id of the generated `<title>` tag. This came up in an accessibility check ...

let me know what you think.

petschki:

petschki:
Also worth noting: the `@@iconresolver.tag` factory could need some caching, because this is called every time when rendering a page ... I've locally tried to decorate it with `@view.memoize_contextless` but that didn't work ...

petschki:

petschki:

petschki:

PR #4232, opened 2025-10-15
Filter out REST API control panels from redirection checks in absolutize_path

### LINKED ISSUE #4231: Redirect cannot be created if it contains socialmedia at the end.
### What I did:

* I go to any page
* I go to Redirects (@@manage-aliases)
* I enter any path with socialmedia at the end (e.g., /news/socialmedia)

### What I expect to happen:
 
* The redirect is created

### What actually happened:

* The redirect cannot be created because it supposedly already exists, which it does not.

https://github.com/plone/Products.CMFPlone/blob/aa911951c4d951a2fc990cffb9c68f5979d80587/src/Products/CMFPlone/controlpanel/browser/redirects.py#L89

* The reason is that `portal.unrestrictedTraverse(path, None)` incorrectly does not lead to `None` 
* In fact, the result is  <plone.restapi.controlpanels.registry.SocialMediaControlpanel object at 0x7eef874cbcd0>

* 6.1.1 (Classic-UI)

rohnsha0:
fixes #4231

### DIFF
--- a/news/4231.bugfix
+++ b/news/4231.bugfix
@@ -1,0 +1,1 @@
+Filter out REST API control panels from redirection checks in absolutize_path @rohnsha0


davisagli:
We already fixed this a different way in https://github.com/plone/plone.restapi/pull/1909

PR #4237, opened 2025-11-13
Fix broken CSS on 404 pages inside private folders

dobri1408:
There is a bug right now in plone:

There is currently a bug in Plone:

When accessing
https://classic.demo.plone.org/en/demo/not-found, the site correctly returns a 404 page:
Screenshot 2025-11-13 at 16 47 31

However, when accessing https://classic.demo.plone.org/en/sandbox/not-found, the CSS is broken and the page renders incorrectly:
Screenshot 2025-11-13 at 16 47 56

It seems this issue appears because “sandbox” is a private folder, and the 404 page is trying to load resources that are not publicly accessible. I attempted a fix that works, but since I’m not primarily a Plone Python developer, someone more experienced may need to review and refine it. Still, this likely requires attention, as the issue affects the default error handling.

### DIFF
--- a/news/4269.bugfix.md
+++ b/news/4269.bugfix.md
@@ -1,0 +1,1 @@
+Do not use the remote url of a link item in navigation.  [maurits]
--- a/src/Products/CMFPlone/browser/navtree.py
+++ b/src/Products/CMFPlone/browser/navtree.py
@@ -139,12 +139,6 @@ class SitemapNavtreeStrategy(NavtreeStrategyBase):
         if portalType is not None and portalType in self.viewActionTypes:
             itemUrl += "/view"
 
-        useRemoteUrl = False
-        getRemoteUrl = getattr(item, "getRemoteUrl", None)
-        isCreator = self.memberId == getattr(item, "Creator", None)
-        if getRemoteUrl and not isCreator:
-            useRemoteUrl = True
-
         isFolderish = getattr(item, "is_folderish", None)
         showChildren = False
         if isFolderish and (portalType is None or portalType not in self.parentTypesNQ):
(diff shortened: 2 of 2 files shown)


dobri1408:

yurj:
Hi! Just to ask: the plone root object should already be cached somewhere in a variable, the API plone.api.portal.get() uses getSite() which usually is locally cached in the Zope thread. But maybe I'm missing some details. 

See this discussion: https://community.plone.org/t/getting-the-portal-easy-and-quick/15493

UPDATE: the code search for the navroot, not the portal. The api for getting the navigatio root uses the same method but have some shortcut if plone.root is set in the registry, and basically uses get_navigation_root_object as suggested by @ale-rt 

I would use the api anyway.

src/Products/CMFPlone/browser/exceptions.py
```diff
@@ -30,7 +30,15 @@ def __call__(self):
...
+        # Always use portal root as context to ensure resources load correctly
+        current = self.__parent__
+        while hasattr(current, "__parent__") and current.__parent__ is not None:
+            if hasattr(current, "meta_type") and current.meta_type == "Plone Site":
+                break
+            current = current.__parent__
+
+        self.context = current
+        self.__parent__ = current  # Also update __parent__ to ensure consistency
         request = self.request
```
ale-rt: This might cause issues on some set up.

If I understand it correctly, you want to get the **navigation root**.

I would check if it works with:
```
from plone.base.navigationroot import get_navigation_root_object
```

In addition, this might break the suggestion:
https://github.com/plone/Products.CMFPlone/blob/2122f8bfe0cb46b8d21f652f3f2c0f0e24856274/src/Products/CMFPlone/browser/templates/error_message.pt#L43-L44

Most probably the issue lies elsewhere.
Gut feeling: I would check the styles viewlet.

davisagli:
I don't really think this fix is safe...I'm sure some sites are relying on being able to get the actual context where the exception occurred.

Let's go back and look at the actual error. @dobri1408 what are the style URLs that it's trying to load which give a 404?

dobri1408:
Moved here https://github.com/plone/Products.CMFPlone/pull/4293

Issue #3928, opened 2024-03-20
PLIP: Switch to PEP 420 (Implicit namespace packages)


mauritsvanrees:
## Abstract

Move from `setuptools/pkg_resources` namespaces to native namespaces.
This is a breaking change that we need for Plone 6.2, because `setuptools` plans to drop its `pkg_resources` module near the end of 2025.

## Motivation

Use of `pkg_resources` from `setuptools` is [deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html).  Native namespaces have been available since Python 3.3 and are now the recommended and most easy way to define namespaces.

## Assumptions

* Zope wants to [make the same move](https://github.com/zopefoundation/meta/issues/194).
Since it is best to move all packages within one namespace at the same time, we need to coordinate with Zope for the `Products` namespace.
* Support for non-native namespaces is slowly being removed from `pip` and `setuptools` and other packaging related software.  Bugs may be less likely to get a fix.
* We want to be less dependent on `setuptools`.  There are other ways to handle packaging and distribution nowadays.
* Theoretically you can combine native namespaces and `pkg_resources` namespaces within the same namespace, but this depends on the tools that you use, and whether you make a development checkout of a package.  A combination that works today, might fail tomorrow with a newer `setuptools` or `pip` version.  It is best to move over all packages from a single namespace at the same time.
* Note that there should be no problem combining different namespace variants in different namespaces: if all `zope.*` packages have moved to non-native namespaces, all `plone.*` packages could still use the old `pkg_resources`-style namespaces.

## Proposal & Implementation

* In Plone 6.2, all repositories in the GitHub `plone` organisation that are used by the `Plone` package should move to native namespaces.  Others too where possible, and where we have control.
* In the upgrade guide, give add-on authors the advice to do the same.  This goes especially for the `collective` namespace.
* The main changes per package:
  * Make a maintenance branch for this package on Plone 6.1.  We want to leave room for new features in Plone 6.1, so the branch should be something like `3.x`, not `3.0.x`.
  * Remove `__init__.py` from the namespace packages.  In `Products.CMFPlone` we would remove `Products/__init__.py`.
[...]

PR #4239, opened 2025-11-18
PEP 420 native namespace

gforcada:
Part of plone/Products.CMFPlone#3928

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,7 +118,6 @@ setup(
         "Products.SiteErrorLog",
         "Products.statusmessages",
         "Products.ZCatalog",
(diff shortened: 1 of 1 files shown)


setup.py
davisagli: Can we remove setuptools from install_requires?

gforcada: 🤔 interesting... Maurits made the changes on `plone.meta` and I made a release of it, so I was expecting that `uvx` would pick up the new version automatically...

I will fix all my PRs, sorry! 🙏🏾

mauritsvanrees: I have removed setuptools.

PR #4247, opened 2025-12-17
TinyMCE license key

petschki:
See https://github.com/plone/plone.base/pull/104

### DIFF
--- a/news/+tinymce_license_key.feature.md
+++ b/news/+tinymce_license_key.feature.md
@@ -1,0 +1,1 @@
+Add configurable license key to TinyMCE options.  @petschki


petschki:
Jenkins should go green. GHA cannot, because of missing `plone.base` PR dependency.

Is there a plan how to test multiple PRs with GHA?

davisagli:
@petschki For now we don't have a solution for that.

Basically we would need something that looks up the checkouts and sources in buildout.coredev and installs them the same way in the GHA setup.

There was also some brainstorming done at ploneconf 2022 about creating a coredev Docker image that could be used for this, but no one followed up to implement it. https://docs.google.com/document/d/1lB22QrynZdrXP50UVObIyWTv3ykyzU9WZMi__0_n6_w/edit?usp=sharing

petschki:
> Basically we would need something that looks up the checkouts and sources in buildout.coredev and installs them the same way in the GHA setup.

Well using (mx)checkouts is not enough if you have unmerged but depending PRs. For this special case we should be able to provide "depending PR urls" for the GHA ... googling for that I found this https://github.com/depends-on/depends-on-action which sounds like it could solve our problem ... 

/cc @gforcada

mauritsvanrees:
GHA will be broken for a bit, but I am making releases to fix that.

mauritsvanrees:
I have make releases of `plone.base` and a few others, and have updated dist.plone.org.  I have restarted the GHA tests, they should end up green again.

mauritsvanrees:
Yes, GHA is green again:
https://github.com/plone/Products.CMFPlone/actions/runs/20334504041

gforcada:
@petschki that sounds interesting indeed, we should check it. Right now CI wise, I'm more worried that we have a small, but growing (and we want), group of core packages that use pytest rather than zope.testrunner that we are only testing in GHA and not in Jenkins, so if any of those packages break any of the ones that use zope.testrunner we have no idea, and likewise the other way around 🤷🏾

PR #4257, opened 2026-01-22
Fix classic toolbar scrolling on small height screens

### LINKED ISSUE #3908: Plone classic side toolbar has no scroll on small height screens
In Plone 5, on smaller screens the toolbar would be given a fixed height, overflow: auto so that it could be scrolled, and it had <li> buttons to assist.

In Plone 6 classic we do not have that. It looks like the toolbar will collapse to 'mobile' mode if the width is below a threshold, but not the height. Also if more actions are added to the toolbar it is not possible to reach them in mobile mode.

Example 1: wide, short screen. No way to reach elements below Display
image

Example 2: Adding another action in classic.demo.plone.org pushes user actions below the screen (beneath the sharing icon)
image

I used https://classic.demo.plone.org/en to test, Plone 6.0.9

shivaansh0610-LUFFY:
Fixes #3908

This PR ensures that the classic Plone left toolbar becomes scrollable
When the viewport height is small, preventing toolbar actions from being
cut off on short screens.

### DIFF
--- a/news/3908.bugfix
+++ b/news/3908.bugfix
@@ -1,0 +1,2 @@
+Fix classic toolbar scrolling on small height screens.
+


shivaansh0610-LUFFY:

petschki:
If you want to implement fixes for the Classic UI Toolbar you have to look here: https://github.com/plone/plonetheme.barceloneta/blob/master/scss/_toolbar.scss

shivaansh0610-LUFFY:
@petschki 
Thanks for helping me and pointing this out! 
I’ll move the fix to the Barceloneta toolbar styles and update the PR accordingly.

Issue #2161, opened 2017-09-26
Zope-level language negotiation


malthe:
The `z3c.pt` package relies on Zope's language negotiation utility `zope.i18n.INegotiator` but `plone.i18n` does not register such a utility, instead relying on Zope's own negotiation logic which is not the same as `plone.i18n`.

The result is that different languages may be used on the same page.

jensens:
would be an enhancement of ``plone.i18n`` - PRs welcome.

jensens:
May still be relevant for Plone 6. plone.i18n still does not register a zope.i18n.INegotiator utility, causing z3c.pt to use different language negotiation. Needs verification.

PR #4268, opened 2026-01-31
Register plone.i18n negotiator as Zope INegotiator utility

shivaansh0610-LUFFY:
Contributes to  #2161

### What I did
- Registered `plone.i18n.negotiate.ptsnegotiator.Negotiator` as the
  global ZCA utility for `zope.i18n.interfaces.INegotiator.`
- Added a regression test to verify that Plone registers the negotiator
  instead of falling back on Zope defaults

### Why
CMFPlone did not register an `INegotiator` utility before, which meant
That Zope language negotiation skipped Plone’s i18n settings.

### Tests
- Added `test_plone_registers_i18n_negotiator`.

### DIFF
--- a/news/2161.bugfix
+++ b/news/2161.bugfix
@@ -1,0 +1,2 @@
+Register plone.i18n negotiator as the Zope INegotiator utility.
+


shivaansh0610-LUFFY:

davisagli (commented):
I don't know how to review this. We need an example of where it makes a difference. I assume that since the issue has been open for 9 years, it might not actually be a problem in practice.

shivaansh0610-LUFFY:
@davisagli 

Thanks for taking a look!

The reason for this change is more about consistency and making the wiring explicit, rather than solving a very obvious bug.

One example where this might actually make a difference is when Zope-level language negotiation is performed (e.g., during template rendering or early request handling) before Plone language settings are fully in effect. In such a case, Zope might resort to its default negotiator rather than Plone’s i18n logic, potentially causing inconsistent language choices throughout a request.

This PR makes the wiring explicit by registering Plone’s existing `plone.i18n` negotiator as the global `INegotiator` utility, ensuring that Zope and Plone use the same negotiation strategy consistently.

I do agree that this might not be an obvious bug to appear in most environments, which might be why this issue has been open for such a long time. The goal here is to make the configuration more correct and predictable with minimal risk, as it’s based on Plone’s own negotiator implementation.

If you feel that this should be documented in a different way, protected, or scoped differently, I’m more than happy to follow your lead.

shivaansh0610-LUFFY:
@davisagli I took a look at the Jenkins failures, and they seem unrelated to this PR.

The failing job shows a huge number of failures (176) across many unrelated
modules (CMFDiffTool, CMFEditions, control panel browser tests, etc.), which look
more like a Jenkins / test environment issue than something caused by registering
The INegotiator utility.

Happy to re-run Jenkins or adjust if you see anything specific related to this change.


Screenshot 2026-02-01 at 1 50 05 AM
Screenshot 2026-02-01 at 1 57 06 AM

davisagli:
@shivaansh0610-LUFFY At first glance, the errors look like a direct consequence of registering this utility. The first one I looked at is failing on a call to Negotiator.getLanguage()

I don't think anyone has a good understanding of the side effects of registering this utility, and we shouldn't change anything unless someone can make a _specific_ report of a problem, instead of a theoretical one.

shivaansh0610-LUFFY:
Thanks for the detailed explanation and for taking a close look.

That makes sense — without a concrete bug report and with observable side effects,
it’s reasonable not to change this behavior. I appreciate the review and guidance.

Issue #4269, opened 2026-02-02
Remote urls in navigation portlet


mauritsvanrees:
Using Plone 6.1/6.2.

In a Classic UI site, create a folder, add a Link CT in it, and set this to an internal url.  Publish the folder and the link.

* In the main menu, assuming the navigation is using a dropdown, you will see as link `http://localhost:8080/Plone/folder/link`.  It does not matter if you are logged in or anonymous.
* In the navigation portlet it depends:
  * If you are logged in and you are the creator of the link, you will see a link to `http://localhost:8080/Plone/folder/link` so you can click and edit it.
  * If you are anonymous, or you are logged in but you are *not* the creator of the link, you will see something like this: `http://localhost:8080/Plone/resolveuid/some-uid`.

Problems:

1. There is a difference between the main menu and the navigation portlet.
2. If you have edit permission on the link, but you are not the original creator, you have to resort to tricks to get to the link object to edit it.
3. It is ugly to literally have `resolveuid` in the link.
4. On a live site, the url may include the id of the Plone site: `http://example.com/Plone/resolveuid/some-uid`.  You get this when your editors can't login on the normal domain, but directly access the Plone port in a VPN.  Best seen with ssh port forwarding.  The link created above would have as `getRemoteUrl` in its brain: `/Plone/resolveuid/...`. If you would edit it on the live site domain, it would be `/resolveuid/...`.  (In my case the extra `/Plone` in the url leads to 404 Not Found errors because something in the Apache rules seems to disallow access to `example.com/Plone`.  This is the main reason I was investigating this.)

Easiest way to solve this, would be to change the `SitemapNavtreeStrategy` in `Products.CMFPlone` to always set `useRemoteUrl` to False.  Then the navigation portlet will simply use the url of the link object.  I have the code ready.

This seems a minor change.  Since this *is* a change of behaviour, perhaps it is best to only do this in Plone 6.2.

BTW, the code lines that I am editing are largely unchanged since their introduction in 2011, in the [fifth ever PR](https://github.com/plone/Products.CMFPlone/pull/5) to CMFPlone.

mauritsvanrees:
I test these PRs together on Jenkins:

* https://github.com/plone/Products.CMFPlone/pull/4270
* https://github.com/plone/plone.app.portlets/pull/213

PR #4270, opened 2026-02-02
Do not use the remote url of a link item in navigation.

mauritsvanrees:
Main reasons:

* The main navigation menu does not use the remote url either.  Let's keep these two in sync.
* Until now, only the creator would get the url of the link item itself.  So other editors would have a hard time traversing to the link to edit it.

For more, see https://github.com/plone/Products.CMFPlone/issues/4269


jensens:

Issue #4183, opened 2025-05-19
ignoreRequest produces an error for RichText Form Widget


Nimo-19:
After updating my forms server form plone 5 to 6, I discovered an error with the z3c_form patch and the richtext form widget: **plone.app.z3cform.widgets.richtext**

When I visit the form through a link from another site, I get the error below..
When I type the link in the url bar the form renders successfully.

Somehow the form seems to not be set on the RichText Widget.
I speculate it has something to do with this "New implementation of pattern widget with z3c.form extendable attributes" ([See plone.app.z3cform Commit b65fc7a](https://github.com/plone/plone.app.z3cform/commit/b65fc7a9bb2aec99875c00754f671192ffba96b1))

### Stacktrace:
```
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 181, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 390, in publish_module
  Module ZPublisher.WSGIPublisher, line 284, in publish
  Module ZPublisher.mapply, line 98, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module plone.z3cform.layout, line 61, in __call__
  Module plone.z3cform.layout, line 55, in update
  Module z3c.form.form, line 168, in render
  Module Products.Five.browser.pagetemplatefile, line 127, in __call__
  Module Products.Five.browser.pagetemplatefile, line 59, in __call__
  Module zope.pagetemplate.pagetemplate, line 134, in pt_render
  Module Products.PageTemplates.engine, line 368, in __call__
  Module z3c.pt.pagetemplate, line 198, in render
  Module chameleon.zpt.template, line 328, in render
  Module chameleon.template, line 200, in render
  Module b4855b310acae6edadbf82fdae853e4e, line 268, in render
  Module 29bd1ef10dfd653097a000a51c8a1633, line 177, in render_form
  Module 29bd1ef10dfd653097a000a51c8a1633, line 943, in render_titlelessform
  Module 29bd1ef10dfd653097a000a51c8a1633, line 1258, in render_fields
  Module 29bd1ef10dfd653097a000a51c8a1633, line 1799, in render_widget_rendering
  Module 29bd1ef10dfd653097a000a51c8a1633, line 1902, in render_field
  Module zope.tales.expressions, line 248, in __call__
  Module Products.PageTemplates.Expressions, line 225, in _eval
  Module Products.PageTemplates.Expressions, line 155, in render
[...]

PR #4271, opened 2026-02-02
Fix sharing modificationdate

shivaansh0610-LUFFY:
### What this does
Changing sharing / local role inheritance (`@@sharing`) should update
security indexes, but **must not change `ModificationDate`**.

This PR ensures that:
- `allowedRolesAndUsers` is reindexed
- `ModificationDate` remains unchanged

### Why
Sharing changes are permission changes, not content edits.
Updating `ModificationDate` causes:
- wrong UI signals
- unnecessary reindexing
- confusing audit history

### Tests
Added a regression test that proves:
- permissions change
- `ModificationDate` stays the same

### How to test
1. Create a document
2. Toggle sharing / inheritance
3. Verify the modification date is unchanged

Contributes to #4183

### DIFF
--- a/src/Products/CMFPlone/tests/testContentSecurity.py
+++ b/src/Products/CMFPlone/tests/testContentSecurity.py
@@ -89,7 +89,7 @@ class TestContentSecurity(PloneTestCase):
         # This should now raise ValueError
         with self.assertRaises(zUnauthorized):
             folder.subfolder.invokeFactory("Document", "new")
-            
+
     def test_sharing_does_not_change_modification_date(self):
         self.login("user1")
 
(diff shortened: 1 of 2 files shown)


shivaansh0610-LUFFY:

davisagli:
You've included unrelated changes in the PR.

Also, the test is invalid. The modification date is usually updated during catalog reindexing, which is deferred, and the test did not do what is necessary to make it happen.

shivaansh0610-LUFFY:
Thanks for the review and for the clear explanation.

You're right — this PR includes unrelated changes, and the test doesn't account for deferred catalog reindexing, which can update the modification date later. I misunderstood the lifecycle involved here.

I’ll close this PR, clean up the branch, and rework the test properly (or open a fresh PR) once I have a correct, minimal reproduction that triggers the full indexing path.

Thanks again for pointing this out — much appreciated.

davisagli:
@shivaansh0610-LUFFY I'd suggest checking whether you can reproduce the problem manually. If it doesn't work manually but the test passes, something is wrong with the test.

PR #4276, opened 2026-02-14
Skip redundant _reindexWorkflowVariables in notifyCreated

### LINKED ISSUE #4275: Redundant catalog indexing during content creation: notifyCreated calls _reindexWorkflowVariables unnecessarily
(this was a finding while creating zodb-pgjsonb)

## Problem

When content is created, `handleContentishEvent` (CMFCatalogAware.py:272-276) does:

```python
if IObjectAddedEvent.providedBy(event):
    wfaware.notifyWorkflowCreated()   # (a) initializes workflow + redundant reindex
    ob.indexObject()                    # (b) full INDEX (queued)
```

Step (a) calls `WorkflowTool.notifyCreated` which calls `_reindexWorkflowVariables(ob)`. This method does two things:

1. `ob.reindexObject(idxs=vars)` — queues a REINDEX for workflow variables (redundant, because `indexObject()` in step (b) covers all indexes)
2. `ob.reindexObjectSecurity()` — calls `unrestrictedSearchResults(path=path)` which forces an **immediate `processQueue()`**, flushing all pending indexing operations

### Why this is expensive

cProfile data from creating 100 content objects (50 Documents + 50 News Items):

- **Total ADD time: 8.31s** (83ms per object)
- **`reindexObjectSecurity` path: ~4.1s (49%)** — this includes the premature queue flush via `unrestrictedSearchResults` → `processQueue()` and the resulting catalog operations
- **`_reindexWorkflowVariables` overall: ~4.0s** — `reindexObject(idxs=vars)` queues a REINDEX that the queue optimizer later has to merge with the INDEX from `indexObject()`

### Why it is redundant for new objects

- `reindexObjectSecurity` searches the catalog for child objects to cascade security changes to. A brand-new object has **no children** — the search always returns 0 results (or just the object itself, which hasn't been indexed yet anyway).
- `reindexObject(idxs=vars)` queues a REINDEX for workflow variable indexes (`review_state`, etc.), but `indexObject()` called immediately after in `handleContentishEvent` indexes ALL fields including those same workflow variables.
- The `wf.notifyCreated(ob)` call that initializes the workflow state (creates initial history, sets review_state attribute on the object) is NOT the problem — that part is needed. Only the subsequent `_reindexWorkflowVariables` call is redundant.

### Callers of `_reindexWorkflowVariables`

There are exactly two callers in `Products.CMFCore.WorkflowTool`:
1. `notifyCreated` (line 296) — during object creation ← **this is the redundant one**
2. `_invokeWithNotification` (line 538) — during workflow transitions ← this one is needed

## Proposed Fix

Override `notifyCreated` in CMFPlone's `WorkflowTool` subclass (`Products/CMFPlone/WorkflowTool.py`) to skip the `_reindexWorkflowVariables` call:

```python
@security.private
def notifyCreated(self, ob):
    """Notify all applicable workflows that an object has been created.

    Overrides CMFCore to skip _reindexWorkflowVariables.
    The caller (handleContentishEvent) always calls indexObject()
    immediately after, which indexes all fields including workflow
    variables and security. Calling _reindexWorkflowVariables here
    is redundant and expensive: reindexObjectSecurity forces an
    immediate queue flush via unrestrictedSearchResults for an object
    that has no children yet.
    """
    wfs = self.getWorkflowsFor(ob)
    for wf in wfs:
        if self.getHistoryOf(wf.getId(), ob):
            continue
        wf.notifyCreated(ob)
```

This preserves workflow state initialization while eliminating the redundant reindex. Workflow transitions continue to use `_reindexWorkflowVariables` via `_invokeWithNotification` on the base class, so those are unaffected.

## Expected Impact

- **Content creation**: ~49% faster (~4.1s saved per 100 objects, from 83ms to ~42ms per object)
- **Workflow transitions**: No change
- **Content deletion**: No change

## Risk Assessment

Low risk:
- Only the creation path is affected
- The CMFCore test `test_object_indexed_after_adding` already expects exactly one index call (`['index /site/bar']`)
- Workflow state initialization is preserved
- The fix should also be upstreamed to `Products.CMFCore.WorkflowTool.notifyCreated` in the long term

jensens:
## Summary

- Override `notifyCreated` in Plone's `WorkflowTool` to skip the redundant `_reindexWorkflowVariables()` call
- Workflow state initialization (`wf.notifyCreated(ob)`) is preserved; only the redundant reindex is removed

## Problem

During content creation, `handleContentishEvent` (CMFCatalogAware.py) calls:
1. `notifyWorkflowCreated()` → `WorkflowTool.notifyCreated()` → `_reindexWorkflowVariables(ob)`
2. `indexObject()` — queues a full INDEX covering all fields

The `_reindexWorkflowVariables` call is redundant because `indexObject()` immediately after covers all indexes. Worse, `reindexObjectSecurity()` inside it forces an immediate `processQueue()` via `unrestrictedSearchResults(path=path)`. For a newly created object this is pointless: it has no children and is not yet in the catalog.

Workflow transitions are not affected — they continue to use `_reindexWorkflowVariables` via `_invokeWithNotification` on the base class.

## Profiling results (100 content objects: 50 Documents + 50 News Items)

| Operation | Before | After | Improvement |
|-----------|--------|-------|-------------|
| **ADD** (100 objects) | 8.31s (83.1ms/obj) | 4.75s (47.5ms/obj) | **-43%** |
| **UPDATE** (100 objects) | 1.23s (12.3ms/obj) | 1.12s (11.2ms/obj) | -9% |
| **DELETE** (100 objects) | 1.00s (10.0ms/obj) | 0.98s (9.8ms/obj) | -2% |
| **Function calls (ADD)** | 17.7M | 11.1M | **-37%** |

## Test plan

- [x] `Products.CMFPlone` WorkflowTool tests pass (10/10)
- [ ] Full Plone test suite
- [ ] Verify workflow state is correctly set on newly created content
- [ ] Verify workflow transitions still work (security reindex via `_invokeWithNotification` is unchanged)

Fixes #4275

### DIFF
--- a/news/4269.bugfix.md
+++ b/news/4269.bugfix.md
@@ -1,0 +1,1 @@
+Do not use the remote url of a link item in navigation.  [maurits]
--- a/src/Products/CMFPlone/browser/navtree.py
+++ b/src/Products/CMFPlone/browser/navtree.py
@@ -139,12 +139,6 @@ class SitemapNavtreeStrategy(NavtreeStrategyBase):
         if portalType is not None and portalType in self.viewActionTypes:
             itemUrl += "/view"
 
-        useRemoteUrl = False
-        getRemoteUrl = getattr(item, "getRemoteUrl", None)
-        isCreator = self.memberId == getattr(item, "Creator", None)
-        if getRemoteUrl and not isCreator:
-            useRemoteUrl = True
-
         isFolderish = getattr(item, "is_folderish", None)
         showChildren = False
         if isFolderish and (portalType is None or portalType not in self.parentTypesNQ):
@@ -163,9 +157,9 @@ class SitemapNavtreeStrategy(NavtreeStrategyBase):
         newNode["Description"] = getattr(item, "Description", None)
         newNode["show_children"] = showChildren
         newNode["no_display"] = False  # We sort this out with the nodeFilter
-        # BBB getRemoteUrl and link_remote are deprecated, remove in Plone 4
+        # BBB link_remote is deprecated, remove in Plone 7
         newNode["getRemoteUrl"] = getattr(item, "getRemoteUrl", None)
-        newNode["useRemoteUrl"] = useRemoteUrl
+        newNode["useRemoteUrl"] = False
         newNode["link_remote"] = (
             newNode["getRemoteUrl"] and newNode["Creator"] != self.memberId
         )


jensens:

jensens:
I’m pretty sure the problem should be addressed at its roots in CMFCore rather than here, so this PR can be closed. But let’s wait a bit.

jensens:
i think with https://github.com/zopefoundation/Products.CMFCore/issues/152 merged we dont need to fight this front any more.

PR #4280, opened 2026-02-19
Skip security tests for `string._re` on Python 3.14.

mauritsvanrees:
It does not exist anymore

Up until Python 3.13, the `string` module imported the `re` module and made it available as `_re`.  Some tests check that this cannot be abused. But in 3.14 this is no longer the case, so you get three test failures like this:

```
$ bin/test -s Products.CMFPlone -m test_expressions -t test_traverse_function_string
...
Error in test test_traverse_function_string (Products.CMFPlone.tests.test_expressions.TestDirectAttackVector.test_traverse_function_string)
  File "/Users/maurits/.pyenv/versions/3.14.2/lib/python3.14/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/Users/maurits/.pyenv/versions/3.14.2/lib/python3.14/unittest/case.py", line 669, in run
    self._callTestMethod(testMethod)
  File "/Users/maurits/.pyenv/versions/3.14.2/lib/python3.14/unittest/case.py", line 615, in _callTestMethod
    result = method()
  File "/Users/maurits/community/plone-coredev/6.2/src/Products.CMFPlone/src/Products/CMFPlone/tests/test_expressions.py", line 143, in test_traverse_function_string
    traverse_function(string, ("_re", "purge"), None)
  File "/Users/maurits/shared-eggs/v5/cp314/Zope-6.0b2-py3.14-macosx-26.2-arm64.egg/Products/PageTemplates/Expressions.py", line 87, in boboAwareZopeTraverse
    found = traversePathElement(object, name, path_items,
  File "/Users/maurits/shared-eggs/v5/cp314/zope.traversing-6.0-py3.14-macosx-26.2-arm64.egg/zope/traversing/adapters.py", line 149, in traversePathElement
    return traversable.traverse(nm, further_path)
  File "/Users/maurits/shared-eggs/v5/cp314/zope.traversing-6.0-py3.14-macosx-26.2-arm64.egg/zope/traversing/adapters.py", line 54, in traverse
    raise LocationError(subject, name)
zope.location.interfaces.LocationError: (<module 'string' from '/Users/maurits/.pyenv/versions/3.14.2/lib/python3.14/string/__init__.py'>, '_re')
```

At first I wanted to patch `string` on 3.14 so we still test similar situations: even if a module is trusted (like `string`), any module/function/class within this module that starts with an underscore must NOT be available in untrusted code.

But we already test this same vulnerability for `random._os`, which is still there on 3.14, so we don't need a hacked up test case.

For the record, this was the patch that I created, to be used in three tests in `test_expressions.py`:

```
from contextlib import contextmanager

@contextmanager
def _patch_string_re():
    """Make sure string._re exists."""
    need_string_patch = not hasattr(string, "_re")
    if need_string_patch:
        string._re = re
    try:
        yield
    finally:
        if need_string_patch:
            del string._re
```

### DIFF
--- a/.meta.toml
+++ b/.meta.toml
@@ -3,7 +3,7 @@
 # See the inline comments on how to expand/tweak this configuration file
 [meta]
 template = "default"
-commit-id = "2.2.2"
+commit-id = "2.3.2.dev0"
 
 [flake8]
 extra_lines = """
--- a/tox.ini
+++ b/tox.ini
@@ -18,6 +18,7 @@ envlist =
 # Add extra configuration options in .meta.toml:
 # - to specify a custom testing combination of Plone and python versions, use `test_matrix`
 #   Use ["*"] to use all supported Python versions for this Plone version.
+# - to disable the test matrix entirely, set `use_test_matrix = false`
 # - to specify extra custom environments, use `envlist_lines`
 # - to specify extra `tox` top-level options, use `config_lines`
 #  [tox]
@@ -62,6 +63,7 @@ description = check if the package defines all its dependencies
 skip_install = true
 deps =
     build
+    setuptools<82.0.0
     z3c.dependencychecker==2.14.3
 commands =
     python -m build --sdist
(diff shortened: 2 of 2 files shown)


mauritsvanrees:

davisagli (approved):

PR #4285, opened 2026-02-27
Python 3.14 testing matrix, trove classifiers.

thet:
Updated with the branch from: https://github.com/plone/meta/pull/330

### DIFF
--- a/src/Products/CMFPlone/PloneTool.py
+++ b/src/Products/CMFPlone/PloneTool.py
@@ -466,7 +466,7 @@ class PloneTool(PloneBaseTool, UniqueObject, SimpleItem):
 
     @security.public
     def createBreadCrumbs(self, context, request=None):
-        # Returns a structure for the portal breadcumbs.
+        # Returns a structure for the portal breadcrumbs.
         if request is None:
             request = self.REQUEST
         return utils.createBreadCrumbs(context, request)


thet:
I'm not sure, if we should consider this for merging or better try to restore the old `black` behavior (I did like two newlines after the imports...) or even switch to ruff for a all-in-one solution.

gforcada:
I did like the 2 lines as well, but I prefer to not have to fight the tooling and just accept what it comes from them 😅 

If we switch to ruff... there is a PR for that on `plone.meta`, so we might do that indeed.

thet:
@gforcada tnx for checking!

I wanted to know if ruff has a different opinion on the two newlines after the imports, so I also ran ruff on the codebase. Ruff has the same opinion on that, but finds a bunch more problems - e.g. from an earlier formatting occurrences of multi-line string concatenations re-combined on one line like:
`"hello " "gil!"`
which ruff changed to
`"hello gil!"`.\
I re-ran black via pre-commit after ruff to make sure not to apply any changes which black would change again and this is my PR update.

@gforcada please re-check!

thet:

gforcada:
Nice that ruff merges together strings, that's something that I always found annoying... but then `black` promises to keep the ast as it is, and if you join two strings the ast is no longer the same...

Anyway, very nice, and 🏅  approved! 🎉

PR #4293, opened 2026-03-17
Fix 404 page

dobri1408:


dobri1408:

src/Products/CMFPlone/browser/exceptions.py
```diff
@@ -15,11 +17,26 @@ class ExceptionView(BrowserView):
...
+
+    @property
+    @memoize
+    def layout_context(self):
+        portal = get_portal()
+        context = self.error_context or portal
+        context = get_navigation_root_object(context, portal)
+        if getSecurityManager().checkPermission("View", context):
+            return context
+        return portal
```
ale-rt: Skimming the code it seems to me this can be simplified to:
```suggestion
        portal = get_portal()
        if not self.error_context:
            return portal
        context = get_navigation_root_object(self.error_context, portal)
        if getSecurityManager().checkPermission("View", context):
            return context
        return portal
```

src/Products/CMFPlone/browser/exceptions.py
```diff
@@ -50,6 +66,12 @@ def __call__(self):
         request.set("disable_border", True)
         request.set("disable_plone.leftcolumn", True)
         request.set("disable_plone.rightcolumn", True)
+        if error_type == "NotFound":
```
ale-rt: Instead of checking the `error_type` I think it would be good to have a different registration for that specific exception:
https://github.com/plone/Products.CMFPlone/blob/db24e4be79830d296f80bd87ce645d8a89e9c90b/src/Products/CMFPlone/browser/configure.zcml#L250-L257

But I would like to hear @davisagli and @mauritsvanrees opinion on this as it might break existing customizations.

dobri1408:
A separate NotFound registration would only scope the workaround; it would not remove the need to fix the viewlet security chain. I verified that changing only the layout context is not enough for styles to render. Avoiding __parent__ rewriting likely requires a more isolated 404 rendering path.

ale-rt:
Did you check this comment: https://github.com/plone/Products.CMFPlone/pull/4237#discussion_r2567465383

> Gut feeling: I would check the styles viewlet.

Might it be that fixing the styles views would be the proper simpler way?

yurj:
> Did you check this comment: [#4237 (comment)](https://github.com/plone/Products.CMFPlone/pull/4237#discussion_r2567465383)
> 
> > Gut feeling: I would check the styles viewlet.
> 
> Might it be that fixing the styles views would be the proper simpler way?

You mean here? https://github.com/plone/Products.CMFPlone/blob/5f7fe6db8efe9ee0acd01ef2cad6007c86c11092/src/Products/CMFPlone/resources/utils.py#L42

It is called here: https://github.com/plone/Products.CMFPlone/blob/5f7fe6db8efe9ee0acd01ef2cad6007c86c11092/src/Products/CMFPlone/resources/webresource.py#L22

dobri1408:
So, I have done some research, and we can add the styles in the 404 page as a link, and the view will look like this:
Screenshot 2026-03-23 at 13 56 31
which is a bit different from the 404 page when the parent is public: 
Screenshot 2026-03-23 at 13 58 09
The thing is that you need the whole main_template, which is not accessible, is not just a CSS bug is more of a layout bug.

mauritsvanrees:
I looked at this.  Diazo makes debugging this a bit harder, so it can help to add `?diazo.off=1` at the end of the url.  Then it of course looks ugly as well, but you can more clearly see what is really in the content html.

What helps for me, is making the scripts and styles viewlets publicly available:

```
$ git diff
diff --git a/src/Products/CMFPlone/resources/browser/configure.zcml b/src/Products/CMFPlone/resources/browser/>
index 9072f0fc3..a6c19d25b 100644
--- a/src/Products/CMFPlone/resources/browser/configure.zcml
+++ b/src/Products/CMFPlone/resources/browser/configure.zcml
@@ -8,14 +8,14 @@
       name="plone.resourceregistries.scripts"
       manager="plone.app.layout.viewlets.interfaces.IScripts"
       class="Products.CMFPlone.resources.browser.resource.ScriptsView"
-      permission="zope2.View"
+      permission="zope.Public"
       />
 
   <browser:viewlet
       name="plone.resourceregistries.styles"
       manager="plone.app.layout.viewlets.interfaces.IHtmlHeadLinks"
       class="Products.CMFPlone.resources.browser.resource.StylesView"
-      permission="zope2.View"
+      permission="zope.Public"
       />
```

Then it looks like this:

Screenshot 2026-05-21 at 18 06 24

So there is nothing at the top: especially the main menu does not show up.  Which makes sense, otherwise the menu might show a link to the folder that you don't have access to.  Or you would see a breadcrumb link to it.

But for the rest it looks nice.

Actually, it is enough to do this for the styles viewlet.

That *should* be safe, I suppose.
What do others think?

ale-rt:
I agree with @mauritsvanrees analysis, see also https://github.com/plone/Products.CMFPlone/pull/4237#discussion_r2567465383.

If some JS or CSS should stay private, the condition field should be used, see

- https://classic.demo.plone.org/@@resourceregistry-controlpanel

Probably it was like that also for the old `portal_css` and `portal_havascript` tool, I can't remember.

We could even argue that protecting them with the `View` permission is a bug.

mauritsvanrees:
Closed in favour of PR #4327.

PR #4294, opened 2026-03-18
Some improvements in recycle bin feature

cekk:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't any other open pull requests for the same change.
- [x] I followed the guidelines in [Contributing to Plone](https://6.docs.plone.org/contributing/index.html).
- [x] I successfully ran code quality checks on my changes locally.
- [x] I successfully ran tests on my changes locally.
- [x] If needed, I added new tests for my changes.
- [ ] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

I fixed some issues related to massive folders deletion.
In the original implementation contents were stored twice: each deleted content were saved also in the tool's root and then inside its parent (in "children" attribute).
Now we store only the object that triggered the deletion and inside its children attribute there are infos about all its children that are being deleted.

I need to discuss about max size logic too:
There is a setting that allows to set a max bin size (default 100 mb).
Now it works that if you are deleting a big content, the tool tries to cleanup space deleting old contents until there is enough free space to store it into the bin. This can cause confusion to users because it's not clear to them that when they are deleting a content, some items in the bin will be removed.

My proposal is to add this information on ["delete_confirmation" popup](https://github.com/plone/plone.app.content/blob/master/src/plone/app/content/browser/configure.zcml#L112).

I need to update this view to show also bin info (something like "This content is too big, so we are going to cleanup older contents into the recycle bin. Are you sure?")

If i directly update the code there, i need to call recyclebin tool from CMFPlone (to check bin status) but i don't know if i can do that from plone.app.content (there isn't a direct dependency)
Alternatively i've seen that delete_confirmation calls a "delete_confirmation_info" view and inject it into the template as some additional infos. That view is already registered for linkintegrity for example, so i can't override it.
I could try to change this behavior and use a list of adapters and not a single view to allow multiple info slots.

I don't know what could be the right way to do it..any suggestion?

Last but not least: restapi does not have a "confirm" route, only @linkintegrity service: i should add info also there.


davisagli:
@cekk 

> In the original implementation contents were stored twice: each deleted content were saved also in the tool's root and then inside its parent (in "children" attribute). Now we store only the object that triggered the deletion and inside its children attribute there are infos about all its children that are being deleted.

Good. I think I was trying to convince Rohan to do this earlier, to keep things simpler.

> I need to discuss about max size logic too: There is a setting that allows to set a max bin size (default 100 mb). Now it works that if you are deleting a big content, the tool tries to cleanup space deleting old contents until there is enough free space to store it into the bin. This can cause confusion to users because it's not clear to them that when they are deleting a content, some items in the bin will be removed.

I would propose getting rid of the max size feature entirely for now. I don't think it is likely to work well without more effort, because the get_size method of content is not reliable.

cekk:
I liked the idea of a size limit to keep the bin small, but deleting a large folder would still result in a bloated bin anyway. 

Removing this feature will prevent content from 'magically disappearing' for editors. I think we can live with just the retention policy (date limit) for now. Let's scrap the size feature.

PR #4167, opened 2025-04-28
Refactor Migration Tool

ericof:

PR #4162, opened 2025-04-24
Added support for custom base profiles without subclassing

rohnsha0:
fixes #4155

Issue #4155, opened 2025-04-10
Support custom base profiles in MigrationTool without subclassing


ericof:
Distributions can support having a base profile other than `Products.CMFPlone:plone`, allowing for more granular control over what gets installed—including dependencies—during site creation.

Currently, to fully benefit from this approach, distribution maintainers must subclass `Products.CMFPlone.MigrationTool.MigrationTool`, as demonstrated [here in portalbrasil.core](https://github.com/portal-br/core/blob/main/backend/src/portalbrasil/core/tools/migration.py#L76).

A more flexible alternative would be to allow setting the **package name** (e.g., `portalbrasil.core`) and the **base profile** (e.g., `portalbrasil.core:base`) as persistent attributes on the existing `MigrationTool`. This would remove the need for subclassing and simplify adoption for other distributions.

mauritsvanrees:
This should be fixed by my PR #4295 which has been merged.

Issue #4154, opened 2025-04-10
Products.CMFPlone.factory.addPloneSite ignores the profile_id when creating a site based on a distribution


ericof:
This [calll](https://github.com/plone/Products.CMFPlone/blob/12b6cf172cc34906fc4cb9787b25112373020361/Products/CMFPlone/factory.py#L164) to `plone.distribution.api.site.create_site` does not pass the profile_id parameter.

Possible fix 
```python
    site = site_api._create_site(
        context=context,
        distribution_name=distribution,
        answers=answers,
        profile_id=profile_id,
    )
```

ujsquared:
[`_create_site`](https://github.com/plone/plone.distribution/blob/843314adbadc7f9da38a79ef8b707344eefd8a57/src/plone/distribution/api/site.py#L98) itself doesn't assign the profile_id from passed `answers` but instead from fetched `distribution` object(using `distribution-name`). @ericof Is the intention of this ticket to include a fix for that as well?

EDIT: Also, I'm guessing that there is a possibility that `profile_id` of the `distribution` object fetched using API might be different from the one being passed to `_create_site` through `answers`. Is this correct?

jensens:
Verified still present. factory.py passes profile_id in answers dict but plone/distribution/api/site.py always uses distribution.profile_id, ignoring the parameter.

mauritsvanrees:
I made related changes in PR #4295 which has been merged.  This does not change anything about `factory.addPloneSite`, but maybe it is not needed.

Wait, let me think about this.

1. Currently, `factory.addPloneSite` *always* passes a `profile_id`.  By default this is `_DEFAULT_PROFILE`, so `Products.CMFPlone:plone`.
2. If you want to use `factory.addPloneSite` to install a distribution, and this distribution defines its own base profile, then the `profile_id` that is passed in answers *should* be ignored.
3. Actually, the only code in Plone (`parts/packages`) that calls `addPloneSite` is `Products/CMFPlone/browser/admin.py`, which is the view that is *only* available when the `plone.distribution` package is not present.  Plus some tests in `parts/packages/Products/CMFPlone/tests/test_factory.py`, including one that passes a distribution name.

Anyway, what may make sense, is:

* Change `addPloneSite` to have keyword argument `profile_id=None`.
[...]

PR #4295, opened 2026-03-23
Refactor migration tool to need less subclassing

mauritsvanrees:
Refs issue #4154 and #4155.
For me this replaces PR #4162 and #4167 and takes code from there.
See my [review comment on that last PR](https://github.com/plone/Products.CMFPlone/pull/4167#issuecomment-3962332812).

I don't use all code from those PRs, for example I do not add type hints (can be nice, but they mean more changed lines and a bigger PR, harder to review), I don't change the version number reporting (related code, but can be handled separately), and I don't set up a `logger` for passing around.  All possibly useful, but not for now, not this late before 6.2 RC.

For the default case with our two standard distributions, nothing should change with this PR.  But I only tested it lightly.

I did not actually try setting a different base profile id or base package name.  This may need a change in `plone.distribution`, which could set `portal_migration.profile` to the `profile_id` that can get passed in the `answers`.  Or you still subclass the `MigrationTool` and override `profile` there.  Maybe something more is needed, but it should definitely be much easier than today.

Please let me know if this PR is good enough and useful enough to put into Plone 6.2 RC 1, or if we should wait for something better in 6.3.  It's late, and for me it does not have to go in, but I thought I would try and help out.

Main changes:

* Make `AddonList` a named utility and register ours under the name `Products.CMFPlone`.
* Set `MigrationTool.profile`, by default `Products.CMFPlone:plone`.
* Add `MigrationTool.package_name` property method, taking the package name from the profile, so by default `Products.CMFPlone`.

If you override the `profile`, you *must* define your own add-on list in a utility with the name of your package.  I guess we could fall back to the default utility if not found, but then I don't think it makes much sense to override that profile.

cc @rohnsha0 (I wanted to add you to the reviewers, but I don't see you in the list.)

### DIFF
--- a/news/+contentbrowser_update.bugfix.md
+++ b/news/+contentbrowser_update.bugfix.md
@@ -1,0 +1,1 @@
+Fix `pat-contentbrowser` markup changes in robottests.  @petschki
--- a/src/Products/CMFPlone/tests/robot/test_edit.robot
+++ b/src/Products/CMFPlone/tests/robot/test_edit.robot
@@ -114,7 +114,7 @@ I select a related item
     # Click first element in first column
     Click item in contenbrowser column    1    1
     # Click the select Button in the Toolbar of column 2
-    Click    //div[contains(@class, "content-browser-wrapper")]//div[contains(@class, "levelColumns")]/div[2]/div[contains(@class, "levelToolbar")]//button[contains(@class, "btn-outline-primary")]
+    Click    //div[contains(@class, "content-browser-wrapper")]//div[contains(@class, "levelColumns")]/div[2]/div[contains(@class, "levelToolbar")]//button[contains(@class, "btn-primary")]
 
 I select a linked item
     # Click the select button
@@ -123,7 +123,7 @@ I select a linked item
     Click item in contenbrowser column    1    1
     # Click the select Button in the Toolbar of column 2
     # This selects the "test-folder"
-    Click    //div[contains(@class, "content-browser-wrapper")]//div[contains(@class, "levelColumns")]/div[2]/div[contains(@class, "levelToolbar")]//button[contains(@class, "btn-outline-primary")]
+    Click    //div[contains(@class, "content-browser-wrapper")]//div[contains(@class, "levelColumns")]/div[2]/div[contains(@class, "levelToolbar")]//button[contains(@class, "btn-primary")]
 
 I save the page
     Click    //button[@name="form.buttons.save"]


mauritsvanrees:
Sorry, I made some changes, making this PR longer, and harder to review.  But it simplifies things.  With this, no changes in `plone.distribution` should be needed.  I did not try that, but as long as `plone.distribution` sets the wanted base profile it should all work.

`MigrationTool.profile` is now a method that gets the base profile id that has been set in the `portal_setup` tool.

> If you override the profile, you must define your own add-on list in a utility with the name of your package. I guess we could fall back to the default utility if not found, but then I don't think it makes much sense to override that profile.

I changed my mind.  We look for an `IAddonList` utility with the name of the package (derived from the profile we found).  If this utility is not found, we fall back to the add-on list from Plone.

Optionally in the `IAddonList` utility you can set a `pre_addon_list`.  If you have a Plone distribution with an own base profile, you may want to add the default Plone profile here, so the core of Plone is updated first:

    pre_addon_list = AddonList([Addon(profile_id="Products.CMFPlone:plone")])

I see that some of the version stuff from the other PRs is still needed.  I did not fully take that over.  But at least the list of versions on the plone-upgrade page and the overview control panel now contain the package name and package version and profile version of the base profile, in addition to everything that is there already.  In the overview control panel I now see this:

```
PluggableAuthService None (1.6.1)
Plone 6.2.0a2.dev0
...
```

This is because in `portal_setup` I activated the `PluggableAuthService:simple` base profile.  The 1.6.1 is from its `metadata.xml`.  The `None` in the versions would normally be the package version, but because of the old manual profile registration in `PluggableAuthService` our code starts looking for a `PluggableAuthService` package, where this should be `Products.PluggableAuthService`.  I don't want to make a special case for that. For a normally registered base profile this would work fine and show the Python package version.

Note: the base profile must be something like `actual.package:name`.  If it is `actual.package.some.sub.folder:name` I fear some things won't work.

I added tests.

src/Products/CMFPlone/controlpanel/browser/overview.py
```diff
@@ -75,11 +75,25 @@ def server_info(self):
...
+                "{} {} ({})".format(
+                    core_versions["core_package"],
+                    core_versions["core_version"],
+                    core_versions["Plone Instance"],
+                )
+            )
+            versions.append(f"Plone {core_versions['Plone']}")
+        else:
+            versions.append(
+                "Plone {} ({})".format(
```
ericof: Same thing, I would go with fstrings

mauritsvanrees:
The GHA tests were passing.  Now they will fail because I moved the `IAddonList` interface to `plone.base`, so this needs https://github.com/plone/plone.base/pull/115 to be merged, released and put on dist.plone.org.

But I will test the two together on Jenkins.

mauritsvanrees:
The Jenkins 6.2 3.10 job failed with one unrelated [robot failure](https://jenkins.plone.org/job/pull-request-6.2-3.10/603/testReport/junit/Products.CMFPlone.tests.test_robot/RobotTestCase/Scenario_Enable_Dublin_Core_Metadata_in_the_Site_Control_Panel/) (playwright process not found).  The 3.14 one passed.

mauritsvanrees:
I merged master in here.  GHA should get green now, because I updated dist.plone.org so it has the latest plone.base.  Jenkins was already green, so can be merged once GHA is green.

mauritsvanrees:
Note: this causes problems on the Zope root when `plone.distribution` is not available.  See issue #4307.  I have a fix in PR #4308.

PR #4298, opened 2026-03-24
Set no-cache headers on login views

### LINKED ISSUE #3403: login browser view should not be cached by proxy cache because of came_from
For a full discussion:
https://community.plone.org/t/varnish-and-login-page/14692/6

The login browser view html output contains, as hidden input, the parameter `came_from`. When you click on login on a public view, the `came_from` parameter in the form input hidden field is set to the url where you clicked "login". The idea is to login and go back in the same point you were before. If you put a proxy cache (eg varnish) in front of Plone and click on "login", the `login` page is cached , thus also the `came_from` input value. So next time you click on "login", you'll get the cached version (because the url is the same) that, after the successful login, will redirect you to the cached `came_from` wrong url, not the actual correct one.

When you just click on "login" in a Plone url, the [login.py](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/login/login.py) browser view will populate the `came_from` value to the http_referrer (which is the url you came from when clicking on "login").

This is different from hitting a private view and get redirected to login, because the came_from is in the url, thus the cache entry will be different and all works correctly.

Use case:
Suppose you're an editor. You go to a page, then click login and then modify to edit it. This simplify a lot the editor experience. If login would not honor the `came_from`, the editor would login, redirected to the home page, and then have to search the page again to modify it.

This is Plone 6.0.0a2 but the same problem apply to every Plone version which supports `came_from` from the referrer.

jensens:
Prevent proxy caches from serving stale `came_from` values by setting `Cache-Control` and `Expires` headers on `LoginForm` and `FailsafeLoginForm`.

The login page reads `came_from` from `HTTP_REFERER`, which varies per request. Without no-cache headers, a proxy serves the cached HTML with the wrong redirect target after login.

Includes tests for both views.

Fixes #3403

### DIFF
--- a/src/Products/CMFPlone/tests/test_login_form.py
+++ b/src/Products/CMFPlone/tests/test_login_form.py
@@ -43,9 +43,7 @@ class TestLoginForm(unittest.TestCase):
         )
 
     def test_failsafe_login_view_sets_no_cache_headers(self):
-        view = getMultiAdapter(
-            (self.portal, self.request), name="failsafe_login"
-        )
+        view = getMultiAdapter((self.portal, self.request), name="failsafe_login")
         view()
         response = self.request.response
         self.assertEqual(


jensens:

yurj:
Is there a chance to set an etag using an hash of the value of HTTP_REFERER? In this way the server can still return a 304, useful also for installations with a caching proxy. This mean change 'private' to 'public'. I'm not a cache expert, so ok to be more conservative if it helps in corner cases (for example navigation or footer/theme changes). Something like:
```
ETag: "hash-on-referer"
Cache-Control: public, max-age=300, must-revalidate
Vary: Referer
```

where the etag is:

```
import hashlib, time

import hashlib
def get_simple_etag(referer):
    # Hash on Referer
    return f'"{hashlib.md5(referer.encode()).hexdigest()}"'
```

In this way, with no reverse proxy, Plone will cache based on referer for max 5 minutes. But I think we could use plone.app.caching settings on /login and keep:
```
 response.setHeader("Expires", "Sat, 1 Jan 2000 00:00:00 GMT")
 response.setHeader("Cache-Control", "max-age=0, must-revalidate, private")
```

So plone.app.caching will add the proper headers that overwrite this, if configured.

But this is another issue, having /login under plone.app.caching. So your PR is the correct way to solve the issue.

For me is an ok, do you want me to make a review so you can merge?

jensens:
@yurj The login view has no caching ruleset registered in plone.app.caching — it falls through the lookup entirely. No Cache-Control headers are set by any framework.

An ETag based on HTTP_REFERER wouldn't help: the problem is that a proxy caches the *rendered HTML* containing the wrong `came_from` hidden field value. Even with ETag validation, the proxy would still serve stale HTML to a different user until the ETag check round-trips to the backend.

`Cache-Control: max-age=0, must-revalidate, private` is the standard pattern used by plone.app.caching's own `NoCaching` operation and by the `OK` view in CMFPlone. Setting it directly on the view is more robust than requiring p.a.caching to be installed.

yurj:
With an ETag with must-revalidate, the browser always contact the server, issues an HEAD and checks if the Etag is changed. Said that, you're right plone.app.caching is not a mandatory package, and the login caching should be solved in plone.app.caching. I've opened an issue: https://github.com/plone/plone.app.caching/issues/157 here

The fix could be the login URL not using any http parameter and use a fast uncached step that just set a cookie or browser storage with the came_from and login using that instead of having it in the template. But it is complex and also has to offer a fall back for non cookie situations.

Thanks for the fix, it solves perfectly the issue.

Issue #2951, opened 2019-09-18
Berlin’s instead of Paris’ default time zone for French Plone sites


dbitouze:
### What I did:
I installed a fresh Plone 5.2 (python3) instance and started to create a new Plone site.
Maybe because of the locale `LANG=fr_FR.UTF-8` on my Linux box, the automatically selected language for the Plone site was French. 

### What I expect to happen:
I expected the automatically selected time zone to be Paris.
### What actually happened:
The automatically selected time zone was Berlin.

### What version of Plone/ Addons I am using:

- Plone 5.2.0 (5207)
- CMF 2.4.0
- Zope 4.1.1
- Python 3.7.3 (default, Apr 4 2019, 08:55:10) [GCC 8.3.1 20190329]
- PIL 6.1.0 (Pillow)
- WSGI : On
- Serveur : waitress 1.3.0

jaroel:
Thank you for your report!

For whom is willing to work in this:
If we're already shipping with moment.js then we can use `https://momentjs.com/timezone/docs/#/using-timezones/guessing-user-timezone/`. If not we could ignore IE11 and use `Intl.DateTimeFormat().resolvedOptions().timeZone` . IMHO the added value isn't that high that I would add polyfills for IE11.

dbitouze:
Aside remark (from [this thread](https://community.plone.org/t/berlins-instead-of-paris-default-time-zone-for-french-plone-sites/9123/8?u=dbitouze)): I'm not sure what made Plone aware of my default language. If it is derived it from the locale of my computer (LANG=fr_FR.UTF-8), it could be nice for it to derive the time zone from LC_TIME. Hence for a computer of a French Canadian user with LANG=fr_CA.UTF-8 and LC_TIME="fr_CA.UTF-8", Plone would select automatically:

- the language "Français (Canada)"
- the default time zone of, say, Montréal.

jensens:
Confirmed still present in Plone 6. The site creation form uses `jstz v1.0.4` (from 2012) which maps UTC+1/DST to `Europe/Berlin`. All users in France, Belgium, Netherlands, Spain, Italy etc. get Berlin as default. Fix: replace jstz with the modern `Intl.DateTimeFormat().resolvedOptions().timeZone` API (supported by all current browsers) and remove the jstz library.

Issue #4115, opened 2024-01-16
PLIP #4115: Enhance timezone support in UI for `plone.app.event` based types


stevepiercy:
# PLIP plone/Products.CMFPlone#4115: Enhance timezone support in UI for `plone.app.event` based types

(Transferred from https://github.com/plone/volto/pull/5629)

## Responsible Persons

- Steve Piercy @stevepiercy
- Add your name and GitHub username

## Abstract

Event content types, or events, have incomplete implementations and sometimes slightly different implementations between the two Plone frontends, Classic UI and Volto. This PLIP proposes to improve the implementations to be more complete and consistent.

In Volto, the event content type does not display a timezone. When visitors are located or travel across multiple time zones, they cannot determine whether the event's time zone is their own local one, the site's default, the author's setting, or the event's location's. This is particularly confusing when an event has online participation, or when participants travel across time zones. In sites that serve its visitors in only one time zone, this is not an issue.

In Classic UI, the event contact type displays only those timezones that have been made available by the Site Administrator.

The history of event content types is in the [`plone.app.event` package](https://github.com/plone/plone.app.event). However it was implemented using `pytz` whereas the standard library `zoneinfo` and `datetime` with an optional third-party package `tzdata` are now recommended for Python 3.9 and later. From the [`pytz` readme](https://pypi.org/project/pytz/):

> Projects using Python 3.9 or later should be using the support now included as part of the standard library, and third party packages work with it such as [tzdata](https://pypi.org/project/tzdata/). `pytz` offers no advantages beyond backwards compatibility with code written for earlier versions of Python.

Plone 6.0 supports Python 3.9-3.13. Plone 6.1 supports Python 3.10-3.13. Now that Python 3.8 support has been dropped, we can use the standard library and `tzdata` instead. Plone 6.2 is expected to be released in Q3 2025. Plone 7.0 is expected to be released in Q2 2026. See [Plone Release Schedule](https://plone.org/download/release-schedule).

See also [PEP 615 – Support for the IANA Time Zone Database in the Standard Library](https://peps.python.org/pep-0615/) and [Talk - Benjamin "Zags" Zagorsky: Handling Timezones in Python](https://www.youtube.com/watch?v=XZlPXLsSU2U).

Several other related minor issues may be addressed as part of this PLIP.

## Motivation

Everyone wants to have a good time with a date. 😉

`zoneinfo` is preferred over `pytz` since Python 3.9.

[...]

PR #4301, opened 2026-03-24
Replace jstz with native Intl API for timezone detection

jensens:
## Summary
- Replace outdated jstz library (v1.0.4, from 2012) with native `Intl.DateTimeFormat().resolvedOptions().timeZone` API
- Delete jstz-1.0.4.min.js and its ZCML registration
- Remove the `<script>` tag loading jstz from the site creation template

jstz could only distinguish timezones by UTC offset + DST behavior, mapping all UTC+1/DST countries (France, Belgium, Netherlands, Italy, Spain) to `Europe/Berlin`. The native Intl API returns the correct IANA timezone.

## Test plan
- [ ] Create a new Plone site — timezone dropdown should auto-select the correct local timezone
- [ ] Test on a machine set to Europe/Paris, Europe/Brussels, etc. — should no longer default to Europe/Berlin
- [ ] Verify the timezone dropdown still works when browser doesn't support Intl (graceful fallback, no JS error)

Fixes https://github.com/plone/Products.CMFPlone/issues/2951

### DIFF
--- a/news/+contentbrowser_update.bugfix.md
+++ b/news/+contentbrowser_update.bugfix.md
@@ -1,0 +1,1 @@
+Fix `pat-contentbrowser` markup changes in robottests.  @petschki
--- a/src/Products/CMFPlone/tests/robot/test_edit.robot
+++ b/src/Products/CMFPlone/tests/robot/test_edit.robot
@@ -114,7 +114,7 @@ I select a related item
     # Click first element in first column
     Click item in contenbrowser column    1    1
     # Click the select Button in the Toolbar of column 2
-    Click    //div[contains(@class, "content-browser-wrapper")]//div[contains(@class, "levelColumns")]/div[2]/div[contains(@class, "levelToolbar")]//button[contains(@class, "btn-outline-primary")]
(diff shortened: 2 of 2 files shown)


jensens:

stevepiercy:
I'd argue not to use any JS library to mutate the local timezone in the browser. There's many problems with doing that. I know this was discussed elsewhere, but can't find the specific conversation where we want to use the following order in preference:

Authenticated user's set timezone > Site's set default timezone > Server's default timezone.

See also https://github.com/plone/documentation/issues/692 and https://github.com/plone/Products.CMFPlone/issues/4115#issuecomment-2665112991.

davisagli:
@stevepiercy I think you missed the context here:
1. This is about the form for adding a new Plone site.
2. There's no mutation. This is just pre-selecting the user's current timezone in the dropdown for picking the site's default timezone. The user can choose a different one, as before.
3. This is not using a JS library. It's removing the prior use of a library in favor of the native API that is now available.

Note, this is the old Plone site add form that is no longer used when plone.distribution is present.

stevepiercy:
Ah, thanks for the explanation.

jensens:

PR #4304, opened 2026-03-25
Fix flaky robot tests: use Wait For Condition instead of Get Text

jensens:
## Summary
- Replace all `Get Text //body contains` with `Wait For Condition Text //body contains` across 21 robot test files (~100 assertions)
- `Get Text` asserts immediately and fails if page hasn't fully rendered after navigation
- `Wait For Condition` retries with a timeout, matching the pattern already used in `keywords.robot`

This fixes the randomly failing "Set Site Language in the Language Control Panel" test and prevents similar flaky failures across all other robot tests.

## Test plan
- [ ] All robot tests should still pass (same assertions, just with retry)
- [ ] The language control panel test should no longer fail randomly

### DIFF
--- a/news/+flaky-robot-tests.bugfix.md
+++ b/news/+flaky-robot-tests.bugfix.md
@@ -1,0 +1,2 @@
+Fix flaky robot tests: replace ``Get Text //body`` with ``Wait For Condition Text //body`` to properly wait for page content after navigation.
+@jensens
--- a/src/Products/CMFPlone/tests/robot/test_actionmenu.robot
+++ b/src/Products/CMFPlone/tests/robot/test_actionmenu.robot
@@ -92,7 +92,7 @@ Scenario:
 an actionsmenu page
     Create content    type=Document    title=${TITLE}
     Go To    ${PLONE_URL}/${PAGE_ID}
-    Get Text    //body    contains    An actionsmenu page
+    Wait For Condition    Text    //body    contains    An actionsmenu page
 
 # WHEN
 
@@ -111,12 +111,12 @@ workflow link is clicked
     Set Suite Variable    ${OLD_STATE}    ${OLD_STATE}
     Click    xpath=//li[@id='plone-contentmenu-workflow']/a
     Click    xpath=(//li[@id='plone-contentmenu-workflow']/ul/li/a)[1]
-    Get Text    //body    contains    Item state changed.
+    Wait For Condition    Text    //body    contains    Item state changed.
 
 I copy the page
     Open Action Menu
     Click    xpath=//li[@id='plone-contentmenu-actions']//a[contains(@class,'actionicon-object_buttons-copy')]
-    Get Text    //body    contains    copied
+    Wait For Condition    Text    //body    contains    copied
 
 I paste
     Go to  ${PLONE_URL}
@@ -152,7 +152,7 @@ state should have changed
     Should Not Be Equal As Strings    ${NEW_STATE}    ${OLD_STATE}
 
 I should see '${message}' in the page
-    Get Text    //body    contains    ${message}
+    Wait For Condition    Text    //body    contains    ${message}
 
 
 # DRY
--- a/src/Products/CMFPlone/tests/robot/test_controlpanel_actions.robot
+++ b/src/Products/CMFPlone/tests/robot/test_controlpanel_actions.robot
@@ -66,7 +66,7 @@ Scenario: Move an action to new category in Actions Control Panel
 
 the actions control panel
     Go to    ${PLONE_URL}/@@actions-controlpanel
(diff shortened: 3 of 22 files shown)


jensens:

wesleybl:
+1

jensens:
In past years I did similar changes already, this PR tries to catch it all over the place.

petschki:
@1letter was porting all tests to playwright ... he might have an opinion on this.

petschki:
My personal opinion is: we had a lot of wait conditions in selenium. Porting everything to playwright promised somehow to get rid of that. Reintroducing the same waits now feels a bit wrong. I think it would be better to rethink the selectors which should contain the desired text. `//body` is known to be unstable because its there, but the content might not have loaded. use more specific selectors would mean, that playwright waits for them itself before checking the content.

jensens:
@petschki You're right. I dug into the robotframework-browser and Playwright source code:

**How it actually works:**

`Get Text` uses `locator.elementHandle()` which waits up to 10s for the element to appear in the DOM. Then the assertion (`contains`) retries for only 1s (`retry_assertions_for` default).

With `//body`: element always exists → 0s element-wait + 1s assertion retry = **1s total tolerance**
With e.g. `//*[@id="global_statusmessage"]`: element appears after render → **10s element-wait** + 1s assertion retry = **~11s tolerance**

`Wait For Condition` just inflates the assertion retry to 10s — it's a band-aid for `//body`, not the proper fix.

**The right approach** is exactly what you said: use specific selectors that leverage Playwright's built-in auto-waiting. The key selectors for Plone:

| Content | Selector |
|---------|----------|
| Status messages | `//*[@id="global_statusmessage"]` |
| Page heading | `//article[@id="content"]//h1` |
| Page content area | `//article[@id="content"]` |
| Footer | `//div[@id="portal-footer-wrapper"]` |

I'll rework the PR with proper selectors instead of the `Wait For Condition //body` approach.

jensens:

1letter:
> @1letter was porting all tests to playwright ... he might have an opinion on this.

be more specific is a good approach.

we have a wrong use case of identifiers, for example: https://jenkins.plone.org/job/pull-request-6.2-3.14/85/testReport/junit/Products.CMFPlone.tests.test_robot/RobotTestCase/Scenario_Enable_use_email_as_login_in_the_Security_Control_Panel 

the "body" aproach fix the test, but the html structure is wrong.

i have currently no time to work on the tests. i'm full in a migration project until the next 4 weeks.

jensens:
Reworked based on feedback from @petschki and @1letter. Third iteration, hopefully the right one:

**Why `//body` was bad:**
Playwright auto-waits for an element to exist before reading text. `//body` always exists → 0s wait. Specific selectors like `//*[@id="global_statusmessage"]` make Playwright wait up to 10s for the element to appear.

**Why specific selectors + `Get Text` alone isn't enough:**
After a `Click` that triggers page navigation, the target element may already exist in the DOM with OLD content from the previous page. Playwright finds it immediately and reads stale text.

**The fix — two rules:**
1. After `Go To` → just use `Get Text` with a specific selector (Go To already waits for `load` event)
2. After `Click` that triggers navigation → add `Wait For Load State    load` before `Get Text`

**Selectors used:**
- Status messages → `//*[@id="global_statusmessage"]`
- Page content → `(//article[@id="content"])[1]` (positional due to duplicate IDs, see #4306)
- Footer → `(//footer)[1]`

jensens:

jensens:
All bullshit. Here we go:

The sequence is:

1. Server renders HTML 
2. Browser receives fully transformed HTML
3. load event fires ← `Wait For Load State` returns here
4. JavaScript (patterns) runs — pat-* patterns initialize, modify DOM
5. Content is visually complete

`Wait For Load State networkidle` might be better than load — it waits until no network activity for 500ms, which would catch patterns that make AJAX calls. But even that wouldn't catch all pure DOM manipulation by patterns. Only with high probability.

`Wait For Condition` retries every 10ms until the condition is met or the timeout (10s) expires.

The original `Get Text` with its default `retry_assertions_for=1s` polled every 10ms, but gave up after only 1 second.

`Wait For Condition ...` (poll until text appears) is the only reliable approach for Plone pages with active patterns.

So we could make a difference in the tests between pure server side code checks and those with patterns involved, but that's a PITA to write. One clear rule (use `Wait For Condition`) is way easier and makes in practice no difference (same speed in execution).

petschki:
there is the css class `patterns-loaded` on the body when all patterns are loaded... maybe that could help?

jensens:
> there is the css class `patterns-loaded` on the body when all patterns are loaded... maybe that could help?

Not sure, but `Wait For Condition` is really the safety belt here and it does not hurt. Everything else may break sooner or later.

PR #4309, opened 2026-03-27
Preserve query parameters in redirect after login

### LINKED ISSUE #4201: Redirect after login is broken if url contains parameter
After login, the redirect goes to the portal homepage, not to the subpage.
This is similar to #4151 but it occurs on Plone 6.1.2 and only if and there are **parameter in the url**

Background: We have "EasyForm" forms that set the default value for form fields from url parameter. Some of this forms are private and users have to login first. 

What I did:

On https://classic.demo.plone.org/

    login as manager
    Goto https://classic.demo.plone.org/en/demo/folder_contents
    Copy "A page"
    logout
    Visit https://classic.demo.plone.org/en/demo/copy_of_a-page?foo=bar
    login
    i see the portal homepage

What I expect to happen:

    i see https://classic.demo.plone.org/en/demo/copy_of_a-page?foo=bar

What actually happened:

    i see the portal homepage

What version of Plone/ Addons I am using:

    Plone 6.1.2 (6109)

jensens:
## Summary

- `came_from` URLs with query strings (e.g. `?foo=bar`) were silently dropped after login, redirecting to the portal root instead
- Root cause: `physicalPathFromURL` doesn't handle query strings — a URL like `/page?foo=bar` produces a path ending in `page?foo=bar`, which fails `unrestrictedTraverse`, triggering the fallback to portal root
- Fix: strip query string and fragment with `urlsplit` before path validation, while preserving the original `came_from` URL for the final redirect

**Security note:** No new attack surface — the URL is already validated as internal via `isURLInPortal()`, and users can visit any internal URL with any params directly anyway.

## Test plan

- [x] New test `test_redirect_to_came_from_with_query_params` verifies params are preserved
- [x] All 12 existing redirect-after-login tests pass

Fixes #4201

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/news/4201.bugfix.md
+++ b/news/4201.bugfix.md
@@ -1,0 +1,3 @@
+Preserve query parameters in redirect after login.
+URLs with query strings (e.g. ``?foo=bar``) no longer redirect to the portal root.
+@jensens
--- a/src/Products/CMFPlone/browser/login/login.py
+++ b/src/Products/CMFPlone/browser/login/login.py
@@ -226,7 +226,13 @@ class LoginForm(form.EditForm):
             came_from = adapter(came_from, is_initial_login)
 
         if came_from:
-            came_from_path = self.request.physicalPathFromURL(came_from)
+            # Strip query string and fragment before path validation,
+            # physicalPathFromURL does not handle them.
+            came_from_url = parse.urlsplit(came_from)
+            came_from_bare = parse.urlunsplit(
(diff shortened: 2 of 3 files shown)


jensens:

jensens:

ale-rt (approved):

ale-rt:

PR #4327, opened 2026-05-22
Always allow access to the scripts and styles viewlets.

mauritsvanrees:
Otherwise the site may look ugly when you don't have access to the context. See discussion in [alternative PR](https://github.com/plone/Products.CMFPlone/pull/4293)

### DIFF
--- a/news/4239.bugfix.md
+++ b/news/4239.bugfix.md
@@ -1,0 +1,3 @@
+Always allow access to the scripts and styles viewlets.
+Otherwise the site may look ugly when you don't have access to the context.
+@mauritsvanrees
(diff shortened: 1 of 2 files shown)


ale-rt (approved):

mauritsvanrees:

avoinea (approved):

mauritsvanrees:
We could even backport this to Plone 6.1.  But doing this so close before the final 6.1 maintenance release (today or next week) makes me uneasy.  So I think I will pass on that one.  The bug may be irritating/ugly, but it is a corner case.

PR #3960, opened 2024-05-16
Remove queryCatalog and getFolderContents skin scripts

jensens:
- part of https://github.com/plone/Products.CMFPlone/issues/3515 
- see also https://github.com/plone/Products.CMFPlone/pull/3949#pullrequestreview-2061606187

PR #4330, opened 2026-05-25
Fix RSS title content reference in search-rss.pt

dobri1408:
Fix a regression introduced in [ #3960.](https://github.com/plone/Products.CMFPlone/pull/3960/changes#diff-1e97d5defdc4c3376e9f3c8ebf483114311da20d0374ca5b6523321f2746db15L8-R33)

The search RSS template previously used `item/title`, but the refactor changed it
to `python:item.Title()`. Since `item` is an IFeedItem adapter
(`DexterityItem` for Dexterity content), it has a `title` property and no
`Title()` method. Restore `item/title`.

Here is the definition https://github.com/plone/Products.CMFPlone/blob/d6e6632e046a91fb558b39b89f85a62659e7f5cf/src/Products/CMFPlone/browser/syndication/adapters.py#L60


src/Products/CMFPlone/browser/syndication/templates/search-rss.pt
```diff
@@ -29,7 +29,7 @@
 
 <tal:block repeat="item python: objects">
   <item rdf:about="" tal:attributes="rdf:about item/link">
-    <title tal:content="python:item.Title()">Title</title>
+    <title tal:content="item/title">Title</title>
```
avoinea, line 32: @dobri1408 For speed, we should use `python:item.title`

avoinea (changes requested):

avoinea (approved):

erral:

Issue #4360, opened 2026-07-31
contact-info form sends email on GET requests


mauritsvanrees:
Plone 6.2.1, 6.1.5.

A client got thousands of spam messages this week.  A part of these were because the standard Classic UI contact form can send an email on a GET request.  I have spam protection from `collective.honeypot`, but this only works on POST requests.

I think Plone should only send an email on a POST request.

I am preparing a PR.

erral:
Fixed by #4361

PR #4361, opened 2026-07-31
Only send mail from contact-info on POST requests.

mauritsvanrees:
See https://github.com/plone/Products.CMFPlone/issues/4360

### DIFF
--- a/news/4360.bugfix.md
+++ b/news/4360.bugfix.md
@@ -1,0 +1,1 @@
+Only send mail from contact-info on POST requests.  @mauritsvanrees
--- a/src/Products/CMFPlone/browser/contact_info.py
+++ b/src/Products/CMFPlone/browser/contact_info.py
@@ -44,6 +44,12 @@ class ContactForm(AutoExtensibleForm, form.Form):
 
     @button.buttonAndHandler(_("label_send", default="Send"), name="send")
     def handle_send(self, action):
+        if self.request.method != "POST":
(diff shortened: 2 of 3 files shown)


mauritsvanrees:

erral (commented):
Can we use [`plone.protect`](https://pypi.org/project/plone.protect/) for this? 

It has decorators or authenticators for this kind of protections. And we would be eating our own dog food, right?

erral:
Or better even, don't z3c.form forms have a ` method = 'post' ` attribute than force them to be a POST?

mauritsvanrees:
`plone.protect` has `postOnly`, see the [readme](https://github.com/plone/plone.protect/#1-restricting-to-http-post).  But using it as a decorator only works when the function has a `REQUEST` parameter.

We could directly pass the request to this function:

```
from plone.protect import PostOnly
PostOnly(self.request)
```

I am fine with allowing the GET request and simply ignoring the query parameters, as in my PR.  But there is something to be said for denying it with `postOnly`.

I don't see anything in `z3c.form` for this.  At least: searching for `self.method` finds nothing, and even searching for `POST` or `GET`.  Ah, there is an [interface](https://github.com/zopefoundation/z3c.form/blob/master/src/z3c/form/interfaces.py#L1052-L1057) with a field `method`, but I don't see any code reading this.

ale-rt (approved):

erral (approved):

PR #4369, opened 2026-09-02
Fix broken ARIA reference on SVG icons

### LINKED ISSUE #3394: Broken ARIA reference in SVG icon
### What I did:

Triggered a status info message just by e.g. logging in, and checking the SVG tag of the icon next to the message text.

The issue came up while testing accessibility with the WAVE Web Accessibility Evaluation Tool browser extension https://wave.webaim.org/extension/.

### What I expect to happen:

The `aria-labelledby` attribute in the SVG element should reference a `title` tag with an `id="title"` attribute (or, even better, a more specific id attribute value) inside the SVG.

```html
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="plone-icon statusmessage-icon mb-1 me-2 bi bi-info-circle-fill" viewBox="0 0 16 16" aria-labelledby="title" style="">
  <path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16zm.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"></path>
  <title id="title">statusmessage_mtype_info</title>
</svg>
```

Furthermore, a `role="img"` attribute could be added to the SVG tag, q.v. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#svg_and_roleimg.

### What actually happened:

An `aria-labelledby` reference exists, but the target for the reference does not exist.

```html
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="plone-icon statusmessage-icon mb-1 me-2 bi bi-info-circle-fill" viewBox="0 0 16 16" aria-labelledby="title" style="">
  <path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16zm.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"></path>
  <title>statusmessage_mtype_info</title>
</svg>
```

Plone 6.0.0a2 (6002)

kunalKumar-13:
Closes #3394.

Implements the approach @jensens set out on the issue in March. Quoting it so the reasoning is visible here:

> All SVG icons get `aria-hidden="true"` — screen readers skip them. Keep `<title>` for browser hover tooltip when `tag_alt` is provided, but not for screen reader labeling. Remove the broken `aria-labelledby`.

Every `icons.tag()` call site renders the icon next to its own text label, so the icons carry no information of their own and are decorative in the ARIA sense.

### The bug

`_add_aria_title` set `aria-labelledby="title"`, referencing `id="title"`. That id was never set on the `<title>` element, and `_strip_id` removes every `id` from the tree afterwards anyway — so the reference could never resolve. This is what the WAVE checker flagged in the original report.

### One extra thing this surfaced

Writing the tests turned up a second, latent problem in the same function. The title was built with `etree.Element("title")`, which lands in **no namespace**, while the lookup on the line above searches for `{svg-ns}title`. So a title created by this code could never be found again, and a second pass over the same tree appended a duplicate `<title>`.

It is now created in the SVG namespace. **The serialised output is unchanged** — lxml emits `<title>Bug</title>` either way, which is why this was invisible in practice — so this is not a rendering change, only a correctness one. There is a test pinning both the idempotency and the serialised bytes.

Happy to split that into its own PR if you would rather keep this one to the ARIA change alone.

### Tests

Six new cases in `SVGAriaTest`, exercising `_add_aria_title` directly so they need no Plone layer:

- icon is `aria-hidden="true"`, with and without an alt text
- `aria-labelledby` is not emitted, and an existing one is removed
- `<title>` kept when an alt is given, absent when it is not
- a second pass does not duplicate the title
- the serialised output contains `aria-hidden`, no `aria-labelledby`, and `<title>Bug</title>`

I ran those six locally against this branch (Python 3.13) and they pass; `black --check` is clean on both files. I have not run the full integration suite locally.

### Note on the earlier discussion

@agitator and @thomasmassmann discussed `role="presentation"` and unique ids back in 2022. This follows @jensens' later proposal instead, which supersedes that — the icons are decorative, so no id-based labelling is needed at all. Happy to revisit if that is not the intent.

### DIFF
--- a/news/3394.bugfix.md
+++ b/news/3394.bugfix.md
@@ -1,0 +1,9 @@
+Fix the broken ARIA reference on SVG icons. The icon resolver emitted
+`aria-labelledby="title"`, pointing at an `id` that was never set — and
+`_strip_id` removes every `id` from the tree in any case, so the reference could
+never resolve. Since every `icons.tag()` call site renders the icon next to its
+own text label, the icons are decorative and are now marked `aria-hidden="true"`
+so screen readers skip them. The `<title>` tag is still emitted when an alt text
+is given, for the browser hover tooltip, and is now created in the SVG namespace
+so a repeated pass finds it instead of appending a duplicate.
+[kunalKumar-13]
--- a/src/Products/CMFPlone/browser/icons.py
+++ b/src/Products/CMFPlone/browser/icons.py
@@ -17,18 +17,28 @@ SVG_MODIFER = {}
 
 
 def _add_aria_title(svgtree, cfg):
+    root = svgtree.getroot()
+    # Every icons.tag() call site renders the icon next to its own text label,
+    # so the icons carry no information of their own and are decorative in the
+    # ARIA sense. Hide them from assistive technology rather than have screen
+    # readers announce them twice.
+    root.attrib["aria-hidden"] = "true"
+    # The previous "aria-labelledby" pointed at id="title", which was never set
+    # on the title element -- and _strip_id removes every id from the tree in
+    # any case, so the reference could never resolve. See issue #3394.
+    root.attrib.pop("aria-labelledby", None)
     if not cfg.get("title"):
         return
-    root = svgtree.getroot()
     ns = root.nsmap.get(None, "")
-    # set title tag
+    # A title tag is still useful: browsers show it as a hover tooltip. It is
+    # not used for screen reader labelling, which aria-hidden now suppresses.
     title = root.find(f"{{{ns}}}title")
     if title is None:
-        title = etree.Element("title")
-        root.append(title)
+        # Build it in the SVG namespace. A bare etree.Element("title") lands in
+        # no namespace, so the lookup above could never find a title this code
+        # had created and a second pass would append a duplicate one.
+        title = etree.SubElement(root, f"{{{ns}}}title" if ns else "title")
     title.text = cfg["title"]
-    # add aria attr
-    root.attrib["aria-labelledby"] = "title"
 
 
 SVG_MODIFER["add_aria_title"] = _add_aria_title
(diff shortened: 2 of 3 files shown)


kunalKumar-13:
Good catch from the Copilot review, and it was a real gap — pushed in c472ef9.

`root.nsmap.get(None, "")` reads the *default* namespace mapping. An SVG that declares its namespace with a prefix (`<svg:svg xmlns:svg="http://www.w3.org/2000/svg">`) has no default entry, so that returned `""` and the title would have been created outside the SVG namespace again — precisely the problem this change was meant to fix. It now uses `etree.QName(root).namespace`, which reads the namespace off the root element itself and handles both forms. There is a test for the prefixed case.

All seven unit tests pass locally on Python 3.13 and `black --check` is clean on both files.

kunalKumar-13:

kunalKumar-13:

jensens (approved):
LGTM

jensens:

kunalKumar-13:
Gentle nudge on this one — everything it was waiting on has come back green:

- Plone Contributors Agreement verifier — passed
- Changelog verifier — passed
- Plone Jenkins CI `pull-request-6.3-3.14` and `6.3-3.10` — both passed after the run you kicked off on 3 September

@jensens it is approved and `mergeable_state` is clean, so I do not think anything is outstanding on my side. Is there anything else you would like changed before it goes in, or is it just waiting for someone with the merge button?

No rush at all — flagging it only in case it slipped off the list rather than because it is blocked on a decision.

